PR #4, opened 2020-06-14
Added 5.0.0 Aerospike Enterprise Server config schema

ashishshinde:


deploy/operator.yaml
```diff
@@ -33,7 +33,8 @@ spec:
       serviceAccountName: aerospike-kubernetes-operator
       containers:
         - name: aerospike-kubernetes-operator
-          image: aerospike/aerospike-kubernetes-operator:v0.0.2-beta
+          image: localhost:5000/ashish/test-operator:v0.0.2
```
sud82, line 36: Please change this before merging in master.

pkg/controller/aerospikecluster/aerospikecluster_controller.go
```diff
@@ -223,16 +223,15 @@ func (r *ReconcileAerospikeCluster) reconcileCluster(aeroCluster *aerospikev1alp
 				needRollingRestart = true
 			}
 		}
+	} else {
```
sud82, line 226: Do we need this? If accessControl and updateStatus failed earlier then this function should update them at the end. Or may be we can't even reach till end....need to discuss.

PR #21, opened 2020-11-12
Plurarize list types, mountable config maps and sidecars

ashishshinde:
- Aerospike config list type field names are not pluralized. Schemas and examples have been updated.
- Config maps can be mounted as volumes in the storage section
- New sidecars can be attached to Aerospike pods using a new podSpec section


pkg/apis/aerospike/v1alpha1/aerospikecluster_types.go
```diff
@@ -366,21 +434,32 @@ func (v *AerospikeStorageSpec) SetDefaults() {
...
+		} else if v.Volumes[i].VolumeMode == AerospikeVolumeModeFilesystem {
 			v.Volumes[i].AerospikePersistentVolumePolicySpec.SetDefaults(&v.FileSystemVolumePolicy)
 		}
 	}
 }
 
 // GetStorageList gives blockStorageDeviceList and fileStorageList
 func (v *AerospikeStorageSpec) GetStorageList() (blockStorageDeviceList []string, fileStorageList []string, err error) {
-	storagePaths := map[string]int{}
+	storagePaths := map[string]int{
```
sud82, line 445: If this path is also given by user then It would be better to give error saying that this path is reserved.

pkg/controller/utils/utils.go
```diff
@@ -151,9 +152,21 @@ func IsCrashed(pod *v1.Pod) bool {
...
-	actualImage := strings.TrimPrefix(image2, DockerHubImagePrefix)
-	return actualImage == desiredImage
+	desiredImageWithVersion := strings.TrimPrefix(image1, DockerHubImagePrefix)
+	actualImageWithVersion := strings.TrimPrefix(image2, DockerHubImagePrefix)
+
+	desiredRegistry, desiredName, desiredVersion := ParseDockerImageTag(desiredImageWithVersion)
+	actualRegistry, actualName, actualVersion := ParseDockerImageTag(actualImageWithVersion)
+
+	// registry name, image name and version should match.
+	return desiredRegistry == actualRegistry && desiredName == actualName && (desiredVersion == actualVersion || (desiredVersion == ":latest" && actualVersion == "") || (actualVersion == ":latest" && desiredVersion == ""))
```
sud82, line 162: Is it just a defensive check to see if desiredVersion is :latest or it can be this? I think you have added a check to disallow user from providing empty or :latest version.

pkg/controller/utils/utils.go
```diff
@@ -187,19 +200,59 @@ func IsPVCTerminating(pvc *corev1.PersistentVolumeClaim) bool {
...
+
+	return nil
+}
+
+// IsPodOnDesiredImage indicates of pod is ready and on desired images for all containers.
+func IsPodOnDesiredImage(pod *corev1.Pod, aeroCluster *aerospikev1alpha1.AerospikeCluster) bool {
+	for _, ps := range pod.Spec.Containers {
+		desiredImage, err := GetDesiredImage(aeroCluster, ps.Name)
+		if err != nil {
+			// Maybe a deleted sidecar. Ignore.
```
sud82, line 246: If container is deleted then should we check for it's status also or just go to the next container?

pkg/controller/aerospikecluster/controller_helper.go
```diff
@@ -983,6 +989,86 @@ func updateStatefulSetSecretInfo(aeroCluster *aerospikev1alpha1.AerospikeCluster
...
+			}
+		}
+
+		if !found {
+			// Add to stateful set containers.
+			st.Spec.Template.Spec.Containers = append(st.Spec.Template.Spec.Containers, newSidecar)
+		}
+	}
+
+	// TODO: remove containers.
```
sud82, line 1009: Remove container TBD?

pkg/controller/utils/utils.go
ashishshinde: Yes this is intentional because if user specifies ":latest" or empty then comparing versions makes no sense. For the latest, the operator will have to check the latest version and we could unintentionally upgrade if we push new versions to the docker hub.

pkg/controller/utils/utils.go
ashishshinde: This is the case where the user has removed the sidecar from podSpec. Since he has removed the sidecar we can assume that we are on the desired Image, since this container will eventually be removed during a rolling upgrade.

pkg/controller/aerospikecluster/controller_helper.go
ashishshinde: Oops.. Needs to be implemented :)

pkg/apis/aerospike/v1alpha1/aerospikecluster_types.go
ashishshinde: Yes will change.

pkg/controller/admission/validate_cluster.go
```diff
@@ -75,8 +75,12 @@ func (s *ClusterValidatingAdmissionWebhook) ValidateUpdate(old aerospikev1alpha1
 	}
 
 	// Jump version should not be allowed
-	newVersion := strings.Split(s.obj.Spec.Build, ":")[1]
-	oldVersion := strings.Split(old.Spec.Build, ":")[1]
+	newVersion := strings.Split(s.obj.Spec.Image, ":")[1]
+	oldVersion := ""
```
sud82, line 79: Was it crashing? old image should always be there in case of update i think. Just curious!

ashishshinde: Oh yes. I had a few running clusters running before this rename :) They were getting stuck and not being deleted because the field name was "build"

sud82 (approved):

PR #39, opened 2021-06-01
Make it possible to set custom docker registry for init container

jfwm2:
- This is done through an env var AEROSPIKE_KUBERNETES_INIT_IMAGE
 - If this env var is not set the default image is used; the default image name is moved to utils.utils.go
 - The env var definition is also set at the same location in utils.go

Signed-off-by: Jean-Francois Weber-Marx <jf.webermarx@criteo.com>
Signed-off-by: Jean-Francois Weber-Marx <jfwm@hotmail.com>

### DIFF
--- a/pkg/controller/aerospikecluster/controller_helper.go
+++ b/pkg/controller/aerospikecluster/controller_helper.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -119,7 +120,7 @@ func (r *ReconcileAerospikeCluster) createStatefulSet(aeroCluster *aerospikev1al
 					//TerminationGracePeriodSeconds: &int64(30),
 					InitContainers: []corev1.Container{{
 						Name:  utils.AerospikeServerInitContainerName,
-						Image: "aerospike/aerospike-kubernetes-init:0.0.12",
+						Image: GetAerospikeServerInitContainerImage(aeroCluster),
 						// Change to PullAlways for image testing.
 						ImagePullPolicy: corev1.PullIfNotPresent,
 						VolumeMounts: []corev1.VolumeMount{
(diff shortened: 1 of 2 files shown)


sud82:
Hi @jfwm2, It looks that you are assuming that the init-container image will be the same across all the aerospike clusters. As the operator handles multiple aerospike clusters, we are thinking of giving an option for init container in the custom resource itself. Another way can be that we can have env var and CR config both. In that case, CR config will override the env var. 
Let us think a bit about this.

jfwm2:
Hello @sud82, indeed it would be better/easier to have a CR config.
I also noticed the init version container used in the operator is 0.0.12, whereas 1.1.0 is in use in https://github.com/aerospike/aerospike-kubernetes-enterprise, an the last one is 1.2.2 if I am not mistaken. Furthermore, versions before 1.0.0 are not available in https://github.com/aerospike/aerospike-kubernetes-init/.
Do you know where I could find the source code of versions 0.0.12?

spkesan:
@jfwm2 The init container used in Aerospike Pods deployed by operator is in https://github.com/aerospike/aerospike-kubernetes-operator/tree/master/docker/init

The `initialize.sh` script is part of configmap. Check https://github.com/aerospike/aerospike-kubernetes-operator/blob/master/pkg/controller/configmap/configdata.go

jfwm2:
> @jfwm2 The init container used in Aerospike Pods deployed by operator is in https://github.com/aerospike/aerospike-kubernetes-operator/tree/master/docker/init
> 
Thanks @spkesan this is exactly what I needed!

Regarding the present PR, I think this is already an improvement to have a configurable (by env var) init container until further development. The configuration using CR config could be done later. What do you think about that?

spkesan:
@jfwm2 Let us discuss this internally with the team. As @sud82 mentioned, it makes sense to have this configurable within the CustomResource and that also seems like an ideal solution. We will keep you posted.

jfwm2:
Thanks @spkesan! Looking forward to hearing from you soon :)

sud82:
Hi @jfwm2, I have bumped up API from v1alpha1 to v1beta1. You can modify this PR to take initContainerImage from CR rather than EnvVar.

jfwm2:
> Hi @jfwm2, I have bumped up API from v1alpha1 to v1beta1. You can modify this PR to take initContainerImage from CR rather than EnvVar.

Hello @spkesan I am not sure I understand; where (which files) should modify for the PR?

sud82:
Released in 2.1.

PR #40, opened 2021-06-07
Add the possibility for alternate rack ids

jfwm2:
When the label "aerospike.com/alternate-rack-id" is set at the node level, it used in the Aerospike configuration (/etc/aerospike/aerospike.conf) as rack-id.
The Aerospike node id is not updated with alternate rack id, we stick to the pattern {rack-id}a{pod-ordinal} already in use. 
To add the alternate rack id to the aerospike server configuration, a new env var containing the node name was added. By using the kubernetes api in the init scripts (running on the init container) the "aerospike.com/alternate-rack-id" label is retrieved. If such label does not exist (or is not an integer) the behaviour does not change.

### DIFF
--- a/pkg/controller/aerospikecluster/controller_helper.go
+++ b/pkg/controller/aerospikecluster/controller_helper.go
@@ -85,6 +85,7 @@ func (r *ReconcileAerospikeCluster) createStatefulSet(aeroCluster *aerospikev1al
 		newEnvVar("MY_POD_NAMESPACE", "metadata.namespace"),
 		newEnvVar("MY_POD_IP", "status.podIP"),
 		newEnvVar("MY_HOST_IP", "status.hostIP"),
+		newEnvVar("MY_NODE_NAME", "spec.nodeName"),
 		newEnvVarStatic("MY_POD_TLS_NAME", getServiceTLSName(aeroCluster)),
 		newEnvVarStatic("MY_POD_CLUSTER_NAME", aeroCluster.Name),
 	}
--- a/pkg/controller/configmap/configdata.go
+++ b/pkg/controller/configmap/configdata.go
@@ -46,6 +46,7 @@ done
 CA_CERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
 TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
 KUBE_API_SERVER=https://kubernetes.default.svc
+NODE=$MY_NODE_NAME
 NAMESPACE=$MY_POD_NAMESPACE
 
 CFG=/etc/aerospike/aerospike.template.conf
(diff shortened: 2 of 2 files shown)


sud82:
Hi @jfwm2, can you please explain your use case. Why do you want an alternate rackID. rackID is already configured by the user.

jfwm2:
Hello @sud82, if I am not mistaken, the rackID that is already present is static and need to by specified at deployment time. The alternate-rack-id that I am proposing is intended to match id of the physical (it could also be logical) rack containing the (physical) node on which a pod is scheduled; because it would be read by the init container, it can be setup at scheduling time. Does it make sense for you or do you need more details?

sud82:
Hi @jfwm2, If I got you correctly then you want to match rack-id with the physical/logical rack-id
rack-id in CR can be given to match your physical/logical rack-id. It should be given is CR because the whole idea behind giving rack-id is to schedule the pods based on rack configuration. 
Can you please give an example, where you need the alternate-rack-id and rack-id is not enough?

jfwm2:
Hello @sud82

> Hi @jfwm2, If I got you correctly then you want to match rack-id with the physical/logical rack-id

Indeed, just to be clear on rack ids, let's have to following naming:
- ARI: Aerospike Rack ID -> Rack ID specified in /etc/aearospike.conf
- ORI: Operator Rack ID -> Rack ID specified in the CR
- PRI: Physical Rack-ID -> ID of the rack containing the physical machine hosting aerospike-server container
We would like to have the ARI set with the value of the PRI, not the ORI as it is the case now.

> rack-id in CR can be given to match your physical/logical rack-id. It should be given is CR because the whole idea behind giving rack-id is to schedule the pods based on rack configuration.
> Can you please give an example, where you need the alternate-rack-id and rack-id is not enough?

I understand the whole idea behind this design, but it does not fit our setup. We are running kubernetes on premises, on bare metal hosts, that do not have a uniform configuration; each rack can contain a lot of various configurations used by different services. This setup is done on purpose, to avoid that the loss of a rack would impact a significant number of entity used in the same service.
Therefore, to make sure we can use the Aerospike native rack awareness, we must update the ARI with the PRI. To do so, we can expose the PRI corresponding to a host with the label "aerospike.com/alternate-rack-id" and the code of the present PR would take care of the rest.

If this is not clear enough, and if you have time, we can have a small video call with screen sharing so I could go in more details if needed.

Thanks
@jfwm2

sud82:
OK. Let's have a call then. Please let me know when you would be available for that. You can drop a mail 
 to me (sudhanshu@aerospike.com)

jfwm2:
Thanks @sud82! I just e-mailed you directly 🙂

nareshmaharaj-consultant:
This is no longer a requirement and confirmed with Peter Goron (3-Feb-23)

PR #43, opened 2021-06-10
Fix helm chart indent issue with resources in deployment.yaml

eb-koddi:
changes indent to nindent so that YAML is properly formed when actually specifying resource requests and limits with the helm chart values.


spkesan (approved):
LGTM

spkesan:
Thanks @eb-koddi for the contribution.

PR #45, opened 2021-06-24
Support list for tls-authenticate-client.

skolosov-snap:
This PR is continue of https://github.com/aerospike/aerospike-management-lib/pull/7
This will change all schemas to support lists for tls-authenticate-client.
Also added validation to prevent mixing "false" or "any" values with other values in a list.
Also added some extra validation on tls configuration preventing set tls params without providing a valid tls-name.

### DIFF
--- a/api/v1alpha1/aerospikecluster_validating_webhook.go
+++ b/api/v1alpha1/aerospikecluster_validating_webhook.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 
 	//"github.com/aerospike/aerospike-kubernetes-operator/api/v1alpha1"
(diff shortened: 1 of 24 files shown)


api/v1alpha1/aerospikecluster_validating_webhook.go
```diff
@@ -381,34 +384,115 @@ func validateAerospikeConfig(aslog logr.Logger, configSpec *AerospikeConfigSpec,
...
+	for _, connectionType := range networkConnectionTypes {
+		if err := validateNetworkConnection(networkConf, tlsNames, connectionType); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func validateTlsClientNames(networkConf map[string]interface{}) error {
```
sud82, line 438: 1. tls-authenticate-client can not be "true", That check needs to be added.
2. How are we going to provide the client name of the operator in the tls-authenticate-client list? Are there any more pull reqs coming related to the tls-authenticate-client?

skolosov-snap: Discussed offline:
1. Aerospike can successfully start with tls-authenticate-client=true. If we still want this validation it should be a separate PR with separate discussion.
2. There are many ways how to extract correct name from the cert (different parts of subject name, different extensions in SAN). In order to make it less error prone and make it more flexible we will let user to manually specify names in a config.

spkesan: I believe there's no validation done by aerospike for `true` in `tls-authenticate-client`. We only check for `any` and `false` and anything else would be treated as a tls name and Aerospike will go forward with client authentication enabled.

PR #47, opened 2021-06-30
Fix multiPodPerHost helm char default override

jfwm2:
- In aerospike cluster helm chart the value of multiPodPerHost true if not defined: {{ .Values.multiPodPerHost | default true }}
  - When the value of multiPodPerHost is set to false in a chart, the value defined about gives true.
  - To fix that, this change set the attribute in values.yaml to true.

Signed-off-by: Jean-Francois Weber-Marx <jf.webermarx@criteo.com>

### DIFF
--- a/helm-charts/aerospike-cluster/templates/aerospike-cluster-cr.yaml
+++ b/helm-charts/aerospike-cluster/templates/aerospike-cluster-cr.yaml
@@ -15,7 +15,7 @@ spec:
   image: {{ .Values.image.repository | default "aerospike/aerospike-server-enterprise" }}:{{ .Values.image.tag | default "5.5.0.7" }}
 
   # Multi pod per host
-  multiPodPerHost: {{ .Values.multiPodPerHost | default true }}
+  multiPodPerHost: {{ .Values.multiPodPerHost }}
 
   # Aerospike access control configuration
   {{- with .Values.aerospikeAccessControl }}
--- a/helm-charts/aerospike-cluster/values.yaml
+++ b/helm-charts/aerospike-cluster/values.yaml
@@ -17,7 +17,7 @@ image:
 imagePullSecrets: {}
 
 ## Multi pod per host
-# multiPodPerHost: true
+multiPodPerHost: true
 
 ## Aerospike access control configuration
 aerospikeAccessControl: {}


spkesan:
@jfwm2 
The default is set to `true` for `multiPodPerHost`. You can override the default value using an "override-values-yaml" file that can be passed in using `-f` or `--values` during `helm install` or otherwise override the default value using `--set multiPodPerHost` command line option.

I'm assuming that you are trying to fetch the chart locally and modifying the `values.yaml` file, which doesn't seem like a right approach to me.

Also, I believe the change that is proposed in this PR would force users to always set `multiPodPerHost` in their "override-values-yaml" file.

jfwm2:
Hello (sorry for the late answer, I was in vacation 🌞 )

As of now `multiPodPerHost` is not set in the default value file and there is `multiPodPerHost: {{ .Values.multiPodPerHost | default true }}` in the template cr file. Three possibilities can occur regarding this setting when using using a yaml file to override defaults value:

1. `multiPodPerHost` is not set; therefore `{{ .Values.multiPodPerHost | default true }}` is evaluated to `true`
2. `multiPodPerHost` is set to `true` and `{{ .Values.multiPodPerHost | default true }}`  is evaluated to `true`
3. `multiPodPerHost` is set to `false` and `{{ .Values.multiPodPerHost | default true }}`  is evaluated to `true` (and this is not what is expected) because `.Values.multiPodPerHost` is evaluated to `false`, which cause the default value (i.e, `true`) to be used; I confirmed this running the `helm --debug` command line when deploying; See also  https://github.com/helm/helm/issues/3308#issuecomment-876076586 for info.

This shows that currently `multiPodPerHost` can only have the value `true` 🙁 

The change I am proposing would set `multiPodPerHost` to `true` in the default value file and use `multiPodPerHost: {{ .Values.multiPodPerHost }}` in the template. Therefore, we would obtain the expected behaviour for the three possibilities that could occur for this setting when using using a yaml file to override defaults value:

1. `multiPodPerHost` is not set; therefore `{{ .Values.multiPodPerHost }}` is evaluated to `true` (from the default value file)
2. `multiPodPerHost` is set to `true` and `{{ .Values.multiPodPerHost }}`  is evaluated to `true`
3. `multiPodPerHost` is set to `false` and `{{ .Values.multiPodPerHost }}`  is evaluated to `false` as expected

Please, @spkesan @arrowplum , just let me know what you think about it.

PR #57, opened 2021-07-18
Allow pods to use HostNetwork

ashishshinde:
- New boolean configuration "hostNetwork" added to pod spec section that enables pods to use host network. This configuration is valid only when multiPodPerHost configuration is false
 - New "dnsPolicy" configuration added to podSpec. This defaults to ClusterFisrtWithHostNet when host network is enabled and ClusterFirst when host network is disabled
 - Fixes in init scripts to not reply on pod's hostname for rack and node id
 - Fixes incorrect fabric tls value used for pod exposed ports
 - Tests to verify that heartbeat and fabric use host network when enabled and not the k8s internal pod network


controllers/scripts/create-aerospike-conf.sh
```diff
@@ -106,6 +104,18 @@ cat $PEERS | while read PEER || [ -n "$PEER" ]; do
...
+
+# ------------------------------------------------------------------------------
+# If host networking is used force heartbeat and fabric to advertise network
+# interface bound to K8s node's host network.
+# ------------------------------------------------------------------------------
+{{- if .HostNetwork}}
+# 8 spaces, fixed in configwriter file config manager lib
+# TODO: The search pattern is not robust. Add a better marker in management lib.
+sed -i -e "/heartbeat {/a \\        address ${MY_POD_IP}" ${CFG}
+sed -i -e "/fabric {/a \\        address ${MY_POD_IP}" ${CFG}
```
skolosov-snap, line 116: What happen if address is already there?
Wouldn't it better to use the same approach we used for "service" with <access-address> placeholder?

skolosov-snap: Just come to my mind: what about tls-address? Do we need the same behavior for it?

PR #58, opened 2021-07-19
Support InitContainers in CR AerospikeCluster.

skolosov-snap:
Supporting configurations of init containers in CRD:
- added an InitSidecars AerospikePodSpec. All containers from this list will be converted into init containers or the pods.
- found all places where Spec.PodSpec.Sidecars and updated it with assumption that we have init-containers in addition to regular containers (technically just duplicated the behavior for init-containers).
- added additional check if containers are updated. This seems to be redundant but provides more meaningful logs for pod restart reasoning than checksum check.


controllers/pod.go
```diff
@@ -107,6 +107,12 @@ func (r *AerospikeClusterReconciler) getRollingRestartTypePod(aeroCluster *asdbv
 		r.Log.Info("Aerospike resources changed. Need rolling restart")
 	}
 
+	// Check if podSpec is updated
+	if isPodSpecUpdatedInAeroCluster(aeroCluster, pod) {
```
skolosov-snap, line 111: This is redundant, but produces more meaningful log messages than checksum check. LMK if you want me to get rid of it.

sud82:
These changes are have already been fixed in other pull req.

PR #59, opened 2021-07-20
Unhardcode port values.

skolosov-snap:
Here is current logic:
- All AS ports are hardcoded in [constants here](https://github.com/aerospike/aerospike-kubernetes-operator/blob/1a3f18c7a01f54b9392cd3832b650bb3a5886378/api/v1alpha1/utils.go#L23-L39).
- All these ports are injected into AS config in [web hook](https://github.com/aerospike/aerospike-kubernetes-operator/blob/1a3f18c7a01f54b9392cd3832b650bb3a5886378/api/v1alpha1/aerospikecluster_mutating_webhook.go#L307-L368). Thus nobody can set other values.
- The same constants are also used to specify ports in [pod's specs](https://github.com/aerospike/aerospike-kubernetes-operator/blob/38a528566e31e986f8f869d7b326547a7a9345e1/controllers/statefulset.go#L850-L921).
- Also these consts are used [to establish connections to AS cluster](https://github.com/aerospike/aerospike-kubernetes-operator/blob/9341bc18de827e4b0c07761c2beb3d9393635c37/controllers/aero_info_calls.go#L181-L200).
- **access** and **alternate access** ports are set to service port (and tls service port). Later these 4 ports [are replaced in the config by aerospike-init container](https://github.com/aerospike/aerospike-kubernetes-operator/blob/38a528566e31e986f8f869d7b326547a7a9345e1/controllers/scripts/create-aerospike-conf.sh#L80-L81) depending on network policy and MultiPodPerHost flag.

What was changed:
- All values for ports are now taken from CR config. Port constants are only used as default values.
- **access** and **alternate access** ports are now set to 65535 and this value is replaced by init container.
- validation added to rack forbidding any network configurations specific to rack. The problem is that we use the same heartbeat configuration for all hosts when adding mesh-seed-address-port in create-aerospike-conf.sh. Thus if different rack would have different port configuration heartbeating gets broken.

### DIFF
--- a/api/v1alpha1/aerospikecluster_mutating_webhook.go
+++ b/api/v1alpha1/aerospikecluster_mutating_webhook.go
@@ -310,16 +310,24 @@ func setDefaultNetworkConf(aslog logr.Logger, configSpec *AerospikeConfigSpec, c
 	// TODO: These values lines will be replaces with runtime info by script in init-container
 	// See if we can get better way to make template
 	serviceDefaults := map[string]interface{}{}
-	serviceDefaults["port"] = ServicePort
-	serviceDefaults["access-port"] = ServicePort // must be greater that or equal to 1024
+	srvPort := GetServicePort(configSpec)
+	serviceDefaults["port"] = srvPort
+	// Here all access ports are explicitly set to hardcoded constant 65535. These values will
+	// be replaced by aerospike-init container with an appropriate port in accordance to
+	// MultiPodPerHost flag (NodePort of Service or Host Port of Pod).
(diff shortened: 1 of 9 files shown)


skolosov-snap:
> @skolosov-snap If the error message pops up for rack scope then something is amiss. I just trying to see if there is an issue there and if we can do without 65535 and use the service port itself. Will comment here soon.

If you are talking about my comment, then the probem is not in the fact of error, but in the misleading message in the error.
Here is the scenario:

1. we create cluster with service.port=1234
2. webhook adds service.access-port=1234
3. now if user will try to set service.access-port=4321 he will get misleading/confusing message "config service.access-port can not have non-default value 4321. It will be set internally"

PR #62, opened 2021-07-26
Storage

sud82:
1: Allow storage volumes to take various storage volume source, e.g. empty dir, configmap, secret, pv
2: Allow storage volumes to take sidecars and initcontainers where this volume should be mounted
3: Allow additional initcontainers and scheduling policies (affinity, toleration..) in podSpec
4: Allow rack level scheduling policies
5: Added storage tests


ashishshinde (approved):

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -920,16 +1162,36 @@ func validateRequiredFileStorage(aslog logr.Logger, configSpec AerospikeConfigSp
...
+	// TODO Add validation for feature key file.
+	featureKeyFilePaths := getFeatureKeyFilePaths(configSpec)
+	tlsPaths := getTLSFilePaths(configSpec)
+
+	var allPaths []string
+	allPaths = append(allPaths, featureKeyFilePaths...)
+	allPaths = append(allPaths, tlsPaths...)
+
+	for _, path := range allPaths {
+		if !storage.IsVolumePresentForAerospikePath(filepath.Dir(path)) {
```
skolosov-snap, line 1178: This will break validation for use cases when credentials are feed not through secret.

ashishshinde: @skolosov-snap This only checks if the path has an associated volume attached. The volume could be any one of a secret, a config map, a shared persistent volume, or a shared empty dir.

PR #63, opened 2021-08-02
Fix 5.6 schema dynamic flag and add missing parameters

jdogmcsteezy:
There were a few parameters missing from the service context.  A number of parameters had an incorrect value for the dynamic flag.  The way I check for dynamic parameters is to check this function: https://github.com/citrusleaf/aerospike-server/blob/5.6.0.8/as/src/base/thr_info.c#L2406.  As of 5.6 though I also need to check this function https://github.com/citrusleaf/aerospike-server-enterprise/blob/master/as/src/base/security_info.c#L145.  Another way I confirm the parameter is dynamic is run the appropriate `set-config` command and look for the `ok` response.  

The `auth-mode` config parameter is different than most and does not show up in these locations even though, in some cases, it is dynamic. Read [this] (https://docs.aerospike.com/docs/reference/configuration/index.html#auth-mode)  for more explanation.  Depending on your intent it could be labeled as dynamic or static.  I changed it to dynamic in the schema. Let me know if you would like me to reverse it.  

Let me know if you have any questions or if you think I am mistaken

### DIFF
--- a/pkg/configschema/5_6_0.go
+++ b/pkg/configschema/5_6_0.go
@@ -16,6 +16,14 @@ const conf5_6_0 = `
         "required": ["feature-key-files"]
       }],
       "properties": {
+        "query-buf-size": {
+          "type": "integer",
+          "default": 2,
+          "minimum": 0,
+          "maximum": 2147483647,
+          "description": "",
+          "dynamic": true
+        },
         "user": {
           "type": "string",
           "default": "",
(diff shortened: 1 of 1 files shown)


jdogmcsteezy:
FYI, I will open a PR soon to update all previous config-schemas

davi17g (approved):
We merged your PR to master, and I modified `
aerospike-server-schema-generator` according to your changes.

PR #66, opened 2021-08-25
Moved schemas to json files

davi17g:


pkg/configschema/schemamap.go
```diff
@@ -1,5 +1,70 @@
 package configschema
 
+import _ "embed"
+
+//go:embed jsonfiles/4_0_0.json
```
ashishshinde, line 5: You can embed the entire folder and create the entire map without listing each version separately - like below
 
https://github.com/aerospike/aerospike-kubernetes-operator/blob/17b6462a23f49c722ed00a9ecade0a6d53581944/controllers/configmap.go#L46

pkg/configschema/jsonfiles/4_0_0.json
```diff
@@ -1,6 +1,3 @@
-package configschema
```
ashishshinde, line 1: Name the folder json instead of jsonfiles

ashishshinde (approved):

PR #67, opened 2021-08-26
Support SecurityContext for aerospike-server container created by operator.

skolosov-snap:
Sometime people use their own Aerospike images with their own startup scripts. I.e. to modify kernel params. They need to configure Security Context for pods in such cases.


api/v1beta1/aerospikecluster_types.go
```diff
@@ -62,6 +62,8 @@ type AerospikeClusterSpec struct {
 	PodSpec AerospikePodSpec `json:"podSpec,omitempty"`
 	// SeedsFinderServices describes services which are used for seeding Aerospike nodes.
 	SeedsFinderServices SeedsFinderServices `json:"seedsFinderServices,omitempty"`
+	// SecurityContext that will be added to aerospike-server container created by operator.
+	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
```
sud82, line 66: I think we can put SecurityContext inside PodSpec

sud82: There is no validation for SecurityContext. According to your use case, do you want to allow the update of SecurityContext by doing a rolling restart or disallow any change?

sud82 (changes requested):

api/v1beta1/aerospikecluster_types.go
skolosov-snap: As discussed offline keeping this field here since it relates to AerospikeContainer only but not to the whole pod. renamed it to AerospikeSecurityContext.

I don't see any reasons to forbid updating it. Thus no need for validation, however supported restarting upon changing in SecurityContext.

api/v1beta1/aerospikecluster_types.go
```diff
@@ -62,6 +62,8 @@ type AerospikeClusterSpec struct {
 	PodSpec AerospikePodSpec `json:"podSpec,omitempty"`
 	// SeedsFinderServices describes services which are used for seeding Aerospike nodes.
 	SeedsFinderServices SeedsFinderServices `json:"seedsFinderServices,omitempty"`
+	// SecurityContext that will be added to aerospike-server container created by operator.
+	AerospikeSecurityContext *corev1.SecurityContext `json:"aerospikeSecurityContext,omitempty"`
```
ashishshinde, line 66: This can be moved to  PodSpec -> AerospikeContainerSpec -> SecurityContext
---
struct PodSpec {
  .
  AerospikeContainerSpec AerospikeContainerSpec `json:"aerospike,omitempty"`
  .
}


struct AerospikeContainerSpec {
   SecurityContext *coverv1.SecurityContext
}
---

This struct can then be used if in the future we want to allow more container-specific properties to be configured.

skolosov-snap: Updated

ashishshinde (approved):

PR #68, opened 2021-08-29
Add Annotations

davi17g:


api/v1beta1/aerospikecluster_types.go
```diff
@@ -136,6 +136,8 @@ func (c *AerospikeOperatorClientCertSpec) validate() error {
 
 // AerospikePodSpec contain configuration for created Aeropsike cluster pods.
 type AerospikePodSpec struct {
+	// MetaData to add to pods.
```
sud82, line 139: Do not use this metadata object. Create a new custom metadata struct having only annotations and labels

type ObjectMeta struct {
	Annotations 
	Labels
}

pkg/utils/utils.go
```diff
@@ -105,6 +105,19 @@ func LabelsForAerospikeClusterRack(clName string, rackID int) map[string]string
 	return labels
 }
 
+// MergeLabels merges operator an user defined labels
+func MergeLabels(operatorLabels, userLabels map[string]string) (map[string]string, error) {
+	for label, value := range userLabels {
+		_, ok := operatorLabels[label]
+		if ok {
+			return nil, fmt.Errorf(
```
sud82, line 113: Add this validation check in validationWebhook. Request should not come here if this check fails there.

api/v1beta1/aerospikecluster_types.go
davi17g: Created new MetaObject

pkg/utils/utils.go
davi17g: Done

pkg/utils/utils.go
```diff
@@ -95,16 +95,25 @@ func GetDesiredImage(aeroCluster *asdbv1beta1.AerospikeCluster, containerName st
...
 func LabelsForAerospikeClusterRack(clName string, rackID int) map[string]string {
 	labels := LabelsForAerospikeCluster(clName)
-	labels["aerospike.com/rack-id"] = strconv.Itoa(rackID)
+	labels[asdbv1beta1.AerospikeRackIdLabel] = strconv.Itoa(rackID)
 	return labels
 }
 
+// MergeLabels merges operator an user defined labels
+func MergeLabels(operatorLabels, userLabels map[string]string) map[string]string {
+	for label, value := range userLabels {
```
ashishshinde, line 111: Make a new map instead of modifying the input and returning

ashishshinde (commented):
Please add a test to verify annotation and labels are applied to running pod

pkg/utils/utils.go
davi17g: Changed

PR #69, opened 2021-08-29
Support non-root user in init container

spkesan:


ashishshinde (commented):
@spkesan You will need to publish the modified init container image to docker hub nad use that as the init container image in statefulset.go to get this to work.

ashishshinde (approved):

PR #73, opened 2021-09-06
Allow not to set fabric port if tls-port is set. Do not override addresses if they are set in CR.

skolosov-snap:
Currently non-tls fabric port is always opened on pod's IP address. This may have some security concerns.
This change is to make non-tls fabric port optional. Now, the default value for non-tls fabric port is set only if TLS is not configured for Fabric.

In addition to that I changed logic in init script which sets address to pod's IP. Now, address is set only if there is no address configuration defined by user.

Also in addition to that support tls-address in init script for Fabric and Heartbeat.


ashishshinde:
Closing this PR since the changes are Incorporated into #85.

PR #75, opened 2021-09-08
Moved aerospikeContainer resources under podSpec.aerospikeContainer

sud82:


controllers/statefulset.go
```diff
@@ -1034,8 +1034,12 @@ func (r *SingleClusterReconciler) getClusterSTSList() (
 }
 
 func (r *SingleClusterReconciler) updateAerospikeContainer(st *appsv1.StatefulSet) {
-	// These resources are for main aerospike container. Other sidecar can mention their own resources.
-	st.Spec.Template.Spec.Containers[0].Resources = *r.aeroCluster.Spec.Resources
+	resources := r.aeroCluster.Spec.PodSpec.AerospikeContainerSpec.Resources
+	if resources != nil {
```
skolosov-snap, line 1038: What if resources were changed from non-null value to null? My guess is that Pod will not be updated. Did I miss anything?

PR #77, opened 2021-09-09
Added full operator docs as docusaurus site.

arrowplum:
Currently collides with the helm charts storage so can't enable it until we figure out a way around that (@spkesan )


arrowplum:
moved to https://aerospike.github.io/kubernetes-operator

PR #83, opened 2021-09-20
Support Operator Deployment on Openshift Platform

spkesan:
- Use RedHat UBI base image.
- Add necessary labels to the container image
- Add license file to the image
- Don't specify user in the image
- Change permissions for entrypoint to root group


Dockerfile
```diff
@@ -18,11 +18,34 @@ COPY pkg/ pkg/
...
+
+# License file
+COPY LICENSE /licenses/
+
 WORKDIR /
+
 COPY --from=builder /workspace/manager .
-USER 65532:65532
+
+RUN chgrp 0 /manager \
```
ashishshinde, line 48: What does this do?

Dockerfile
```diff
@@ -18,11 +18,34 @@ COPY pkg/ pkg/
...
+# Base image
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+# Maintainer
+LABEL maintainer="Aerospike <support@aerospike.com>"
+
+# Labels
+LABEL name="aerospike-kubernetes-operator" \
+    vendor="Aerospike" \
+    version="2.0.0" \
```
ashishshinde, line 30: Can you pass this version from the makefile file as build args or replace using sed before the operator is built

Dockerfile
spkesan: It changes file's group ownership to root.
This is required to be done for openshift where arbitrary users belong to root group.
It is safe for non-openshift deployments as well.

Dockerfile
spkesan: Done. Version will be taken from docker build arg

ashishshinde (approved):

PR #85, opened 2021-09-27
Remove port defaults

sud82:


controllers/scripts/create-aerospike-conf.sh
```diff
@@ -112,10 +117,12 @@ done
 # interface bound to K8s node's host network.
 # ------------------------------------------------------------------------------
 {{- if .HostNetwork}}
-# 8 spaces, fixed in configwriter file config manager lib
+# 8 spaces, fixed in config writer file config manager lib
 # TODO: The search pattern is not robust. Add a better marker in management lib.
 sed -i -e "/heartbeat {/a \\        address ${MY_POD_IP}" ${CFG}
+sed -i -e "/heartbeat {/a \\        tls-address ${MY_POD_IP}" ${CFG}
 sed -i -e "/fabric {/a \\        address ${MY_POD_IP}" ${CFG}
+sed -i -e "/fabric {/a \\        tls-address ${MY_POD_IP}" ${CFG}
```
skolosov-snap, line 125: What if address or tls-address is already set?

ashishshinde: Address including tls-adress are lists. User-defined addresses will be appended.

skolosov-snap: @ashishshinde I mean can this break anything for client. I.e. node has multiple network interfaces and user want to open fabric port on particular one, but you add another one by this logic.

PR #86, opened 2021-09-28
Support different network types to connect to AS from the tests. Support AWS and fix tests.

skolosov-snap:
Changes:
- Currently when test connects to Aerospike, it always connects to hostExternal IPs. Now, you can specify what address to use to connect using ***connect-through-network-type*** parameter. Possible options are 'pod', 'hostInternal' or 'hostExternal'. You may want to specify it depending on where tests are run. I.e. if you run tests inside the K8s cluster and external IPs are not accessible.
- for targets ***test-deploy***/***test-undeploy*** in Makefile, instead of patching it by replacing strings we patch it by KUSTOMIZE patches. This works on all platforms and more reliable.
- When deleting pod service, we do not fail if pod service doesn't exist.
- Cleaning up all clusters from previous run in BeforeEach to avoid side effects of failed tests.
- Some small logging changes.


arrowplum:
Thank you for this pull request. We are still in the process of reviewing.

PR #95, opened 2021-11-03
Update Helm Chart for Aerospike Cluster Custom Resource

spkesan:
- Update Helm Chart for Aerospike Cluster Custom Resource
- Fix Aerospike Kubernetes Operator Helm Chart


helm-charts/aerospike-cluster/README.md
```diff
@@ -9,34 +9,46 @@ A Helm chart for `AerospikeCluster` custom resource to be used with the Aerospik
 
 ## Usage
 
-### Add Aerospike Helm Repository
+<!-- ### Add Aerospike Helm Repository
 
 ```sh
 helm repo add aerospike https://aerospike.github.io/aerospike-kubernetes-operator
+``` -->
+
```
sud82, line 17: You can add about cert-manager in operator's helm. I think, it's missing there

helm-charts/aerospike-cluster/templates/aerospike-cluster-rbac.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   # Do not change name, its hard-coded in operator
-  name: aerospike-cluster
+  name: aerospike-operator-controller-manager
```
sud82, line 6: As this name is hardcoded, we get an error(SA already exists) if we create another cluster in the same namespace. Is there a way to skip the creation and deletion of SA?

sud82: Or would it be possible that SA, Role, Rolebinding is created by operator helm for every watched namespace?

helm-charts/aerospike-cluster/README.md
spkesan: Will add.

helm-charts/aerospike-cluster/templates/aerospike-cluster-rbac.yaml
spkesan: Alright. So apparently we have to move the rbac creation to operator's helm chart as there will be conflict of ownership of the ServiceAccount (since it's hardcoded to a single name) if two AerospikeCluster(s) are deployed onto same namespace.

spkesan: Good catch @sud82

PR #96, opened 2021-11-10
Add 5.7 schema support

davi17g:


ashishshinde:
@davi17g Can you change tests to use 5.7 version of areospike by default and run 5.6 in all versions tests.

PR #103, opened 2021-12-09
initializedVolumePaths must not be null

ashangit:
crd define it as a list of string so trying to initializing it as None failed
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "AerospikeCluster.asdb.aerospike.com \"aerospikes99\" is invalid: status.pods.aerospikes99-21-0.initializedVolumePaths: Invalid value: \"null\": status.pods.aerospikes99-21-0.initializedVolumePaths in body must be of type array: \"null\"",
  "reason": "Invalid",
  "details": {
    "name": "aerospikes99",
    "group": "asdb.aerospike.com",
    "kind": "AerospikeCluster",
    "causes": [
      {
        "reason": "FieldValueInvalid",
        "message": "Invalid value: \"null\": status.pods.aerospikes99-21-0.initializedVolumePaths in body must be of type array: \"null\"",
        "field": "status.pods.aerospikes99-21-0.initializedVolumePaths"
      }
    ]
  },
  "code": 422
}
```

### DIFF
--- a/controllers/scripts/create_pod_status_patch.py
+++ b/controllers/scripts/create_pod_status_patch.py
@@ -170,7 +170,7 @@ value = {
         'tlsName': os.environ.get('MY_POD_TLS_NAME', '')
     },
     'initializedVolumes': initialized,
-    'initializedVolumePaths': None,
+    'initializedVolumePaths': [],
     'aerospikeConfigHash': confHash,
     'networkPolicyHash': networkPolicyHash,
     'podSpecHash': podSpecHash,


ashishshinde:
InitializedVolumePaths can be null according to the rc2 CRD. 

https://github.com/aerospike/aerospike-kubernetes-operator/blob/master/api/v1beta1/aerospikecluster_types.go#L715

Are you sure you are using the correct version of the CRD with 2.0.0-rc2 operator?

If not please use the CRD from rc2 version of the operator.

ashangit:
Seem sthat the crd file inside the aerospike-cluster helm is not updated accordingly: https://github.com/aerospike/aerospike-kubernetes-operator/blob/c95f68cd5cc381ee95ecf0e7cca5f3f701f992b3/helm-charts/aerospike-cluster/crds/customresourcedefinition_aerospikeclusters.asdb.aerospike.com.yaml#L6709

ashangit:
Also from the definition it looks to me that the value can be omitted but can't be set as None, no?

ashishshinde:
Can you use the HELM charts from the master? They have the correct CRD.

> Also from the definition it looks to me that the value can be omitted but can't be set as None, no?
It can be set as None, None, and [] are different. In the k8s object store, the field will be unset when it is set to null/None from python.

ashishshinde:
@ashangit  I pushed your exact same payload to my cluster running 2.0.0-rc2. and here is the output. I still suspect your CRD is incorrect. Output below. Null is a valid value for a field that can be omitted.


```shell
curl -vvv -k -X PATCH -d @/tmp/patch.json -H 'Accept: application/json' \
     -H 'Content-Type: application/json-patch+json' \
     "http://localhost:8080/apis/asdb.aerospike.com/v1beta1/namespaces/test/aerospikeclusters/aerocluster/status?fieldManager=pod"`
```
```
* Connected to localhost (127.0.0.1) port 8080 (#0)
> PATCH /apis/asdb.aerospike.com/v1beta1/namespaces/test/aerospikeclusters/aerocluster/status?fieldManager=pod HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.74.0
> Accept: application/json
> Content-Type: application/json-patch+json
> Content-Length: 1332
> 
* upload completely sent off: 1332 out of 1332 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Audit-Id: 4abb2081-1aa5-41ba-8c46-db3730f43fe6
< Cache-Control: no-cache, private
< Content-Type: application/json
< Date: Tue, 14 Dec 2021 16:30:40 GMT
< X-Kubernetes-Pf-Flowschema-Uid: 4e34e908-51c9-4fc0-8cb8-b92d91c23364
< X-Kubernetes-Pf-Prioritylevel-Uid: b87476b5-eb44-4222-9136-867496289da2
< Transfer-Encoding: chunked
< 
```

Please verify the CRD by fetching it like so

```
kubectl  get crd aerospikeclusters.asdb.aerospike.com -o yaml
```

ashangit:
I confirm that there are the same
[crd_on_cluster.txt](https://github.com/aerospike/aerospike-kubernetes-operator/files/7713367/crd_on_cluster.txt)

We are using kubernetes release 1.16.15

ashishshinde:
@ashangit This seems to be triggered by specific k8s version. You were right, we should mark the field as nullable in the CRD.

I have another PR #107 with the changes to the CRD and helm chart.

Can you give that PR a shot? An even quicker way to verify would be to edit your crd like so

```shell
kubectl edit crd aerospikeclusters.asdb.aerospike.com 

```

and add nullable flag to the initializeVolumePaths field like so

```yaml
...
                    initializedVolumePaths:                                                              
                      description: InitializedVolumePaths is deprecated version of                       
                        InitializedVolumes.                                                              
                      items:                                                                             
                        type: string                                                                     
                      nullable: true  
---
```

ashangit:
Sorry for late feedback it indeed work fine with the nullable

ashishshinde:
@ashangit Thanks. Glad to know it worked. I will close this PR in favor of #107.

Issue #102, opened 2021-12-07
Operator doesn't apply pod affinity change if there are other changes in the manifest


Tema:
Action: 
Apply manifest with several changes like pod affinity to move to other node-pool and new version of some sidecar image.

Expected behavior: 
Operator runs single rolling restart to apply all changes.

Observed behavior: Operator runs rolling restart to bring a new sidecar version, but ignores pod affinity changes. Subsequent manifest re-apply is ignored as operator doesn't see any changes.

Workaround: 
Re-apply manifest with previous pod affinity value which will be effectively no-op. Then apply pod affinity changes again without any other changes, then operator executes rolling restart and honors new pod affinity.

Part of manifest in question:
```
  podSpec:
    aerospikeContainer:
      securityContext:
        privileged: true
    multiPodPerHost: false
    hostNetwork: true
    affinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
            - matchExpressions:
                - key: "aerospike.com/rack-label"
                  operator: "In"
                  values:
                    - {{node_pool_label}}
    sidecars:
      - image: {{aerospike_gcp_maintenance_image}}
```

sud82:
@Tema, Thanks for reporting. We will look into it.

PR #104, opened 2021-12-13
Update statefulset spec also, if spec and image both are changed.

sud82:
Update statefulset spec also, if spec and image both are changed.
https://github.com/aerospike/aerospike-kubernetes-operator/issues/102


arrowplum:
@ashishshinde @sud82 is this the last thing before we GA?

ashishshinde (approved):

PR #105, opened 2021-12-14
Add initImage parameter to set custom docker registry for init container

ashangit:


sud82:
Released in 2.1.

PR #122, opened 2022-03-21
update default block cleaner to use blkdiscard

realmgic:
This was set to use shred. Shred is zeroing out multiple times to the disks. On bigger disks this can take a very long time to finish.


sud82:
This has been added to the documentation.

PR #127, opened 2022-05-04
Add Operator Bundle Automation

davi17g:


Makefile
```diff
@@ -1,10 +1,12 @@
 # # /bin/sh does not support source command needed in make test
 #SHELL := /bin/bash
 
-ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-
+ROOT_DIR=$(shell git rev-parse --show-toplevel)
```
sud82, line 4: Any specific need to change the old one?

Makefile
```diff
@@ -153,9 +155,9 @@ CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
 	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.1)
 
-KUSTOMIZE = $(shell pwd)/bin/kustomize
+KUSTOMIZE = $(GOBIN)/kustomize
```
sud82, line 158: Any reason to change this? It's not consistent with 'CONTROLLER_GEN', 'ENVTEST'. They are using '$(shell pwd)/bin/'.

Makefile
```diff
@@ -153,9 +155,9 @@ CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
...
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.2.0)
+	go install sigs.k8s.io/kustomize/kustomize/v4@v4.5.4
```
sud82, line 160: This will not download if kustomize is missing. Inconsistent with the way 'controller-gen' and 'setup-envtest' are downloaded and installed.

Makefile
```diff
@@ -195,6 +197,73 @@ bundle: manifests kustomize
...
+bundle-ocp: bundle-operatorhub
+
+.PHONY: bundle-rhmp
+bundle-rhmp: DISTRIBUTION=rhmp
+bundle-rhmp: manifests kustomize
+	$(eval BUNDLE_DIR:= $(ROOT_DIR)/bundle/$(DISTRIBUTION)/$(VERSION)/)
+	$(eval ANNOTATIONS_FILE_PATH:= $(BUNDLE_DIR)/metadata/annotations.yaml)
+	$(eval KUSTOMIZE_DIR:= $(OVERLAYS_DIR)/$(DISTRIBUTION))
+	$(eval BUNDLE_METADATA_OPTS:= $(BUNDLE_METADATA_OPTS) --use-image-digests)
+
```
sud82, line 241:
```suggestion
	if [ -f $(ROOT_DIR)/bundle.Dockerfile ]; then \
        rm -f $(ROOT_DIR)/bundle.Dockerfile; \
      fi; \
    if [ -d $(BUNDLE_DIR) ]; then \
      rm -r $(BUNDLE_DIR); \
      fi; \
```
can be replaced with 
```
rm -rf bundle.Dockerfile bundle/
```
This is the same as we do in
```
.PHONY: bundle
bundle: manifests kustomize
```

Makefile
```diff
@@ -195,6 +197,73 @@ bundle: manifests kustomize
...
+	if [ -f $(ROOT_DIR)/bundle.Dockerfile ]; then \
+        rm -f $(ROOT_DIR)/bundle.Dockerfile; \
+      fi; \
+    if [ -d $(BUNDLE_DIR) ]; then \
+      rm -r $(BUNDLE_DIR); \
+      fi; \
+
+	operator-sdk generate kustomize manifests -q
+	cd $(ROOT_DIR)/config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
```
sud82, line 250:
```
	cd $(ROOT_DIR)/config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
```
For the same thing, we use given line in old `bundle` command
```
cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
```
Please make sure that all types of `bundle` are consistent.

config/manifests/bases/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
sud82, line 1: Any reason to add this file?

config/overlays/base/kustomization.yaml
```diff
@@ -1,7 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
sud82, line 1: Any reason to change the dir name for this file?

sud82 (changes requested):

Makefile
davi17g: Much clearer and shorter, IMO

Makefile
davi17g: Change to previous definition

Makefile
davi17g: Changed to previous definition

config/manifests/bases/kustomization.yaml
davi17g: yes it's instead running the following commands:
`sed -i "s@createdAt: dateplaceholder@createdAt: $(DATE)@g" bundle/manifests/aerospike-kubernetes-operator.clusterserviceversion.yaml; \`
`sed -i "s@containerImage: controller:latest@containerImage: $(IMG)@g" bundle/manifests/aerospike-kubernetes-operator.clusterserviceversion.yaml; \`

config/overlays/base/kustomization.yaml
davi17g: Because `rhmp`  is another overlay and it is also one of the `olm` bundle types, I agree that `base` dir name is not perfect but it also cannot be called `olm`, suggestions for the other name are more than welcome.

config/manifests/bases/kustomization.yaml
sud82: But this line is still there in Makefile.

davi17g: removed lines from a Makefile

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: docker.io/aerospike/aerospike-kubernetes-operator-nightly
-  newTag: 2.1.0-dev
+  newName: docker.io/davi17g/aerospike-kubernetes-operator-nightly
```
tanmayja, line 15: Is it ok to use davi17g repo instead of pushing the image to aerospike repo itself. So that we don't have to change the name.

davi17g: no,I pushed it by mistake, I'll fix that

PR #128, opened 2022-05-05
Removing prefix from mountpath in customer init containers. #KO-39

tanmayja:
We add prefix "/workdir/block-volumes" and "/workdir/filesystem-volumes" in mount path for mounting volumes in initContainer. This is needed only for our initContainer aerospike-init but we do this for customers’ initContainers also. 
Hence removing the prefix for the customer’s initContainers.


test/podspec_test.go
```diff
@@ -39,10 +39,17 @@ var _ = Describe(
 			},
 		}
 
+		volumeMounts := []corev1.VolumeMount{
```
sud82, line 42: This mount will bypass your changes. We add a prefix in volume given in the storage section.
You need to add a volume in storage and then test.

sud82 (changes requested):

PR #129, opened 2022-05-09
Providing customisations for the reserved aerospike init container 

sharathsteel:
1. Cluster deployment doesn't allow to override aerospike-init image path. its hardcoded to download from internet. the issue persists in both OLM and Helm. 
2. Aerospike init container doesn't accept securityContext by default. This setting needs to be exposed. 
3. Supporting resources for kube-rbac-proxy and init-container


sud82:
Released in 2.1.

PR #132, opened 2022-05-18
Add-Support-for-6.0

davi17g:


controllers/scripts/create_pod_status_patch.py
```diff
@@ -1,195 +1,255 @@
...
-    initialized.append(volume['name'])
-
-
-def isIpv6Address(host):
+def get_volumes(pod_name, config):
+    rack = get_rack(pod_name=pod_name, config=config)
+    try:
+        print(f"pod-name: {pod_name} Looking for volumes in rack.storage.volumes")
+        volumes = rack['storage']['volumes']
+        if not volumes:
```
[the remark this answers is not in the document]
davi17g: No, KeyError will not be raised when the volume section is defined and holds an empty list.

api/v1beta1/aerospikecluster_types.go
```diff
@@ -390,17 +390,40 @@ const (
 	AerospikeVolumeInitMethodDeleteFiles AerospikeVolumeInitMethod = "deleteFiles"
 )
 
+// AerospikeVolumeWipeMethod specifies how block volumes should be wiped.
+// +kubebuilder:validation:Enum=dd;blkdiscard;deleteFiles
+// +k8s:openapi-gen=true
+type AerospikeVolumeWipeMethod string
+
+const (
```
ashishshinde, line 398: I think you can reuse the init method enum instead of creating a new wipe method enum.

controllers/scripts/update-pod-status.sh
```diff
@@ -73,7 +73,7 @@ else
     echo "Pod restarted"
 fi
 
-echo $AERO_CLUSTER_JSON | python3 create_pod_status_patch.py $MY_POD_NAME
```
ashishshinde, line 76: The cluster JSON could be larger than command line character limits.
Instead, write the json to a file and pass the filename to the patch script.

test/storage_wipe_test.go
```diff
@@ -0,0 +1,272 @@
...
+							)
+							Expect(err).ToNot(HaveOccurred())
+
+							Expect(len(pvcs)).Should(
+								Equal(0), "PVCs not deleted",
+							)
+						}
+					},
+				)
+			},
```
ashishshinde, line 167: You can add one more test where a non-aerospike volume is added. Meaning a volume is created but is not used for Aerospike. This block volume should not be wiped after an upgrade.

controllers/scripts/create_pod_status_patch.py
```diff
@@ -1,195 +1,267 @@
...
+            image = ""
+        print(f"pod-name: {args.pod_name} - Checking if volumes should be wiped")
+        if image:
+            prev_major_ver = get_image_version(image=image)[0]
+            print(f"pod-name: {args.pod_name} - Checking if volumes should be wiped: "
+                  f"next-major-version: {next_major_ver} prev-major-version: {prev_major_ver}")
+            if (next_major_ver >= BASE_WIPE_VERSION > prev_major_ver) or \
+                (next_major_ver < BASE_WIPE_VERSION <= prev_major_ver):
+                print(f"pod-name: {args.pod_name} - Volumes should be wiped")
+                volumes = process_volumes(pod_name=args.pod_name,
```
ashishshinde, line 240: Unlike init which should happen for ALL volumes, here we need to only wipe volumes attached to an Aerospike namespace. 

It's possible there is a volume created that is not used by the Aerospike server container but another sidecar. We should not be wiping that volume.

controllers/scripts/create_pod_status_patch.py
```diff
@@ -1,195 +1,267 @@
...
+                                          dd=dd, find=find,
+                                          blkdiskard=blkdiskard_wipe,
+                                          effective_method_key="effectiveWipeMethod",
+                                          volumes=get_volumes(pod_name=args.pod_name, config=config))
+                update_status(pod_name=args.pod_name, metadata=metadata, volumes=volumes)
+                return
```
ashishshinde, line 246: Use if-else here instead of return. The common post-processing section, update_status should be common and written only once. It's possible there will be more follow-up common steps.  It's also hard to quickly see what is going on.

controllers/scripts/create_pod_status_patch.py
```diff
@@ -1,195 +1,267 @@
...
+        find = "find {volume_path} -type f -delete"
+        blkdiskard_init = "blkdiscard {volume_path}"
+        blkdiskard_wipe = "blkdiscard -z {volume_path}"
+        next_major_ver = get_image_version(image=config["spec"]["image"])[0]
+        try:
+            image = config["status"]["image"]
+        except KeyError:
+            image = ""
+        print(f"pod-name: {args.pod_name} - Checking if volumes should be wiped")
+        if image:
```
ashishshinde, line 233: Could be made a separate method like handle_version_change()

controllers/scripts/create_pod_status_patch.py
```diff
@@ -1,195 +1,267 @@
...
+    "access": "accessEndpoints",
+    "alternate-access": "alternateAccessEndpoints",
+    "tls-access": "tlsAccessEndpoints",
+    "tls-alternate-access": "tlsAlternateAccessEndpoints"
+}
 
 
-def getRack(data, podName):
-    print('Checking for rack in rackConfig')
+def process_volumes(pod_name, find, dd, blkdiskard, volumes, effective_method_key):
```
ashishshinde, line 21: Probably wipe_volumes or initialize_volumes is a better name

api/v1beta1/aerospikecluster_types.go
sud82: If we use init enum then we have to add a check to invalidate the `AerospikeVolumeInitMethodNone` method

test/utils.go
```diff
@@ -525,9 +525,11 @@ func getAerospikeStorageConfig(
 	// TODO
 	ddInitMethod := asdbv1beta1.AerospikeVolumeInitMethodDD
 	blkDiscardInitMethod := asdbv1beta1.AerospikeVolumeInitMethodBlkdiscard
+	blkDiscardWipeMethod := asdbv1beta1.AerospikeVolumeWipeMethodBlkdiscard
```
sud82, line 528: You can include the wipe method for dd and deleteFiles too.

test/storage_wipe_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+				It(
+
+					"Should validate all storage-wipe policies", func() {
+
+						var rackStorageConfig asdbv1beta1.AerospikeStorageSpec
+						rackStorageConfig.BlockVolumePolicy.WipeMethod = asdbv1beta1.AerospikeVolumeWipeMethodBlkdiscard
+						rackStorageConfig.BlockVolumePolicy.InitMethod = asdbv1beta1.AerospikeVolumeInitMethodDD
+						if cloudProvider == CloudProviderAWS {
+							rackStorageConfig.BlockVolumePolicy.InitMethod = asdbv1beta1.AerospikeVolumeInitMethodDD
```
sud82, line 60: What is this `if` condition doing. init and wipe methods are still the same. Are you trying to make them both DD for AWS cloud?

controllers/scripts/update-pod-status.sh
davi17g: I'll include a request for json inside the script it will eliminate the problem

controllers/scripts/create_pod_status_patch.py
davi17g: `wipe_volumes ` or `initialize_volumes` names specific cases of the script, IMO the name should be more general than that

test/utils.go
davi17g: It's not added on purpose - to make sure that defaults are used correctly.

test/storage_wipe_test.go
```diff
@@ -0,0 +1,272 @@
...
+							rackStorageConfig.BlockVolumePolicy.WipeMethod = asdbv1beta1.AerospikeVolumeWipeMethodBlkdiscard
+						}
+						racks := []asdbv1beta1.Rack{
+							{
+								ID:      1,
+								Storage: rackStorageConfig,
```
sud82, line 66: This storage will replace the primary storage for the rack. This doesn't have any volume so setting a rack-level storage policy will not make any difference.

api/v1beta1/aerospikecluster_types.go
ashishshinde: Yes ... None is not allowed

controllers/scripts/update-pod-status.sh
ashishshinde: Or you could store the JSON in a file from the curl directly and the pass the filename to the python script to eliminate the bash variable AERO_CLUSTER_JSON altogether.

test/storage_wipe_test.go
ashishshinde: Those volumes will not be used in the namespace storage.

api/v1beta1/access_control_validate.go
```diff
@@ -275,14 +294,23 @@ func isRoleNameValid(roleName string) (bool, error) {
...
+			return false, err
+		}
+		if cmp < 0 {
+			// First part of the privilege is not part of defined privileges.
+			return false, fmt.Errorf("invalid privilege %s", privilege)
+		}
+		if _, ok := Post6Privileges[parts[0]]; !ok {
+			return false, fmt.Errorf("invalid privilege %s", privilege)
+		}
 	}
```
ashishshinde, line 315: since Privileges is not the complete list of all privileges the code that follows will be incorrect for post60 privilege. See line 324

api/v1beta1/access_control_validate.go
```diff
@@ -51,6 +52,12 @@ var PredefinedRoles = map[string]struct{}{
 	"write":          {},
 }
 
+var Post6PredefinedRoles = map[string]struct{}{
```
ashishshinde, line 55: I think Roles should be all roles including 6.0 roles. Validation should fail if version < 6.0 and post6.0 role is configured. Simpler is code to use Roles map as being the complete set in other code post validation instead of looking at 2 arrays and possibly more in the furure.

api/v1beta1/access_control_validate.go
```diff
@@ -68,6 +75,12 @@ var Privileges = map[string][]PrivilegeScope{
 	"user-admin":     {Global},
 }
 
+var Post6Privileges = map[string][]PrivilegeScope{
```
ashishshinde, line 78: I think Privileges should be all privileges including 6.0 privileges. Validation should fail if version < 6.0 and post6.0 privilege is configured. Simpler is code to use Privileges as being the complete set.

api/v1beta1/access_control_validate.go
ashishshinde: See the previous comment on making Privileges as the complete set.

api/v1beta1/access_control_validate.go
```diff
@@ -68,6 +75,12 @@ var Privileges = map[string][]PrivilegeScope{
...
+	"truncate":      {Global, NamespaceSet},
+	"sindex-admin ": {Global},
```
ashishshinde, line 80: extra space here

api/v1beta1/access_control_validate.go
ashishshinde: Fixed

controllers/scripts/create_pod_status_patch.py
davi17g: Let's leave it to be separate functions in case we would like to make wiping to be asynchronous.

test/storage_wipe_test.go
davi17g: yes, because blkdiscard is not supported on AWS.

test/storage_wipe_test.go
davi17g: rcak ID: 2 has its own storage configuration so it can be easily configured to use other storage policy configurations if needed - the following test just focuses on testing the wiping process.

PR #133, opened 2022-05-25
Adding Kubernetes events support and events for various operations. #KO-48

tanmayja:


controllers/reconciler.go
```diff
@@ -99,12 +114,16 @@ func (r *SingleClusterReconciler) Reconcile() (ctrl.Result, error) {
...
+		r.Recorder.Event(r.aeroCluster, corev1.EventTypeWarning, "ACLUpdateFailed",
+			fmt.Sprintf("Failed to setup Access Control %s/%s", r.aeroCluster.Namespace, r.aeroCluster.Name))
 		return reconcile.Result{}, err
 	}
 
 	// Update the AerospikeCluster status.
 	if err := r.updateStatus(); err != nil {
 		r.Log.Error(err, "Failed to update AerospikeCluster status")
+		r.Recorder.Event(r.aeroCluster, corev1.EventTypeWarning, "StatusUpdateFailed",
+			fmt.Sprintf("Failed to update AerospikeCluster status %s/%s", r.aeroCluster.Namespace, r.aeroCluster.Name))
```
sud82, line 126: Check the above events. Use either cluster or AerospikeCluster

controllers/statefulset.go
```diff
@@ -410,7 +411,15 @@ func (r *SingleClusterReconciler) buildSTSConfigMap(
 		return fmt.Errorf("failed to build config map data: %v", err)
 	}
 
-	// Replace config map data since we are supposed to create a new config map.
+	// Replace config map data if differs since we are supposed to create a new config map.
+	if reflect.DeepEqual(confMap.Data, configMapData) {
+		r.Log.Info(
+			"Already existed configmap is same as current configmap, so no change required",
```
sud82, line 417: This log may become very frequent.
I think we can remove this log and add the log only if configMap is getting updated.

sud82 (changes requested):

controllers/rack.go
```diff
@@ -472,6 +502,9 @@ func (r *SingleClusterReconciler) scaleDownRack(
 		"ScaleDown AerospikeCluster statefulset", "desiredSz", desiredSize,
 		"currentSz", *found.Spec.Replicas,
 	)
+	r.Recorder.Event(r.aeroCluster, corev1.EventTypeNormal, "RackScaleDown",
```
sud82, line 505: It's better to place this event and the above log when scale down is needed. Basically after this condition
`if *found.Spec.Replicas <= desiredSize {
		return found, reconcileSuccess()
	}`

controllers/reconciler.go
```diff
@@ -51,7 +53,12 @@ func (r *SingleClusterReconciler) Reconcile() (ctrl.Result, error) {
 		r.Log.V(1).Info("Deleting AerospikeCluster")
 		// The cluster is being deleted
 		if err := r.handleClusterDeletion(finalizerName); err != nil {
+			r.Recorder.Event(r.aeroCluster, corev1.EventTypeWarning, "DeleteFailed",
+				fmt.Sprintf("Unable to handle AerospikeCluster delete operations %s/%s", r.aeroCluster.Namespace, r.aeroCluster.Name))
 			return reconcile.Result{}, err
+		} else {
+			r.Recorder.Event(r.aeroCluster, corev1.EventTypeNormal, "Deleted",
+				fmt.Sprintf("Deleted cluster %s/%s", r.aeroCluster.Namespace, r.aeroCluster.Name))
```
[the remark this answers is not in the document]
tanmayja: using AerospikeCluster now

sud82 (approved):

ashishshinde (approved):

PR #136, opened 2022-06-11
Support hostPath in VolumeSource.

skolosov-snap:
Support hostPath volume type in VolumeSource.

### DIFF
--- a/api/v1beta1/aerospikecluster_types.go
+++ b/api/v1beta1/aerospikecluster_types.go
@@ -506,6 +506,11 @@ type VolumeSource struct {
 	// +optional
 	ConfigMap *corev1.ConfigMapVolumeSource `json:"configMap,omitempty" protobuf:"bytes,19,opt,name=configMap"`
 
+	// HostPath represents a hostPath that should populate this volume.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#hostpath
+	// +optional
+	HostPath *corev1.HostPathVolumeSource `json:"hostPath,omitempty" protobuf:"bytes,20,opt,name=hostPath"`
+
 	// +optional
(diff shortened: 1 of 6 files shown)


ashishshinde:
@skolosov-snap Do you have a use case for using hostPath with Aerospike? Or is this driven by some sidecar or init-container needing access to hostPath? We had left this out because of the inherent security risk and lack of a valid use case.

skolosov-snap:
> @skolosov-snap Do you have a use case for using hostPath with Aerospike? Or is this driven by some sidecar or init-container needing access to hostPath? We had left this out because of the inherent security risk and lack of a valid use case.

Yes. We got rid of using EBS for storing SMD because EBS nvme devices messed up with with local SSD after rebooting. Thus we start getting an issue with using EBS. We tried to fix it by linking PVs (in Local Volume Provisioner) to devices using links based on device serial umbers. But for some reason it didn't work. And event for correctly created links Pods still sees naming which was before node restarts. We are discussing it with AWS guy right now.
In addition to that it seems us correct to tire SDM to data on the same node (local ssd). If Pod will be rescheduled to another node SDM will be migrated to that node, but data will not be.

unix-way:
I've also been working on this. Your PR does not cover the use case of mounting hostPaths as block devices. Mine does https://github.com/aerospike/aerospike-kubernetes-operator/pull/137

skolosov-snap:
> I've also been working on this. Your PR does not cover the use case of mounting hostPaths as block devices. Mine does #137

Thanks, Kirill. Yes, we haven't supported DeviceBlock mode for hostPath. I'm happy to use your PR instead.

sud82:
Duplicate of #137

PR #137, opened 2022-06-14
Add hostPath volumeSource

unix-way:


unix-way:
The goal of this PR is to allow mounting directories and block devices from the host system as hostPaths. The primary reasoning behind this was to mount NVMe devices in environments where all Kubernetes nodes have identical NVMe configurations (e.g. GKE) as BlockDevice hostPaths. This makes it possible to create configurations where pods can freely migrate among nodes when necessary, for example during node auto upgrades or maintenances. BlockDevice hostPaths can only be used in privileged mode.

skolosov-snap (commented):
Any reasons why you haven't supported hostPath in [device initialization logic](https://github.com/aerospike/aerospike-kubernetes-operator/blob/9386e48ca7548b77a770b6e6534d5f69fb749f8a/controllers/scripts/create_pod_status_patch.py#L66-L107).

unix-way:
@skolosov-snap Mainly lack of time and a little bit of oversight. It's added now.

unix-way:
This does not support other hostPath types though, i.e. FileOrCreate, File, Socket and CharDevice. I can add support for them, or we can do it later as needed.

skolosov-snap (approved):

unix-way:
@sud82 @abhishekdwivedi3060 @tanmayja Is there any chance for this to be reviewed? Thank you!

sud82:
@unix-way, we will review this and let you know.

sud82:
Hi @unix-way, we have looked at it briefly. There are a few concerns related to security, aerospike behavior, and your use case. These things will need a deeper look and maybe a discussion related to your use case. Unfortunately, it will be difficult to give time to this at present because of our immediate priorities related to the next release. We will review this issue when we are done with our current release.

abhishekdwivedi3060:
@unix-way @skolosov-snap 
We've a few concerns around supporting hostPath volumes for storing data:
1. As per Kubernetes security practices, hostPath volumes (if used) should be mounted ReadOnly. This won’t work for your use-case as you want to store Aerospike data in hostPath volumes
2. HostPath poses security threat and mentioned in [k8s doc](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath)
3. Requires root with privileged container to write file/dir created as part of hostPath Volume: ReadOnly won’t work for us as we need to write data
HostPath poses security threat and mentioned in k8s doc: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath
Requires root with privileged container to write file/dir created as part of hostPath [https://kubernetes.io/docs/concepts/storage/volumes/#hostpath:~:text=The%20files%20or%20directories[…]able%20to%20write%20to%20a%20hostPath%20volume](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath:~:text=The%20files%20or%20directories%20created%20on%20the%20underlying%20hosts%20are%20only%20writable%20by%20root.%20You%20either%20need%20to%20run%20your%20process%20as%20root%20in%20a%20privileged%20Container%20or%20modify%20the%20file%20permissions%20on%20the%20host%20to%20be%20able%20to%20write%20to%20a%20hostPath%20volume) 
4. Aerospike Cluster with MultiPodPerHost set won't work as multiple pods (scheduled on same node) will try to use the same underlying hostPath volume on the same host. This will result in data overwrite or data conflict.
5. There will be unnecessary data migration in case of Aerospike cluster pod Restart as pod might get scheduled on to a new node. Kubernetes scheduler doesn't take care of scheduling the restarted pod again on the same node for hostPath Volumes.


For using Local storage of the k8s nodes there are other different ways possible which overcomes some of the above limitations:
1. Local Provisioner: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner
2. OpenEBS: https://openebs.io/

Let us know if above mentioned Local storage options solve your use-case.

unix-way:
@abhishekdwivedi3060 thank you for your response!

As for security concerns, i.e. points 1-3, I guess that's up to a database administrator to decide whether it's a security concern or not. There are plenty of cases where hostPaths are used in read-write mode and containers are privileged, and with a few precautions such systems are secure and reliable, especially in local and sole-tenant clusters.

As for point 4, yes, this will only work in one pod per node configurations.

Point 5 is the exact reason why I thought about using hostPaths for mounting NVMe devices at all. PVC-based solutions, such as Local Provisioner or OpenEBS, create PVs with node affinity, which prevents automatic cluster upgrades or node replacement in case of errors. With hostPaths, pods can freely migrate to new nodes with identical configurations, and disruptions can be controlled with PDBs. Such events do not happen often, so infrequent migrations can be tolerated for the sake of flexibility.

abhishekdwivedi3060:
Thanks for your inputs @unix-way 
Regarding point 5, I agree cluster upgrade or node replacement are not that frequent. But during the normal lifecycle of Aerospike Clusters, pods are restarted (Aerospike Cold Restart) multiple times to bring into effect the latest changes made to Aerospike Cluster CR like version upgrade, pod spec change etc. These scenarios are pretty common and happen frequently. In these scenarios, there will be unnecessary data migration as pointed out in my previous comment.

Also, if a  pod is scheduled on to a new K8s node during cluster maintenance and that K8s node has old Aerospike data left on hostPath by some old Aerospike pods then that data might become part of Aerospike cluster there by impacting the integrity of cluster data.

unix-way:
@sud82 Do I get it right that you have decided not to implement this feature?

@abhishekdwivedi3060 Thanks for the answers! Since Aerospike is running as a stateful set, and with one pod per node in this case, after a restart a pod should be scheduled on the same node. As for leftover data on NVMe devices, this is not an issue on new nodes with fresh drives. If a old node is reused and an NVMe device indeed contains some stale data, the TPM should resolve the conflict based on either generation count or LUT.

sud82:
Hi @unix-way, 

- Looks like in the case of hostPath there is no guarantee that the pod will be rescheduled on the same k8s node. https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/#how-is-it-different-from-a-hostpath-volume
- If there is any leftover Aerospike data then Aerospike clearly recommends wiping the entire data. https://docs.aerospike.com/server/operations/configure/storage/ssd_init

sud82:
Hi @unix-way, You are right that pod will not migrate to another node until it's recreated. But, the warm restart happens only in case of an aerospike config change. For any other operation like rolling restart or upgrade, the pod is recreated therefore making it open for rescheduling in other nodes.

It may be possible to identify pod migration (I haven't looked at it yet). But, device cleanup can take a significant amount of time, that too during simple upgrades and rolling restarts.

I think the challenges of this feature/use case outweigh its possible benefits. 
Many of our customers use local storage and I agree that it has its own challenges. But we are trying to document and automate whatever can be done to reduce all those frictions. 

I would recommend you rethink using local storage
Local Provisioner: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner
OpenEBS: https://openebs.io/

unix-way:
Thank you! We have found a way to tailor local provisioner to our needs. Closing PR.

PR #142, opened 2022-07-22
[OPS-5904] Allow external volumes to be injected by other controllers/operators

tanmayja:
Adding the support to mount additional volumes to the pod.


test/statefulset_storage_test.go
```diff
@@ -0,0 +1,202 @@
...
+								HostPath: &v1.HostPathVolumeSource{
+									Path: "/etc/localtime",
+								},
+							},
+						})
+
+						sts.Spec.Template.Spec.Containers[0].VolumeMounts = mountedVolumes
+						sts.Spec.Template.Spec.Volumes = volumes
+
+						err = updateSTS(k8sClient, ctx, sts)
```
ashishshinde, line 86: AKO also updates the STS in reconcile. Are you sure it hasn't removed the tzdata volume that was added by this update?

tanmayja: Modified the logic such that external volumes won't be affected with AKO reconciliation

controllers/statefulset.go
```diff
@@ -1295,29 +1338,41 @@ func (r *SingleClusterReconciler) initializeSTSStorage(
...
+		externalDevices, volumesForDevice := r.getExternalStorageDevices(&st.Spec.Template.Spec.Containers[i], rackState, st.Spec.Template.Spec.Volumes)
+		specVolumes = append(specVolumes, volumesForDevice...)
 		st.Spec.Template.Spec.Containers[i].VolumeDevices = []corev1.VolumeDevice{}
+		st.Spec.Template.Spec.Containers[i].VolumeDevices = append(st.Spec.Template.Spec.Containers[i].VolumeDevices, externalDevices...)
 	}
 
 	st.Spec.Template.Spec.Volumes = getDefaultSTSVolumes(
 		r.aeroCluster, rackState,
 	)
+	st.Spec.Template.Spec.Volumes = append(st.Spec.Template.Spec.Volumes, specVolumes...)
```
sud82, line 1375: Here also, append may lead to duplicate entries. the same volume can be mounted in initContainer and regular container.

sud82 (changes requested):

controllers/statefulset.go
```diff
@@ -1295,29 +1338,41 @@ func (r *SingleClusterReconciler) initializeSTSStorage(
 	rackState RackState,
 ) {
 	// Initialize sts storage
+	var specVolumes []corev1.Volume
 	for i := range st.Spec.Template.Spec.InitContainers {
+		externalMounts, volumesForMount := r.getExternalStorageMounts(&st.Spec.Template.Spec.InitContainers[i], rackState, st.Spec.Template.Spec.Volumes)
+		specVolumes = append(specVolumes, volumesForMount...)
```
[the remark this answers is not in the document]
tanmayja: Made changes, such that only unique volume will be added in specVolumes.

controllers/statefulset.go
tanmayja: Made changes to handle this

test/statefulset_storage_test.go
```diff
@@ -0,0 +1,490 @@
...
+var _ = Describe(
+	"STSStorage", func() {
+		ctx := goctx.TODO()
+
+		Context(
+			"When doing valid operations", func() {
+
+				BeforeEach(
+					func() {
+						clusterName := "sts-storage"
```
sud82, line 26: Make it global, at the context level. No need to repeat in every `IT`
`clusterName := "sts-storage"
						clusterNamespacedName := getClusterNamespacedName(
							clusterName, namespace,
						)`

test/statefulset_storage_test.go
```diff
@@ -0,0 +1,490 @@
...
+		getNamespacedNameForSTS(aeroCluster, rackID),
+		found,
+	)
+	if err != nil {
+		return nil, err
+	}
+	return found, nil
+}
+
+func validateExternalVolumeInContainer(sts *appsv1.StatefulSet, index int) (
```
sud82, line 419: validateExternalVolumeInContainer, validateExternalVolumeInInitContainer 
can be merged into a single function, take a flag as param to distinguish container or initContainer

sud82 (approved):

ashishshinde (approved):

PR #154, opened 2022-10-08
[KO-45] Allow batch upgrade and restart of pods in a given rack 

sud82:


api/v1beta1/aerospikecluster_types.go
```diff
@@ -315,6 +315,12 @@ type RackConfig struct {
 	// Racks is the list of all racks
 	// +nullable
 	Racks []Rack `json:"racks,omitempty"`
+
+	// RestartPercentage is the percentage of rack nodes that will be restarted simultaneously
+	RestartPercentage int `json:"restartPercentage,omitempty"`
+
+	// RestartNodesCount is number of nodes that will be restarted simultaneously
+	RestartNodesCount int `json:"restartNodesCount,omitempty"`
```
abhishekdwivedi3060, line 323: Does it make sense to combine these two fields and handle both exact count and percentage?
K8s Deployment has similar field for RollingUpdate which can take both exact count or percentage and based on the dataType, further processing is done.

```
// The maximum number of pods that can be unavailable during the update.
	// Value can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%).
	// Absolute number is calculated from percentage by rounding down.
	// This can not be 0 if MaxSurge is 0.
	// Defaults to 25%.
	// Example: when this is set to 30%, the old ReplicaSet can be scaled down to 70% of desired pods
	// immediately when the rolling update starts. Once new pods are ready, old ReplicaSet
	// can be scaled down further, followed by scaling up the new ReplicaSet, ensuring
	// that the total number of pods available at all times during the update is at
	// least 70% of desired pods.
	// +optional
	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty" protobuf:"bytes,1,opt,name=maxUnavailable"`
```

controllers/rack.go
```diff
@@ -702,31 +685,46 @@ func (r *SingleClusterReconciler) rollingRestartRack(
...
-				pod.Name,
-			)
+			r.Log.Info("This Pod doesn't need rolling restart, Skip this", "pod", pod.Name)
 			continue
 		}
+		podsToRestart = append(podsToRestart, &pod)
+	}
 
-		res := r.rollingRestartPod(rackState, pod, restartType, ignorablePods)
+	// Failed pods should always come before active pods.
```
tanmayja, line 706: Can we batch failed pods and active pods separately and not mix them in single batch? doing this we can make sure failed pods batch gets handled first.

controllers/rack.go
```diff
@@ -702,31 +685,46 @@ func (r *SingleClusterReconciler) rollingRestartRack(
...
+	// We make info calls before restarting active pods,
+	// if active pods are checked before failed pods then
```
tanmayja, line 708: In rollingRestartPods, we are anyway restarting failed pods first. Is it really required to arrange failed and active pods like this?

controllers/rack.go
```diff
@@ -702,31 +685,46 @@ func (r *SingleClusterReconciler) rollingRestartRack(
...
+	// info call will fail in failed pods and operator will not go ahead
+	rearrangedPods := getRearrangedFailedAndActivePods(podsToRestart)
+
+	// Create batch of pods
+	podsBatchList := r.getPodsBatchToRestart(rearrangedPods, len(podList))
+
+	// Restart batch of pods
+	for _, podsBatch := range podsBatchList {
```
tanmayja, line 716: this for loop is never going to hit more than once as, it has returned in every scenario.

tanmayja (changes requested):

controllers/pod.go
```diff
@@ -185,155 +160,227 @@ func (r *SingleClusterReconciler) quickRestart(
...
+				"Pods are running and ready", "pods",
+				podNames,
 			)
-			time.Sleep(time.Second * 5)
+			return reconcileSuccess()
+		}
+		time.Sleep(retryInterval)
+	}
+
+	r.Log.Info(
```
tanmayja, line 255: print what all pods were unable to come online and timeout happened.

controllers/pod.go
```diff
@@ -185,155 +160,227 @@ func (r *SingleClusterReconciler) quickRestart(
...
+}
+
+func (r *SingleClusterReconciler) safelyDeletePodsAndEnsureImageUpdated(
+	rackState RackState, podsToUpdate []*corev1.Pod, ignorablePods []corev1.Pod,
+) reconcileResult {
 
-		r.Log.Info("Pod is restarted", "podName", updatedPod.Name)
-		started = true
-		break
+	failedPods, activePods := getFailedAndActivePods(podsToUpdate)
```
tanmayja, line 289: can we design in such a way that getFailedAndActivePods will not be called again as we already using it in rearranging failed and active pods before this step.

controllers/pod.go
```diff
@@ -185,155 +160,227 @@ func (r *SingleClusterReconciler) quickRestart(
...
 	}
-	r.Log.V(1).Info("Pod deleted", "podName", p.Name)
 
-	r.Recorder.Eventf(r.aeroCluster, corev1.EventTypeNormal, "PodWaitUpdate",
-		"[rack-%d] Waiting to update Pod %s", rackState.Rack.ID, p.Name)
-	// Wait for pod to come up
+	return r.ensurePodsImageUpdated(podsToUpdate)
+}
+
+func (r *SingleClusterReconciler) ensurePodsImageUpdated(podsToCheck []*corev1.Pod) reconcileResult {
```
tanmayja, line 329: ensurePodsImageUpdated and ensurePodsRunningAndReady seems similar with a check change. Do you think these can be merge and use that by passing an argument to differentiate?

tanmayja (changes requested):

controllers/pod.go
sud82: We do not check for all pods. Whenever we get an unready pod, we go out. To print all unready pods, we would need to check all pods.

abhishekdwivedi3060 (approved):

PR #163, opened 2022-11-25
Zeroize header (first 8MiB) when initMethod or wipeMethod is set to blkdiscard

spkesan:
- Aerospike 6+ will not startup if the first 8MiB on device are not zeroed to protect from accidentally using an unintended volume. This is applicable when using fresh block device.
- This PR enables init script to zeroize the header (first `8MiB`) when `initMethod` or `wipeMethod` are set to `blkdiscard`.
- This is also backward compatible with previous versions of Aerospike.

Reference:
- https://docs.aerospike.com/server/operations/plan/ssd/ssd_init
- https://docs.aerospike.com/server/operations/troubleshoot/startup#the-header-has-not-been-zeroized

### DIFF
--- a/controllers/scripts/create_pod_status_patch.py
+++ b/controllers/scripts/create_pod_status_patch.py
@@ -399,9 +399,9 @@ def init_volumes(pod_name, config):
 
             elif volume.effective_init_method == "blkdiscard":
 
-                blkdiskard = "blkdiscard {volume_path}".format(
+                blkdiscard_cmd = "blkdiscard {volume_path} && blkdiscard -z --length 8MiB {volume_path}".format(
                     volume_path=volume.get_mount_point())
-                execute(blkdiskard)
+                execute(blkdiscard_cmd)
                 logging.info(f"{volume} - Initialized")
 
             elif volume.effective_init_method == "none":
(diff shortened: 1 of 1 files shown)


ashishshinde:
@spkesan Aerospike requires the entire disk to be zeroed. See the notes on using [blkdiscard](
https://docs.aerospike.com/cloud/kubernetes/operator/Cluster-configuration-settings#initmethod). 

You should use blkdiscard `only` when you are sure it will effectively zeroize the disk. So this change is moot.
If blkdiscard does not work, zeroizing the header also does not help. Aerospike will fail at startup if the disk is not effectively zeroed, even if the disk header is zeroized.

kportertx:
Actually, `blkdiscard -z` will always zeroize - see the man page.

kportertx:
Zeroizing the header should be enough to get the server to start. Running blkdiscard on the rest invokes TRIM which will let the firmware know that those blocks are not in use.

ashishshinde:
The server does start and then fails if it find non zero bits even after header. So this is moot. 
See the docs link I posted where we recommend blkdiscard only when the user is sure that it will effectively zeroize the disk as in it supports RZAT.

kportertx:
We only check at startup and the check is only on the header.

PR #168, opened 2022-12-23
KO-36: Added support for automatic migrate-fill-delay reset during scale down

abhishekdwivedi3060:
- Added support for automatic migrate-fill-delay reset during scale down
- Fixed some go-lint issues

TODO: Update `aerospike/aerospike-management-lib` module version in go.mod when corresponding PR Is merged


controllers/reconciler.go
```diff
@@ -132,6 +133,14 @@ func (r *SingleClusterReconciler) Reconcile() (ctrl.Result, error) {
 		return reconcile.Result{}, err
 	}
 
+	// revert migrate-fill-delay to original value if it was set to 0 during scale down
+	// Passing first rack from the list as all the racks will have same migrate-fill-delay
+	if res := r.setMigrateFillDelay(r.getClientPolicyFromSpec(), r.aeroCluster.Spec.RackConfig.Racks[0],
```
tanmayja, line 138: Instead of rack, send config only, as there is no other use of rack object in function definition.

controllers/rack.go
```diff
@@ -620,6 +620,12 @@ func (r *SingleClusterReconciler) scaleDownRack(
 		}
 	}
 
+	// set migrate-fill-delay to 0 across all nodes of cluster to scale down fast
+	// temporarily add the pod to be deleted in ignorablePods slice to avoid setting migrate-fill-delay to pod when pod is not ready
+	if res := r.setMigrateFillDelay(r.getClientPolicy(), rackState.Rack, false, append(ignorablePods, *pod)); !res.isSuccess {
```
tanmayja, line 625: is there any reason of using r.getClientPolicy() while clearing migrate-fill-delay and r.getClientPolicyFromSpec() while setting migrate-fill-delay back to original value?

tanmayja (changes requested):

controllers/rack.go
```diff
@@ -620,6 +620,12 @@ func (r *SingleClusterReconciler) scaleDownRack(
 		}
 	}
 
+	// set migrate-fill-delay to 0 across all nodes of cluster to scale down fast
+	// temporarily add the pod to be deleted in ignorablePods slice to avoid setting migrate-fill-delay to pod when pod is not ready
```
tanmayja, line 624: Have we faced any scenario that pod is not ready at this point of time? Because we are reducing the replicas after this step.

controllers/rack.go
tanmayja: I think it is because of reconcileAccessControl() being called in later point of time, nevermind.

controllers/rack.go
abhishekdwivedi3060: This is only for the case when the pod which we are going to delete is not in running state. In that case, we skip the `waitForMultipleNodesSafeStopReady` check and then reduce the replica

controllers/rack.go
abhishekdwivedi3060: For the first time we need user from the spec which was created with sys-admin role because at this point AccessControl is not updated in the status. Later when Scaledown is triggered, we use AccessControl from the status to use the already created user as new AccessControl is yet to be reconciled

controllers/reconciler.go
abhishekdwivedi3060: Done

tanmayja (approved):

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -423,7 +424,7 @@ func (c *AerospikeCluster) validateResourceAndLimits(resources *v1.ResourceRequi
 	return nil
 }
 
-func (c *AerospikeCluster) validateRackConfig(aslog logr.Logger) error {
+func (c *AerospikeCluster) validateRackConfig(_ logr.Logger) error {
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Can be removed. But left it as it is to keep in sync with other funcs and for future ref

sud82 (approved):

PR #173, opened 2023-01-09
Handling index removal from namespace

tanmayja:
When a mount point from a namespace index has been removed, that namespace has to go to cold restart by server to re-build the index for the same. So in operator when a mount point from index will get removed, operator will delete all the files from all mount points in that particular namespace.


controllers/rack.go
```diff
@@ -540,6 +540,12 @@ func (r *SingleClusterReconciler) upgradeRack(
 
 		podNames := getPodNames(podsBatch)
 
+		// For each block volume removed from a namespace, pod status dirtyVolumes is appended with that volume name.
```
sud82, line 543: I think we should put this function inside deletePodAndEnsureImageUpdated function, which is called after checking for migrations, quiesce.

controllers/rack.go
```diff
@@ -747,16 +747,17 @@ func (r *SingleClusterReconciler) rollingRestartRack(
...
 		podsBatch := podsBatchList[0]
 
-		res := r.rollingRestartPods(rackState, podsBatch, ignorablePods, restartTypeMap)
-		if !res.isSuccess {
-			return found, res
-		}
-
 		// For each block volume removed from a namespace, pod status dirtyVolumes is appended with that volume name.
 		// For each file removed from a namespace, it is deleted right away.
 		if err := r.handleNSOrDeviceRemoval(rackState, podsBatch); err != nil {
```
sud82, line 752: This we can put inside `restartPods()` function

sud82 (changes requested):

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -336,10 +337,17 @@ func (c *AerospikeCluster) validateRackUpdate(
 				}
 
 				if len(oldRack.AerospikeConfig.Value) != 0 || len(newRack.AerospikeConfig.Value) != 0 {
+					var statusRack Rack
```
sud82, line 340: statusRack will become the last rack from c.Status.RackConfig.Racks.

controllers/rack.go
tanmayja: done

api/v1beta1/aerospikecluster_validating_webhook.go
tanmayja: fixed

abhishekdwivedi3060 (approved):

sud82 (approved):

PR #174, opened 2023-01-10
KO-149: Added support for custom kubernetes cluster domain and dnsConfig

abhishekdwivedi3060:


controllers/statefulset.go
```diff
@@ -1132,6 +1132,9 @@ func (r *SingleClusterReconciler) updateSTSFromPodSpec(
 	st.Spec.Template.ObjectMeta.Annotations = r.aeroCluster.Spec.PodSpec.AerospikeObjectMeta.Annotations
 
 	st.Spec.Template.Spec.DNSPolicy = r.aeroCluster.Spec.PodSpec.DNSPolicy
+	if r.aeroCluster.Spec.PodSpec.DNSConfig != nil {
```
sud82, line 1135: Due to this `if check`, the user will not be able to reset the DNS config to nil.

test/cluster_test.go
```diff
@@ -967,6 +991,36 @@ func negativeUpdateClusterValidationTest(
...
+				"InvalidDNSConfiguration", func() {
+					It(
+						"InvalidDnsPolicy: should fail when dnsPolicy is set to 'Default'",
+						func() {
+							aeroCluster, err := getCluster(
+								k8sClient, ctx, clusterNamespacedName,
+							)
+							Expect(err).ToNot(HaveOccurred())
+							aeroCluster.Spec.PodSpec.DNSPolicy = v1.DNSDefault
+							err = deployCluster(k8sClient, ctx, aeroCluster)
```
sud82, line 1004: Update cluster

test/cluster_test.go
```diff
@@ -967,6 +991,36 @@ func negativeUpdateClusterValidationTest(
...
+
+					It(
+						"MissingDnsConfig: Should fail when dnsPolicy is set to 'None' and no dnsConfig given",
+						func() {
+							aeroCluster, err := getCluster(
+								k8sClient, ctx, clusterNamespacedName,
+							)
+							Expect(err).ToNot(HaveOccurred())
+							aeroCluster.Spec.PodSpec.DNSPolicy = v1.DNSNone
+							err = deployCluster(k8sClient, ctx, aeroCluster)
```
sud82, line 1017: Update cluster

test/cluster_test.go
```diff
@@ -894,6 +894,30 @@ func negativeDeployClusterValidationTest(
...
+						func() {
+							aeroCluster := createDummyAerospikeCluster(clusterNamespacedName, 1)
+							aeroCluster.Spec.PodSpec.DNSPolicy = v1.DNSDefault
+							err := deployCluster(k8sClient, ctx, aeroCluster)
+							Expect(err).Should(HaveOccurred())
+						},
+					)
+
+					It(
+						"MissingDnsConfig: should fail when dnsPolicy is set to 'None' and no dnsConfig given",
```
sud82, line 911: Add a positive test case where dnsPolicy is `none` and dnsConfig is given

sud82 (changes requested):

controllers/statefulset.go
tanmayja: can we make this check over DNSPolicy?

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -1607,3 +1611,15 @@ func ValidateAerospikeObjectMeta(aerospikeObjectMeta *AerospikeObjectMeta) error
 	}
 	return nil
 }
+
+func validateDNS(dnsPolicy v1.DNSPolicy, dnsConfig *v1.PodDNSConfig) error {
+	if dnsPolicy == v1.DNSDefault {
```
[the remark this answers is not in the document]
abhishekdwivedi3060: `Default` policy is not going to work out of the box for us as we are dependent on cluster internal DNS lookup.
It might be possible only if coreDNS related configuration is set across all nodes, which is very unlikely.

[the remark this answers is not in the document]
abhishekdwivedi3060: `Default` policy is not going to work out of the box for us as we are dependent on cluster internal DNS lookup.
It might be possible only if coreDNS related configuration is set across all nodes, which is very unlikely.

test/cluster_test.go
abhishekdwivedi3060: Done

controllers/statefulset.go
abhishekdwivedi3060: Removed if condition to handle both create and update scenario

tanmayja (approved):

PR #176, opened 2023-01-11
Bugfix 2.4.0

sud82:


controllers/reconciler.go
```diff
@@ -333,6 +348,41 @@ func (r *SingleClusterReconciler) updateStatus() error {
...
+		return err
+	}
+
+	// AerospikeAccessControl
+	statusAerospikeAccessControl := &asdbv1beta1.AerospikeAccessControlSpec{}
+	lib.DeepCopy(
+		statusAerospikeAccessControl, r.aeroCluster.Spec.AerospikeAccessControl,
+	)
+
+	newAeroCluster.Status.AerospikeClusterStatusSpec.AerospikeAccessControl = statusAerospikeAccessControl
```
abhishekdwivedi3060, line 374: Can we add a check here to update the `AerospikeAccessControl` only when there is a diff between spec and status. This avoid patchStatus calls

controllers/reconciler.go
```diff
@@ -669,28 +719,25 @@ func (r *SingleClusterReconciler) checkPreviouslyFailedCluster() error {
...
+		return nil, err
+	}
 
-	for _, rackStatus := range r.aeroCluster.Status.RackConfig.Racks {
-		for _, statusNamespace := range rackStatus.AerospikeConfig.Value["namespaces"].([]interface{}) {
-			statusNamespaces[statusNamespace.(map[string]interface{})["name"].(string)] = true
-		}
+	statusNamespaces := sets.NewString()
+	for _, namespaces := range nodesNamespaces {
+		statusNamespaces.Insert(namespaces...)
```
abhishekdwivedi3060, line 730:
```
	statusNamespaces := sets.NewString(nodesNamespaces)

```

abhishekdwivedi3060 (changes requested):

PR #180, opened 2023-01-18
[KO-157] Allow user to update only config file without any rolling restart after making dynamic config change.

tanmayja:


sud82:
This is on hold, I will reopen it in the future.

PR #182, opened 2023-01-23
Fixing a race condition in reconciliation flow.

tanmayja:


test/cluster_helper.go
```diff
@@ -642,19 +627,14 @@ func updateSTS(
 }
 
 // TODO: remove it
```
abhishekdwivedi3060, line 629: We can remove the TODO now

abhishekdwivedi3060 (approved):

test/cluster_helper.go
tanmayja: done

sud82 (approved):

PR #183, opened 2023-01-31
KO-164: Optimised migrate-fill-delay scale down flow

abhishekdwivedi3060:
- Optimised migrate-fill-delay scale down flow


controllers/rack.go
```diff
@@ -301,6 +299,16 @@ func (r *SingleClusterReconciler) reconcileRack(
 		}
 	}
 
+	// revert migrate-fill-delay to original value if it was set to 0 during scale down
+	// Reset will be done only in case of Scale down
+	if r.aeroCluster.Status.Size > r.aeroCluster.Spec.Size {
```
sud82, line 304: This will miss one use case, where the whole cluster size is not changing but rack distribution is changing due to addition or deletion of a rack

sud82 (changes requested):

controllers/rack.go
abhishekdwivedi3060: Yeah, that use-case is handled by the fall back logic of revert migrate-fill-delay in the main reconcile loop.

sud82 (approved):

PR #187, opened 2023-02-08
Fix golint issues

sud82:


api/v1beta1/aerospikecluster_mutating_webhook.go
```diff
@@ -189,23 +190,29 @@ func (c *AerospikeCluster) updateRacksPodSpecFromGlobal(asLog logr.Logger) error
...
 	}
-	return nil
 }
 
 func (c *AerospikeCluster) updateRacksAerospikeConfigFromGlobal(asLog logr.Logger) error {
-	for i, rack := range c.Spec.RackConfig.Racks {
-		var m map[string]interface{}
-		var err error
+	for i := range c.Spec.RackConfig.Racks {
+		rack := c.Spec.RackConfig.Racks[i]
```
tanmayja, line 197: Reason why we have done this is because we were copying n number of bytes unnecessary. Is it really helping in that. Lint issue is fix but we are still creating a rack var of size n bytes.

tanmayja (changes requested):

controllers/aero_info_calls.go
```diff
@@ -227,49 +209,51 @@ func (r *SingleClusterReconciler) newAllHostConnWithOption(ignorablePods []corev
...
 			return nil, err
 		}
+
 		hostConns = append(hostConns, hostConn)
 	}
 
 	return hostConns, nil
 }
 
+//nolint:unparam // todo: add error
```
abhishekdwivedi3060, line 246: Shouldn't we remove error return type instead of adding ?

controllers/reconciler.go
```diff
@@ -651,19 +677,27 @@ func (r *SingleClusterReconciler) deleteExternalResources() error {
...
 	}
 
-	// removePVCs should be passed only filtered pvc otherwise rack pvc may be removed using global storage cascadeDelete
+	// removePVCs should be passed only filtered pvc otherwise rack pvc may be removed using global storage
+	// cascadeDelete
 	var filteredPVCItems []corev1.PersistentVolumeClaim
-	for _, pvc := range pvcItems {
+
+	for pvcIdx := range pvcItems {
+		pvc := pvcItems[pvcIdx]
```
abhishekdwivedi3060, line 685: nit: Can use pointer here and only dereference it if !found

controllers/statefulset.go
```diff
@@ -310,18 +322,22 @@ func (r *SingleClusterReconciler) waitForSTSToBeReady(st *appsv1.StatefulSet) er
 			context.TODO(),
 			types.NamespacedName{Name: st.Name, Namespace: st.Namespace}, st,
 		)
+
 		if err != nil {
```
abhishekdwivedi3060, line 326: Can be clubbed with above err var

abhishekdwivedi3060 (changes requested):

pkg/jsonpatch/jsonpatch.go
```diff
@@ -10,43 +10,48 @@ import (
 
 var errBadJSONDoc = fmt.Errorf("invalid JSON Document")
 
-type JsonPatchOperation struct {
+type PatchOperation struct {
```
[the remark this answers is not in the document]
sud82: It also has info about the type of operation. So we can append `operation`. 
We also use jsonpatch as the module name. So whoever imports this will have to use jsonpatch.JsonPatch() which doesn't look good.

controllers/access_control.go
```diff
@@ -625,62 +620,73 @@ func (userCreate AerospikeUserCreateUpdate) Execute(
...
+				aeroCluster, corev1.EventTypeWarning, "UserCreateFailed",
+				"Failed to Create User %s", userCreate.name,
+			)
 		}
+
 		return err
 	}
+
 	errorUpdate := userCreate.updateUser(client, adminPolicy, user, logger, recorder, aeroCluster)
 	if errorUpdate != nil {
```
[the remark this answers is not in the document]
sud82: errorUpdate is used in the return statement.

api/v1beta1/aerospikecluster_types.go
```diff
@@ -328,21 +344,23 @@ type RackConfig struct {
...
-type Rack struct {
+type Rack struct { //nolint:govet // for readability
 	// Identifier for the rack
 	ID int `json:"id"`
 	// Zone name for setting rack affinity. Rack pods will be deployed to given Zone
 	Zone string `json:"zone,omitempty"`
 	// Region name for setting rack affinity. Rack pods will be deployed to given Region
 	Region string `json:"region,omitempty"`
-	// RackLabel for setting rack affinity. Rack pods will be deployed in k8s nodes having rackLabel {aerospike.com/rack-label: <rack-label>}
+	// RackLabel for setting rack affinity. Rack pods will be deployed in k8s nodes having rackLabel {aerospike.
```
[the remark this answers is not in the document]
sud82: We can break them at logical points. Some comments are too long to be ignored.

abhishekdwivedi3060 (approved):

tanmayja (approved):

PR #191, opened 2023-02-14
chore: move pods rbac to aerospike-cluster helm chart

clemcvlcs:
## Context 

Using the `aerospike-kubernetes-operator` helm chart, by default, the user needs to specify a list of `watchNamespaces`, which is used by the operator as the environement variable `WATCH_NAMESPACE`. 

This implies the creation of rome rbac manifests, `Clusterrole`, `Clusterrolebinding` and `serviceAccount` that needs to be create in the target "watched" namespace. 

This process has a major drawback, which is the fact that the watched namespace needs to exist before the deployment/update of the operator. 

This process is really not convenient in a context where a user (like me :) ) needs to create multiple clusters in different namespaces, since for each new cluster we have to : 
- Create the target namespace
- Update the `watchNamespaces` in the `aerospike-kubernetes-operator` helm chart and redeploy the operator
- Deploy the `aerospike-cluster` helm chart in the target namespace. 

this definitely doesn't fits with most of the deployment process.

## This PR: 
- sets the `watchNamespace` value to `""` which will [make the operator running with cluster scope](https://github.com/aerospike/aerospike-kubernetes-operator/blob/792ab0702324cd2125ffae2109ffd07f608f662f/main.go#L188). 
- move the clusterroles rbac to the `aerospike-cluster`

This way, we don't need to touch the operator to deploy a new cluster. 

## Breaking changes

In the current state, an upgrade from chart 2.4.0 to this new version will break the operator, until the `aerospike-cluster` is also deployed. Not ideal at all

if for any reason someone want to deploy multiple helm releases of `aerospike-cluster` helm chart, he should only set the value `rbac.create: true` for the first release

## How Has This Been Tested?

Deployed on a local cluster with default values.

### DIFF
--- a/helm-charts/aerospike-cluster/templates/rbac.yaml
+++ b/helm-charts/aerospike-cluster/templates/rbac.yaml
@@ -1,17 +1,16 @@
-{{- range $key, $ns := (split "," .Values.watchNamespaces) }}
-
+{{- if .Values.rbac.create }}
 # Service account used by the cluster pods to obtain pod metadata.
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   # Do not change name, it's hard-coded in operator
   name: aerospike-operator-controller-manager
-  namespace: "{{ $ns }}"
+  namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ template "aerospike-kubernetes-operator.fullname" $ }}
-    chart: {{ $.Chart.Name }}
-    release: {{ $.Release.Name }}
-{{- with $.Values.imagePullSecrets }}
+    app: {{ template "aerospike-cluster.commonName" . }}
+    chart: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+{{- with .Values.imagePullSecrets }}
 imagePullSecrets:
   {{- range . }}
   - name: {{ . }}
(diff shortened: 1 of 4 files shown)


alon-dotan-starkware:
this will also help our use case of deploy multiple test envs on dynamic namespaces

sud82:
Hi @clemcvlcs, In this PR service account is being created for every cluster but we use the same service account for all the clusters in a single namespace. So, this will not work if there are multiple clusters in a single namespace.

We are planning to make the following changes in our RBAC

- Bifurcate RBAC for operator and cluster
  - Operator's clusterRole, clusterRoleBinding, and serviceAccount
  - Cluster's clusterRole, clusterRoleBindig/RoleBinding and serviceAccount
- Our Operator bundle/helm chart will create RBAC for only the operator and a clusterRole for the cluster will also be created
- For every k8s namespace user has to create serviceAccount and clusterRoleBinding/roleBinding using the above clusterRole
- We would not make the cluster's RBAC automatically. This way user can create either the clusterRoleBinding or the roleBinding and restrict the cluster's access to a k8s namespace only.

Even at present OLM bundle create RBAC only for the namespace where it has been deployed. The user has to create RBAC for other namespaces. To keep the helm in sync with OLM bundle deployment and to allow user to give restricted access to the cluster, we are planning to make the above changes.

Please let me know your thoughts.

PR #194, opened 2023-02-27
[KO-197] Remove python dependancy from init scripts

tanmayja:


init/pkg/set_host_port_env_util.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+
+			if add.Type == "InternalIP" {
+				nodeInternalIP = add.Address
+				continue
+			}
+
+			if add.Type == "ExternalIP" {
+				nodeExternalIP = add.Address
+				continue
```
abhishekdwivedi3060, line 116: `continue` seems redundant here

abhishekdwivedi3060 (changes requested):

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,842 @@
...
+				if _, err := os.Stat(volume.getMountPoint()); err != nil {
+					return dirtyVolumes, fmt.Errorf("mounting point %s does not exist %v", volume.getMountPoint(), err)
+				}
+
+				switch volume.effectiveWipeMethod {
+				case DDMethod:
+					stderr, err := os.CreateTemp("/tmp", "init-stderr")
+					if err != nil {
+						return dirtyVolumes, err
+					}
```
abhishekdwivedi3060, line 568: I think we can move this part inside go-routine itself

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,842 @@
...
+						wg.Done()
+
+						if err := execute(cmd, nil); err != nil {
+							panic(err.Error())
+						}
+
+						log.Println("INFO: Execution completed", "cmd", cmd)
+						<-guard
+
+						dirtyVolumes = remove(dirtyVolumes, volume.volumeName)
```
abhishekdwivedi3060, line 616: Same can be done here also. Maybe create 2 func like `runDD` and `runBLkDiscard`

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,842 @@
...
+
+	podSpecHash := string(data)
+	metadata.Image = podImage
+	metadata.InitializedVolumes = volumes
+	metadata.DirtyVolumes = dirtyVolumes
+	metadata.AerospikeConfigHash = confHash
+	metadata.NetworkPolicyHash = networkPolicyHash
+	metadata.PodSpecHash = podSpecHash
+
+	for podAddrName, confAddrName := range AddressTypeName {
```
abhishekdwivedi3060, line 785: Shouldn't this be ?
```suggestion
	for confAddrName, podAddrName := range AddressTypeName {
```

abhishekdwivedi3060 (changes requested):

init/cmd/root.go
```diff
@@ -0,0 +1,47 @@
...
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
+	rootCmd.PersistentFlags().StringVarP(&podName, "pod-name", "p", "", "Current pod name")
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "", "Namespace under which cluster is running")
+	rootCmd.PersistentFlags().StringVarP(&clusterName, "cluster-name", "c", "", "Aerospike cluster name")
```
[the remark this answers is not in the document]
tanmayja: clusterName we can move to its sub command but other two flags are common in more than 1 sub-commands, so keeping them here only.

controllers/scripts/common-env.sh
```diff
@@ -15,18 +15,30 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 # ------------------------------------------------------------------------------
-set -e
-set -x
```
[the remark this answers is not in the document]
tanmayja: common-env.sh is being sourced, so if parent bash has set certain fields, there is no need to add them in source.

init/pkg/update_pod_status_util.go
```diff
@@ -32,8 +30,6 @@ const (
 	BaseWipeVersion      = 6
 	BlockVolumeMode      = "Block"
 	FilesystemVolumeMode = "Filesystem"
```
abhishekdwivedi3060, line 32: User corev1.PersistentVolumeBlock and corev1.PersistentVolumeFilesystem instead

init/pkg/update_pod_status_util.go
```diff
@@ -137,50 +131,50 @@ func getPodImage(k8sClient client.Client, ctx goctx.Context, podNamespacedName t
...
 
 //nolint:unparam // generic function
-func getEnv(key, fallback string) string {
+func getEnv(key string) string {
 	if value, ok := os.LookupEnv(key); ok {
 		return value
 	}
 
-	return fallback
+	return ""
```
abhishekdwivedi3060, line 139: We can remove this func all together and use os.Getenv()

init/pkg/update_pod_status_util.go
```diff
@@ -190,19 +184,16 @@ func getInitializedVolumes(podName string, aeroCluster *asdbv1beta1.AerospikeClu
...
-		return volumes
-	}
+func getAttachedVolumes(rack *asdbv1beta1.Rack) []asdbv1beta1.VolumeSpec {
+	logrus.Info("Looking for volumes in rack.effectiveStorage.volumes ", "rack-id=", rack.ID)
 
-	log.Println("INFO: Looking for volumes in spec.storage.volumes")
+	volumes := rack.Storage.Volumes
 
-	return aeroCluster.Spec.Storage.Volumes
+	return volumes
```
abhishekdwivedi3060, line 196: nit : `return rack.Storage.Volumes`

init/pkg/update_pod_status_util.go
```diff
@@ -228,13 +219,49 @@ func newVolume(podName string, vol *asdbv1beta1.VolumeSpec) Volume {
...
 	if vol.Aerospike != nil {
-		volume.attachmentType = "aerospike"
 		volume.aerospikeVolumePath = vol.Aerospike.Path
 	}
 
-	return volume
+	return &volume
 }
 
+func runDD(cmd []string, wg sync.WaitGroup, guard chan struct{}) {
```
abhishekdwivedi3060, line 228: Should pass WaitGroup pointer for reusing the main WaitGroup

init/pkg/update_pod_status_util.go
```diff
@@ -228,13 +219,49 @@ func newVolume(podName string, vol *asdbv1beta1.VolumeSpec) Volume {
...
+		}
+	}
+
+	stderr.Close()
+	os.Remove(stderr.Name())
+	logrus.Info("Execution completed ", "cmd=", cmd)
+	<-guard
+}
+
+func runBlkdiscard(cmd []string, wg sync.WaitGroup, guard chan struct{}) {
```
abhishekdwivedi3060, line 255: same here

init/pkg/update_pod_status_util.go
```diff
@@ -743,18 +644,16 @@ func ManageVolumesAndUpdateStatus(podName, namespace, clusterName string, restar
 
 	metadata := getNodeMetadata()
 
-	log.Println("INFO: Updating pod status", "podname", podName)
+	logrus.Info("Updating pod status ", "podname=", podName)
 
-	if err := updateStatus(k8sClient, goctx.TODO(), aeroCluster, podName, podImage,
-		metadata, volumes, dirtyVolumes); err != nil {
+	if err := updateStatus(goctx.TODO(), k8sClient, aeroCluster, podName, podImage, metadata, volumes, dirtyVolumes); err != nil {
```
abhishekdwivedi3060, line 649: There are around 8 params passed to this func and some of them are directly assigned to metadata struct inside that func. Can't we assign such param like volumes, dirtyVolumes, podImage here only and then pass metadata and remaining params to avoid params clutter

abhishekdwivedi3060 (changes requested):

init/pkg/update_pod_status_util.go
tanmayja: there are certain flags which needs to be available in env else system should panic, Hence adding panic if those flags not found.

sud82:
Subset of #196

PR #196, opened 2023-03-09
[KO-197] converting init scripts to go binary

tanmayja:


init/pkg/refresh-cmap-restart-asd.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	asdProc, err := ps.FindProcess(asdPid)
+	if err != nil {
+		return err
+	}
+
+	if asdProc != nil {
+		return fmt.Errorf("aborting warm start - Aerospike server did not terminate")
+	}
```
abhishekdwivedi3060, line 76: Instead of finding process again, we can set a var `isTerminated` in above for loop in case of else condition and use that var over here for checking the final termination

abhishekdwivedi3060 (changes requested):

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,700 @@
...
+	metadata.PodSpecHash = podSpecHash
+
+	var AddressTypeName = [4]string{
+		Access,
+		AlternateAccess,
+		TLSAccess,
+		TLSAlternateAccess,
+	}
+
+	for _, podAddrName := range AddressTypeName {
```
abhishekdwivedi3060, line 622: Should be 
```suggestion
	for _, confAddrName := range AddressTypeName {
```

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,700 @@
...
+
+func (initp *InitParams) getEndpoints(addressType string) []string {
+	var (
+		endpoint   []string
+		globalAddr string
+		globalPort int32
+	)
+
+	switch addressType {
+	case "access":
```
abhishekdwivedi3060, line 667: Use already defined const in cases

abhishekdwivedi3060 (changes requested):

init/pkg/create-aerospike-conf.go
```diff
@@ -0,0 +1,184 @@
...
+			confString = fmt.Sprintf("%sheartbeat {\n        tls-mesh-seed-address-port %s %d%s",
+				strdataSplit[0], peer, initp.initTemplateInput.HeartBeatTLSPort, strdataSplit[1])
+		}
+	}
+
+	if initp.initTemplateInput.HostNetwork {
+		if initp.initTemplateInput.HeartBeatPort != 0 {
+			strdataSplit := strings.Split(confString, "heartbeat {")
+			confString = fmt.Sprintf("%sheartbeat {\n        address %s%s",
+				strdataSplit[0], os.Getenv("MY_POD_IP"), strdataSplit[1])
```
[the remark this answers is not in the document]
tanmayja: 1. Both approaches are fine.

init/pkg/restart-asd.go
```diff
@@ -0,0 +1,27 @@
...
+
+	strdata := string(data)
+	if !(strings.Contains(strdata, "tini") && strings.Contains(strdata, "-r")) {
+		return fmt.Errorf("warm restart not supported - aborting")
+	}
+
+	// Create new Aerospike configuration
+	if err := initp.copyTemplates(filepath.Join(configVolume, "configmap"), configVolume); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
tanmayja: moved to QuickRestart

init/pkg/initialize.go
```diff
@@ -0,0 +1,57 @@
...
+	if err := initp.makeWorkDir(); err != nil {
+		return err
+	}
+
+	// Copy required files to config volume for initialization.
+	if err := initp.copyTemplates("/configs", configVolume); err != nil {
+		return err
+	}
+
+	filesToCopy := [2]string{"/workdir/bin/akoinit", "/configs/features.conf"}
```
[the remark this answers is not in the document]
tanmayja: @sud82 please confirm if it is ok to copy the features.conf file in warm restart.

init/pkg/create-aerospike-conf.go
```diff
@@ -58,117 +62,107 @@ func (initp *InitParams) createAerospikeConf() error {
...
-	if addressType == TLSAccess || addressType == TLSAlternateAccess {
-		if addressType == TLSAccess {
-			networkType = initp.initTemplateInput.NetworkPolicy.TLSAccessType
-		} else {
-			networkType = initp.initTemplateInput.NetworkPolicy.TLSAlternateAccessType
-		}
-
-		podPort = initp.initTemplateInput.PodTLSPort
-		mappedPort = initp.mappedTLSPort
+	if addressType == tlsAccess || addressType == tlsAlternateAccess {
```
abhishekdwivedi3060, line 127: See if we can use this `if os.Getenv("MY_POD_TLS_ENABLED") == "true"` instead just like at other places

abhishekdwivedi3060 (changes requested):

init/pkg/create-aerospike-conf.go
tanmayja: why to do a os call, if we can simply use this if.

init/pkg/utils.go
```diff
@@ -9,23 +9,41 @@ import (
...
+	FabricPort              int32
+	FabricTLSPort           int32
+	PodPort                 int32
+	PodTLSPort              int32
+	HeartBeatPort           int32
+	HeartBeatTLSPort        int32
+	mappedPort              int32
+	mappedTLSPort           int32
+	MultiPodPerHost         bool
+	HostNetwork             bool
```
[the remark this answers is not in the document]
tanmayja: converting all to private

init/pkg/create-aerospike-conf.go
abhishekdwivedi3060: Both ways are fine, just added comment to keep this TLS differentiator flow in sync. At one place we use that MY_POD_TLS_ENABLED var

init/pkg/utils.go
```diff
@@ -0,0 +1,271 @@
...
+func (initp *InitParams) makeWorkDir() error {
+	if initp.workDir != "" {
+		defaultWorkDir := filepath.Join("workdir", "filesystem-volumes", initp.workDir)
+
+		requiredDirs := [3]string{"smd", "usr/udf/lua", "xdr"}
+		for _, d := range requiredDirs {
+			toCreate := filepath.Join(defaultWorkDir, d)
+			logrus.Info("Creating directory ", toCreate)
+
+			if err := os.MkdirAll(toCreate, 0644); err != nil { //nolint:gocritic // file permission
```
sud82, line 168: I think we should add execute permission too for directories.
0755.
Check other places as well.

init/pkg/common-env.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	nodeID, err := GetNodeIDFromPodName(podName)
+	if err != nil {
+		return nil, err
+	}
+
+	workDir := asdbv1beta1.GetWorkDirectory(rack.AerospikeConfig)
```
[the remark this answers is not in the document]
tanmayja: GetWorkDirectory is used in [aerospikecluster_validating_webhook.go] and [controllers/configmap.go].

init/pkg/initialize.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, file := range filesToCopy {
+		if _, err := os.Stat(file); err == nil {
```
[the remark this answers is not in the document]
tanmayja: os.Stat command is used here to check if file exist or not before copying, otherwise cp command will fail. And in scripts also we were not failing the script if features.conf file was not there.

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,703 @@
...
+			}
+
+		default:
+			return volumeNames, fmt.Errorf("invalid volume-mode %s", volume.volumeMode)
+		}
+
+		volumeNames = append(volumeNames, volume.volumeName)
+	}
+
+	close(guard)
```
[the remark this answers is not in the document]
tanmayja: there is no harm in closing the channel though

init/pkg/update_pod_status_util.go
```diff
@@ -0,0 +1,703 @@
...
+	stderr.Close()
+	os.Remove(stderr.Name())
+	logrus.Info("Execution completed ", "cmd=", cmd)
+	<-guard
+}
+
+func runBlkdiscard(cmd []string, wg *sync.WaitGroup, guard chan struct{}) {
+	defer wg.Done()
+
+	if err := execute(cmd, nil); err != nil {
```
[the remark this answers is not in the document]
tanmayja: no, we used stdErr to check for "grep -q "No space left on device"" this particular error which only comes in dd command after clearing last block.

init/pkg/initialize.go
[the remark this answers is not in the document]
sud82: Add a comment that we shouldn't fail if features.conf is not present. akoinit will always be there because it is getting checked in entrypoint.sh

init/pkg/common-env.go
```diff
@@ -32,6 +41,15 @@ func PopulateInitParams() (*InitParams, error) {
 		scheme    = runtime.NewScheme()
 	)
 
+	logger := ctrl.Log.WithName("setup")
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
```
abhishekdwivedi3060, line 49: Seems redundant as init binary is not user facing to set this flag

init/pkg/update_pod_status_util.go
```diff
@@ -46,52 +47,49 @@ func (v *Volume) getMountPoint() string {
...
+	if err := command.Run(); err != nil {
+		return err
 	}
 
 	return nil
 }
 
-func getPodImage(ctx goctx.Context, k8sClient client.Client, podNamespacedName types.NamespacedName) (string, error) {
-	logrus.Info("Get pod image ", "pod-name=", podNamespacedName)
+func (initp *InitParams) getPodImage(podNamespacedName types.NamespacedName) (string, error) {
```
abhishekdwivedi3060, line 88: Why did we remove ctx from func param?
Ideally context should be defined once (parent context) in the top level func and the same context should be passed to every child func called instead of creating of creating context again.

See if we can incorporate this, if this is too much change then can defer this to next python-to-go PR

abhishekdwivedi3060 (approved):

init/pkg/update_pod_status_util.go
tanmayja: made changes in this itself

PR #199, opened 2023-03-13
[KO-200] Allow user to configure IP for access-address and alternate access ad…

sud82:
…dress using node labels

### DIFF
--- a/controllers/scripts/common-env.sh
+++ b/controllers/scripts/common-env.sh
@@ -46,6 +46,9 @@ print(getport(data, podname))")"
 export PODIP="$MY_POD_IP"
 INTERNALIP="$MY_HOST_IP"
 
+export CONFIGURED_ACCESSIP_LABEL="aerospike.com/configured-access-address"
+export CONFIGURED_ALTERNATE_ACCESSIP_LABEL="aerospike.com/configured-alternate-access-address"
+
 # Get External IP
 DATA="$(curl --cacert $CA_CERT -H "Authorization: Bearer $TOKEN" "$KUBE_API_SERVER/api/v1/nodes")"
 
@@ -56,6 +59,8 @@ host = '${MY_HOST_IP}';
 def gethost(data, host):
     internalIP = host
     externalIP = host
+    configuredAccessIP = ''
+    configuredAlternateAccessIP = ''
 
     # Iterate over all nodes and find this pod's node IPs.
     for item in data['items']:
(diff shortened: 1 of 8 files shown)


controllers/scripts/common-env.sh
```diff
@@ -79,14 +84,25 @@ def gethost(data, host):
...
 
-    return internalIP + ' ' + externalIP
+    return internalIP + ' ' + externalIP + ' ' + configuredAccessIP + ' ' + configuredAlternateAccessIP
 
 print(gethost(data, host))")"
 
 export INTERNALIP=$(echo $HOSTIPS | awk '{print $1}')
 export EXTERNALIP=$(echo $HOSTIPS | awk '{print $2}')
+export CONFIGURED_ACCESSIP=$(echo $HOSTIPS | awk '{print $3}')
+export CONFIGURED_ALTERNATE_ACCESSIP=$(echo $HOSTIPS | awk '{print $4}')
```
abhishekdwivedi3060, line 105: I guess this might cause issue when only `CONFIGURED_ALTERNATE_ACCESSIP_LABEL` label is present.
In that case,  `configuredAccessIP ` value be empty and `awk '{print $3}'` will actually give `configuredAlternateAccessIP` value and `awk '{print $4}'` will return empty value

abhishekdwivedi3060 (changes requested):

tanmayja (commented):
Like internal and external IP should we also mention configured ip to pod status?

https://github.com/aerospike/aerospike-kubernetes-operator/controllers/scripts/create_pod_status_patch.py#L200

controllers/scripts/common-env.sh
sud82: Fixed

test/network_policy_test.go
```diff
@@ -236,7 +332,7 @@ func validateNetworkPolicy(
 func validatePodEndpoint(
 	ctx goctx.Context, pod *corev1.Pod,
 	aeroCluster *asdbv1beta1.AerospikeCluster,
-	networkType asdbv1beta1.AerospikeNetworkType, isTLS bool, actual []string,
+	networkType asdbv1beta1.AerospikeNetworkType, isTLS bool, actual []string, configuredIP string,
 ) error {
 	podIP, hostInternalIP, hostExternalIP, _ := getIPs(ctx, pod)
```
tanmayja, line 337: Any reason, why are we not fetching configuredIP from node like internal and externalIP, as we have already set in labels(if done successfully)

abhishekdwivedi3060 (changes requested):

test/network_policy_test.go
abhishekdwivedi3060: Makes sense to return all IPs from `getIPs` func

test/network_policy_test.go
```diff
@@ -477,3 +581,51 @@ func getAerospikeClusterSpecWithNetworkPolicy(
 		},
 	}
 }
+
+func setNodeLabel(ctx goctx.Context, key, val string) error {
```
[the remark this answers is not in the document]
sud82: There are just two labels so I did feel the need to create a list/map and then `for` loop for them.

test/network_policy_test.go
```diff
@@ -477,3 +581,51 @@ func getAerospikeClusterSpecWithNetworkPolicy(
...
+
+		if err := k8sClient.Update(ctx, node); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func deleteNodeLabel(ctx goctx.Context, key string) error {
```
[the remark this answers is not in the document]
sud82: same as above ans

test/network_policy_test.go
```diff
@@ -477,3 +581,51 @@ func getAerospikeClusterSpecWithNetworkPolicy(
...
+
+		if err := k8sClient.Update(ctx, node); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func removeCustomNodeLabel(ctx goctx.Context) error {
```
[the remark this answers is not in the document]
sud82: same as above ans

test/network_policy_test.go
sud82: This func is common for `access` and `alternate-access`. So, if we fetch IP from getIPs() then we have to know which IP to match. At present we are already sending only that IP that needs to be matched.

tanmayja: I think as of now it is fine, but we can write validatePodEndpoint in a better way in future, where we pass less arguments.

tanmayja (approved):

abhishekdwivedi3060 (approved):

PR #200, opened 2023-03-14
KO-130: Added support for adding custom interfaces

abhishekdwivedi3060:
- Added support for adding custom interfaces for Access, AlternateAccess and Fabric


controllers/scripts/common-env.sh
```diff
@@ -103,8 +103,60 @@ export MAPPED_PORT="$POD_PORT"
...
+def getNetworkIP(data, networks):
+    annotations = data['metadata']['annotations']
+    key = 'k8s.v1.cni.cncf.io/network-status'
+    if key in annotations:
+      netIPs = []
+      for net in json.loads(annotations[key]):
+        if net['name'] in networks:
+          netIPs.extend(net['ips'])
+      return ' '.join(netIPs)
+print(getNetworkIP(data, networks))")"
```
abhishekdwivedi3060, line 125: Intentionally kept support for multiple networkNames even though only 1 networkName is allowed in CRD Schema

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -1833,3 +1837,93 @@ func validateDNS(dnsPolicy v1.DNSPolicy, dnsConfig *v1.PodDNSConfig) error {
...
+	networkPolicy := &c.Spec.AerospikeNetworkPolicy
+
+	annotations := c.Spec.PodSpec.AerospikeObjectMeta.Annotations
+	networks := annotations["k8s.v1.cni.cncf.io/networks"]
+	networkList := strings.Split(networks, ",")
+	networkSet := sets.NewString()
+
+	for ind := range networkList {
+		netName := strings.TrimSpace(networkList[ind])
+		namespacedName := strings.Split(netName, "/")
```
tanmayja, line 1851: This logic is similar to part of setNamespaceDefault function, can we re-use it somehow?

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -1833,3 +1837,93 @@ func validateDNS(dnsPolicy v1.DNSPolicy, dnsConfig *v1.PodDNSConfig) error {
...
+
+func (c *AerospikeCluster) validateNetworkPolicy(namespace string) error {
+	networkPolicy := &c.Spec.AerospikeNetworkPolicy
+
+	annotations := c.Spec.PodSpec.AerospikeObjectMeta.Annotations
+	networks := annotations["k8s.v1.cni.cncf.io/networks"]
+	networkList := strings.Split(networks, ",")
+	networkSet := sets.NewString()
+
+	for ind := range networkList {
```
tanmayja, line 1849: idx

test/cluster_test.go
```diff
@@ -1128,6 +1216,115 @@ func negativeUpdateClusterValidationTest(
 				},
 			)
 
+			FContext(
```
tanmayja, line 1219: why FContext?

api/v1beta1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: Modified func and reused, however it'll result in 1 for loop to insert into set

test/cluster_test.go
abhishekdwivedi3060: For testing, removed

test/network_policy_test.go
```diff
@@ -62,12 +64,271 @@ var _ = Describe(
...
+	)
+}
+
 func doTestNetworkPolicy(
 	multiPodPerHost bool, enableTLS bool, ctx goctx.Context,
 ) {
+	var aeroCluster *asdbv1beta1.AerospikeCluster
+
+	AfterEach(func() {
+		err := deleteCluster(k8sClient, ctx, aeroCluster)
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Actually, we don't need to get the cluster to delete it, just name and namespace is sufficient.
We might have to fix this behaviour at other places of AfterEach to  avoid unnecessary `getCluster` call.
However, we can add a safe check of notFound error to avoid AfterEach failure

sud82 (changes requested):

controllers/scripts/common-env.sh
```diff
@@ -103,8 +103,60 @@ export MAPPED_PORT="$POD_PORT"
...
+networks = networks.strip().split(' ')
+def getNetworkIP(data, networks):
+    annotations = data['metadata']['annotations']
+    key = 'k8s.v1.cni.cncf.io/network-status'
+    if key in annotations:
+      netIPs = []
+      for net in json.loads(annotations[key]):
+        if net['name'] in networks:
+          netIPs.extend(net['ips'])
+      return ' '.join(netIPs)
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Referred network is already checked in webhook to fail fast.
If CNI is not working as expected and netIPs is empty then empty value will be updated in corresponding interface in aerospike.conf and server will crash

api/v1beta1/aerospikecluster_mutating_webhook.go
```diff
@@ -284,6 +287,51 @@ func (c *AerospikeCluster) setDefaultAerospikeConfigs(
...
+		n.TLSAccessType = AerospikeNetworkTypeHostInternal
+	}
+
+	if n.TLSAlternateAccessType == AerospikeNetworkTypeUnspecified {
+		n.TLSAlternateAccessType = AerospikeNetworkTypeHostExternal
+	}
+}
+
+func (n *AerospikeNetworkPolicy) setNetworkNamespace(namespace string) {
+	if n.CustomAccessNetworkNames != nil {
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Removed all if conditions, for loop inside `setNamespaceDefault` func will take care of nil slice

api/v1beta1/aerospikecluster_validating_webhook.go
```diff
@@ -1833,3 +1865,86 @@ func validateDNS(dnsPolicy v1.DNSPolicy, dnsConfig *v1.PodDNSConfig) error {
...
+	networks := annotations["k8s.v1.cni.cncf.io/networks"]
+	networkList := strings.Split(networks, ",")
+	networkSet := sets.NewString()
+
+	setNamespaceDefault(networkList, namespace)
+
+	networkSet.Insert(networkList...)
+
+	if networkPolicy.AccessType == AerospikeNetworkTypeOthers {
+		if networkPolicy.CustomAccessNetworkNames == nil {
```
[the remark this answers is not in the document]
abhishekdwivedi3060: 2. Can't be empty if given in CR because of CRD schema validation of minItems=1. It'll error out before reaching validation

test/network_policy_test.go
```diff
@@ -62,12 +64,271 @@ var _ = Describe(
 				)
 			},
 		)
+
```
sud82, line 67: Few more tests:

1. HostNetworking related validation
2. Network Namespace different from CR
3. Invalid network name given in both (pod annotations and network policy) should fail
4. Max, Min networkName len 1
5. Fabric networkNames update should fail
6. Recovery test: pass wrong networkNames, pod should stuck, then update CR with right networkNames, pod should succeed.

api/v1beta1/aerospikecluster_types.go
```diff
@@ -738,44 +737,80 @@ const (
...
 )
 
 // AerospikeNetworkPolicy specifies how clients and tools access the Aerospike cluster.
 type AerospikeNetworkPolicy struct {
 	// AccessType is the type of network address to use for Aerospike access address.
 	// Defaults to hostInternal.
+	// +kubebuilder:validation:Enum=pod;hostInternal;hostExternal;others
 	AccessType AerospikeNetworkType `json:"access,omitempty"`
 
+	// CustomAccessNetworkNames is the list of networks used to attach to the pod for Aerospike access address.
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Not adding container specific description because those interfaces will be attached to pod and visible to all containers inside that

test/network_policy_test.go
abhishekdwivedi3060: 4.  I don't think these test-cases are required as these limits are handled at CRD schema level
5. Not added because can't test with fabric interfaces at the moment as it fails in cluster create flow. Reason is it tries to bind the random IPs provided in the annotations

controllers/scripts/common-env.sh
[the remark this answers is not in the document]
sud82: let's crash in init itself. That will be easier to debug the problem. Also if the user has given a wrong network name and a right network name in the networkNames and podAnnotations. Then netIPs will have the ips from the right network only and will be added to the config file. The server will not crash in this case. We should fail in that case otherwise user will not know if the network name is wrong.

test/network_policy_test.go
sud82: 4. I think tests should not be dependent on how the validations are happening internally. It should mainly care about the expectation made to the user. If validations are removed or changed in the future from CRD then tests should be able to catch that.

controllers/scripts/common-env.sh
```diff
@@ -103,8 +103,70 @@ export MAPPED_PORT="$POD_PORT"
...
+            raise KeyError(\"ips key missing for network {} in pod annotations key k8s.v1.cni.cncf.io/network-status\".format(net['name']))
+
+      if len(netIPs) == 0:
+        raise ValueError(\"networks {} not found in pod annotations key k8s.v1.cni.cncf.io/network-status\".format(networks))
+      return ' '.join(netIPs)
+print(getNetworkIP(data, networks))")"
+}
+
+{{- if eq .NetworkPolicy.AccessType "customInterface"}}
+parseCustomNetworkIP "{{ .NetworkPolicy.CustomAccessNetworkNames}}"
```
tanmayja, line 139: any reason for a space between {{ and .Network?

test/network_policy_test.go
```diff
@@ -62,12 +73,497 @@ var _ = Describe(
...
+					aeroCluster.Spec.AerospikeNetworkPolicy.AccessType = asdbv1beta1.AerospikeNetworkTypeCustomInterface
+					aeroCluster.Spec.AerospikeNetworkPolicy.CustomAccessNetworkNames = []string{}
+					err := deployCluster(k8sClient, ctx, aeroCluster)
+					Expect(err).Should(HaveOccurred())
+				},
+			)
+
+			It(
+				"TwoCustomAccessNetworkNames: should fail when access is set to 'customInterface' and "+
+					"customAccessNetworkNames is added with 2 entries",
```
sud82, line 259: To bring more clarity:
MaxAllowedNetworkNames. It should fail when more than the max allowed network names(only 1) are given in CR.

sud82: Add this in updateTests also

test/network_policy_test.go
```diff
@@ -115,11 +608,132 @@ func doTestNetworkPolicy(
...
 			Expect(err).ToNot(HaveOccurred())
-
-			err = deleteCluster(k8sClient, ctx, aeroCluster)
-			Expect(err).ToNot(HaveOccurred())
 		},
 	)
+
+	Context("customInterface", func() {
+		clusterNamespacedName := getClusterNamespacedName(
+			"np-custom-interface", multiClusterNs1,
```
sud82, line 616: Let's use a different namespace for cluster and networkName.

sud82 (changes requested):

test/network_policy_test.go
abhishekdwivedi3060: Seems not much useful as per our discussion

test/network_policy_test.go
abhishekdwivedi3060: Test-case with different namespace seems not useful as we are mocking the CNI behaviour. 
Added a comment around the limitations of current test-suite for customInterfaces

controllers/scripts/common-env.sh
abhishekdwivedi3060: Remove space to keep it consistent

sud82 (approved):

tanmayja (approved):

PR #205, opened 2023-04-03
KO-186: Added test-cases for samples files

abhishekdwivedi3060:
- Added test-cases for samples files
- Minor refactor in test-suite


test/cleanup-test-namespace.sh
```diff
@@ -12,12 +12,13 @@ echo "Removing Aerospike clusters"
 kubectl -n test delete aerospikecluster --all
 kubectl -n test1 delete aerospikecluster --all
 kubectl -n test2 delete aerospikecluster --all
-kubectl -n test3 delete aerospikecluster --all
+kubectl -n aerospike delete aerospikecluster --all
```
tanmayja, line 15: is it true from now on that we can't have any aerospikecluster running in aerosike namespace if we need to run test suite?

abhishekdwivedi3060: If you run test-suite using test script then yes it'll cleanup everything before running test-suite. However, if you run tests using `make test` then you can have existing aerospikecluster running

sud82 (approved):

tanmayja (approved):

PR #207, opened 2023-04-18
KO-219: Skipped nodePort service creation for pod network type

abhishekdwivedi3060:
- Skipped nodePort service creation for pod network type
- Skipped ServiceList and NodeList API call in case of pod network type
- Added new fields `podTlsPort` and `serviceTlsPort` in pod status for TLS mapping
- Fixed port related issue in pod status


controllers/pod.go
```diff
@@ -251,6 +252,18 @@ func (r *SingleClusterReconciler) restartPods(
...
 	for idx := range podsToRestart {
 		pod := podsToRestart[idx]
+
+		// Create services for all pods if network policy is changed and rely on nodePort service.
+		// Check podServiceNeeded condition for both status and spec network policy
+		if !podServiceNeeded(r.aeroCluster.Status.PodSpec.MultiPodPerHost, &r.aeroCluster.Status.AerospikeNetworkPolicy) &&
+			podServiceNeeded(r.aeroCluster.Spec.PodSpec.MultiPodPerHost, &r.aeroCluster.Spec.AerospikeNetworkPolicy) {
+			if err := r.createPodService(
+				pod.Name, r.aeroCluster.Namespace,
+			); err != nil && errors.IsAlreadyExists(err) {
```
tanmayja, line 262: why are we returning error if service is already exist and bypassing other errors?

abhishekdwivedi3060: Fixed and moved code block to diff location

controllers/rack.go
```diff
@@ -667,6 +691,20 @@ func (r *SingleClusterReconciler) upgradeRack(statefulSet *appsv1.StatefulSet, r
 			"rollingUpdateBatchSize", r.aeroCluster.Spec.RackConfig.RollingUpdateBatchSize,
 		)
 
+		// Create services for all pods if network policy is changed and rely on nodePort service.
```
tanmayja, line 694: Now we have a common function upgradeOrRollingRestartRack, can you check if we can re-structure some code and carve out some common functionality from upgradeRack and rollingRestartRack function.

abhishekdwivedi3060: `upgradeOrRollingRestartRack` func doesn't affect the functionality of upgradeRack and rollingRestartRack, so restructuring is not possible at the moment. However, I've moved the service creation code out in a func to be used for both

tanmayja (approved):

controllers/rack.go
```diff
@@ -232,6 +232,20 @@ func (r *SingleClusterReconciler) createRack(rackState *RackState) (
 		return nil, reconcileError(err)
 	}
 
+	// Create pod service if node network is used in network policy
```
sud82, line 235: Service has to be created before creating the pod. Initcontainer uses the service to fetch the mapped port for the pod. Are you sure this flow is working?

config/crd/bases/asdb.aerospike.com_aerospikeclusters.yaml
```diff
@@ -13185,19 +13185,26 @@ spec:
                       description: PodSpecHash is ripemd160 hash of PodSpec used by
                         this pod
                       type: string
+                    podTlsPort:
```
sud82, line 13188: Need to update the description of multiPodPerHost.

sud82 (changes requested):

controllers/rack.go
```diff
@@ -499,6 +513,15 @@ func (r *SingleClusterReconciler) reconcileRack(
 		return reconcileError(err)
 	}
 
+	// Delete all pod service if network policy is updated to some non node network type in network policy
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Added a safe check to always try to delete pod services if networkType is pod

controllers/rack.go
abhishekdwivedi3060: Services are always created before pod in the scaleup function not in createSTS func. Even in the current master service creation logic in the createSTS func was of no use as rack.Size passed was 0, so no for loop execution.
Removed that code block completely.

sud82: Rename the func name to reflect the changes

sud82 (approved):

PR #210, opened 2023-05-02
Some enhancements in init logs

tanmayja:
Improving init code flow and logs.


init/pkg/update_pod_status_util.go
```diff
@@ -299,36 +293,31 @@ func initVolumes(logger logr.Logger, podName string, aeroCluster *asdbv1beta1.Ae
...
 }
 
-func getNamespaceVolumePaths(logger logr.Logger, podName string, aeroCluster *asdbv1beta1.AerospikeCluster) (
-	devicePaths, filePaths []string, err error) {
-	rack, err := getRack(logger, podName, aeroCluster)
-	if err != nil {
-		return devicePaths, filePaths, fmt.Errorf("failed to get rack of pod %s %v", podName, err)
-	}
-
+func (initp *InitParams) getNamespaceVolumePaths() (
```
abhishekdwivedi3060, line 301: This func is also called at 2 places, wont't this print storage-engine logs multiple times as well?

abhishekdwivedi3060 (changes requested):

init/pkg/update_pod_status_util.go
tanmayja: that's fine, one place in under wipeVolumes, which is being called if wipe is needed.

sud82 (approved):

PR #225, opened 2023-06-15
Adding newly formatted initialisedVolumes

tanmayja:


tanmayja:
Need to change init tag before merging.

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+
+	setupLog.Info("Formatting Initialised Volumes name")
+
+	if err := addNewlyFormattedInitialisedVolumeNames(ctx, client, setupLog); err != nil {
+		setupLog.Error(err, "problem patching Initialised volumes")
+		os.Exit(1)
+	}
+}
+
+func addNewlyFormattedInitialisedVolumeNames(ctx context.Context, client crClient.Client, setupLog logr.Logger) error {
```
abhishekdwivedi3060, line 254: nit: 
```suggestion
func migrateInitialisedVolumeNames((ctx context.Context, client crClient.Client, setupLog logr.Logger) error {
```

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+	aeroClusterList := &asdbv1.AerospikeClusterList{}
+	if err := client.List(ctx, aeroClusterList); err != nil {
```
abhishekdwivedi3060, line 256: Shouldn't we process `AerospikeCluster` as per the watchNamespaces value?

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+		if errors.IsNotFound(err) {
+			// Request objects not found.
+			return nil
+		}
+		// Error reading the object.
+		return err
+	}
+
+	for idx := range aeroClusterList.Items {
+		aeroCluster := aeroClusterList.Items[idx]
```
abhishekdwivedi3060, line 266:
```suggestion
		aeroCluster := &aeroClusterList.Items[idx]
```

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+
+		podList, err := getClusterPodList(ctx, client, &aeroCluster)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				// Request objects not found.
+				return nil
```
abhishekdwivedi3060, line 272:
```suggestion
				continue
```

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+			}
+			// Error reading the object.
+			return err
+		}
+
+		var patches []jsonpatch.PatchOperation
+
+		for podIdx := range podList.Items {
+			pod := podList.Items[podIdx]
```
abhishekdwivedi3060, line 281:
```suggestion
			pod := &podList.Items[podIdx]
```

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+
+		// Since the pod status is updated from pod init container,
+		// set the field owner to "pod" for pod status updates.
+		setupLog.Info("Patching status with updated initialised volumes")
+
+		if err = client.Status().Patch(
+			context.TODO(), &aeroCluster, constantPatch, crClient.FieldOwner("pod"),
+		); err != nil {
+			return fmt.Errorf("error updating status: %v", err)
+		}
```
abhishekdwivedi3060, line 337: - Update the comment
- Use context from above
- Shouldn't the field owner be `controller` in this case?

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+				if len(initVolInfo) < 2 {
+					oldFormatInitVolNames = append(oldFormatInitVolNames, initializedVolumes[idx])
+				} else {
+					initializedVolumesMap[initVolInfo[0]] = initVolInfo[1]
+				}
+			}
+
+			for idx := range oldFormatInitVolNames {
+				if _, ok := initializedVolumesMap[oldFormatInitVolNames[idx]]; !ok {
+					pvcUID, pvcErr := getPVCUid(context.TODO(), client, &pod, initializedVolumes[idx])
```
abhishekdwivedi3060, line 297:
```suggestion
					pvcUID, pvcErr := getPVCUid(ctx, client, &pod, oldFormatInitVolNames[idx])
```

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
...
+					if pvcErr != nil {
+						return pvcErr
+					}
+
+					// Appending volume name as <vol_name>@<pvcUID> in initializedVolumes list
+					initializedVolumes = append(initializedVolumes, fmt.Sprintf("%s@%s", initializedVolumes[idx], pvcUID))
+				}
+			}
```
abhishekdwivedi3060, line 305: These 2 can be combined in 1 for loop

abhishekdwivedi3060 (changes requested):

main.go
tanmayja: [Here](https://github.com/aerospike/aerospike-kubernetes-operator/blob/f3bd4303156b43141f4861aab35d4d1e7b9a10f4/controllers/pod.go#L987) also we are updating patch from controller only with owner as pod and comment is regarding that only.

main.go
```diff
@@ -190,6 +203,163 @@ func main() {
 	eventBroadcaster.Shutdown()
 }
 
+func migrationRoutine(client crClient.Client, leaderElectionID string) {
+	for {
+		// Checking if controller is ready.
+		resp, err := http.Get("http://localhost:8081/readyz")
```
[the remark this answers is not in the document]
tanmayja: could not find the endpoint name populated in options, reading port from options.

main.go
tanmayja: Combining these loops is not possible, but I changed the initializedVolumesMap to sets of string.

main.go
```diff
@@ -180,6 +191,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	go migrationRoutine(client, options.LeaderElectionID)
```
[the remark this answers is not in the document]
abhishekdwivedi3060: Sorry for the typo from my end. It should be `migrateAerospikeClusters`

main.go
```diff
@@ -158,38 +170,220 @@ func main() {
...
+							return pvcErr
+						}
+
+						// Appending volume name as <vol_name>@<pvcUID> in initializedVolumes list
+						initializedVolumes = append(initializedVolumes, fmt.Sprintf("%s@%s", oldFormatInitVolNames[idx], pvcUID))
+					}
+				}
+
+				if len(initializedVolumes) > len(aeroCluster.Status.Pods[pod.Name].InitializedVolumes) {
+					setupLog.Info("Got updates initialised volumes list", "initvol", initializedVolumes, "pod-name", pod.Name)
```
abhishekdwivedi3060, line 331:
```suggestion
					setupLog.Info("Got updated initialised volumes list", "initVolumes", initializedVolumes, "pod-name", pod.Name)
```

abhishekdwivedi3060 (approved):

test/rack_utils.go
```diff
@@ -472,6 +472,13 @@ func getClusterNamespacedName(name, namespace string) types.NamespacedName {
 	}
 }
 
+func getPVCNamespacedName(name, namespace string) types.NamespacedName {
```
abhishekdwivedi3060, line 475: Can create a generic func called `getNamespacedName` which will return `types.NamespacedName` for any object

test/storage_init_test.go
```diff
@@ -338,9 +337,139 @@ var _ = Describe(
...
+
+						aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+						Expect(err).ToNot(HaveOccurred())
+
+						newInitVol := aeroCluster.Status.Pods[unscheduledPod].InitializedVolumes
+
+						By("Validating initialisedVolumes list")
+
+						err = compareInitialisedVolumes(oldInitVol, newInitVol)
+						Expect(err).ToNot(HaveOccurred())
```
abhishekdwivedi3060, line 422: This only covers scenario of Pvc UID change.
Shouldn't we add a scenarios/steps where initialisedVolumes contains both pvcname and pvcname@UID and init should cleanup pvcname from the list?

test/suite_test.go
```diff
@@ -132,6 +133,23 @@ func cleanupPVC(k8sClient client.Client, ns string) error {
 	return nil
 }
 
+func deletePVC(k8sClient client.Client, pvcNamespacedName types.NamespacedName) error {
+	pvc := &corev1.PersistentVolumeClaim{}
+	if err := k8sClient.Get(goctx.TODO(), pvcNamespacedName, pvc); err != nil {
```
abhishekdwivedi3060, line 138: Check notFound error

abhishekdwivedi3060 (changes requested):

test/rack_utils.go
abhishekdwivedi3060: Can rename the existing func`getClusterNamespacedName` to `getNamespacedName`

test/storage_init_test.go
```diff
@@ -338,9 +340,215 @@ var _ = Describe(
...
+					func() {
+						zones, err := getZones(ctx, k8sClient)
+						Expect(err).ToNot(HaveOccurred())
+
+						zone1 := zones[0]
+						zone2 := zones[0]
+						racks := []asdbv1.Rack{
+							{ID: 1, Zone: zone1},
+							{ID: 2, Zone: zone2},
+						}
```
abhishekdwivedi3060, line 361: If we are deploying in the same zone, then I think there is no need to set the zone, rack id is sufficient

abhishekdwivedi3060 (approved):

main.go
```diff
@@ -158,38 +170,220 @@ func main() {
...
+
+		if resp.StatusCode == http.StatusOK {
+			setupLog.Info("Controller is ready")
+			resp.Body.Close()
+
+			break
+		}
+
+		setupLog.Info("Controller is not ready, retrying...")
+		resp.Body.Close()
```
sud82, line 226: How many retry it usually takes?...should we add a small sleep before retrying?

sud82 (changes requested):

main.go
```diff
@@ -158,38 +170,220 @@ func main() {
...
+				}
+				// Error reading the object.
+				return err
+			}
+
+			var patches []jsonpatch.PatchOperation
+
+			for podIdx := range podList.Items {
+				pod := &podList.Items[podIdx]
+				initializedVolumes := aeroCluster.Status.Pods[pod.Name].InitializedVolumes
```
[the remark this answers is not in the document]
abhishekdwivedi3060: As it is a map, it'll return the zero value for the type `AerospikePodStatus{}` and in turn will return nil slice

test/storage_init_test.go
```diff
@@ -338,9 +340,215 @@ var _ = Describe(
...
+		if _, ok := newInitVolMap[initVol[0]]; ok {
+			return fmt.Errorf("found more than one occurrences of same volume in initialisedVolume list")
+		}
+
+		newInitVolMap[initVol[0]] = initVol[1]
+	}
+
+	for idx := range oldInitVol {
+		initVol := strings.Split(oldInitVol[idx], "@")
+		if initVol[0] == "ns" {
```
sud82, line 527: If `ns` is not found then also this will pass. Add a check.

sud82 (changes requested):

test/storage_init_test.go
```diff
@@ -338,9 +340,215 @@ var _ = Describe(
...
+						patches = append(patches, patch1)
+
+						jsonPatchJSON, err := json.Marshal(patches)
+						Expect(err).ToNot(HaveOccurred())
+
+						constantPatch := crClient.RawPatch(types.JSONPatchType, jsonPatchJSON)
+
+						err = k8sClient.Status().Patch(
+							ctx, aeroCluster, constantPatch, crClient.FieldOwner("pod"),
+						)
```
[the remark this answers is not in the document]
tanmayja: both ways are solving the same purpose in this case.

main.go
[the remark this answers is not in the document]
sud82: If the pod is there but it's not updated in status then it's an invalid situation. we should fail here and retry again.

main.go
```diff
@@ -158,38 +170,220 @@ func main() {
...
 		os.Exit(1)
 	}
 
 	if err := mgr.AddReadyzCheck("check", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up ready check")
+		setupLog.Error(err, "Unable to set up ready check")
 		os.Exit(1)
 	}
 
+	go migrateAerospikeClusters(client, &options, watchNs)
```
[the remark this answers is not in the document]
tanmayja: We can only migrate clusters after manager, webhooks are all setup. which is the part of `mgr.Start()` call. And this is the blocking call which means code written after this will not get executed. Hence wrote it as a routine which starts before `mgr.Start()` and wait for the controller to be ready internally.
2- No, there is no dependency between reconciler and this go routine.

main.go
tanmayja: Actually, I have not seen a single retry in my many runs. This is mostly a safe check.

PR #231, opened 2023-07-04
Bug fix: Handle scale down if there are failed pods in the Rack

sud82:


controllers/rack.go
```diff
@@ -379,7 +387,7 @@ func (r *SingleClusterReconciler) upgradeOrRollingRestartRack(found *appsv1.Stat
 			return found, reconcileError(nErr)
 		}
 
-		if needRollingRestartRack || len(failedPods) != 0 {
+		if needRollingRestartRack {
```
tanmayja, line 390: If user update the CR back to original value to recover the cluster, in that case spec and status will be same. in that case will 'r.rollingRestartRack' get triggered?

controllers/rack.go
```diff
@@ -86,14 +86,15 @@ func (r *SingleClusterReconciler) reconcileRacks() reconcileResult {
 
 		failedPods, _ := getFailedAndActivePods(podList)
 		if len(failedPods) != 0 {
-			r.Log.Info("Upgrading or rolling restart rack with failed pods", "rackID", state.Rack.ID, "failedPods", failedPods)
+			r.Log.Info("Reconcile the failed pods in the Rack", "rackID", state.Rack.ID, "failedPods", failedPods)
 
-			_, res = r.upgradeOrRollingRestartRack(found, state, ignorablePods, failedPods)
-			if !res.isSuccess {
+			if res = r.reconcileRack(
```
tanmayja, line 91: Another approach:
- Here we are handling failed pods only, so can we fetch the scaled down pod names in failed pods and scale down those in a single go.
- reconcileRack will do scale-up also, should we do scale-up operation in failed pods loop or in next loop.

controllers/rack.go
sud82: No, it won't. Now I remember why we were restarting failed pods always...

controllers/rack.go
sud82: - Does it any much value? In case of scale down of failed pod, we don't wait anywhere so even the current flow should be quick.
- Thinking about scale-up... Not decided

tanmayja: we may not be waiting before downsizing the Sts in `scaleDownRack ` but [here](https://github.com/aerospike/aerospike-kubernetes-operator/blob/8593a827e046b934d4c8dbc3a238275e63b9a279/controllers/rack.go#L798) we are waiting for the Sts to be ready after each scale-down, which might be time taking.

controllers/rack.go
```diff
@@ -785,39 +804,42 @@ func (r *SingleClusterReconciler) scaleDownRack(
...
-	// If scale down leads to unavailable or dead partition then we should scale up the cluster,
-	// This can be left to the user but if we would do it here on our own then we can reuse
-	// objects like pvc and service. These objects would have been removed if scaleup is left for the user.
-	// In case of rolling restart, no pod cleanup happens, therefor rolling config back is left to the user.
-	if err = r.validateSCClusterState(policy, ignorablePods); err != nil {
-		// reset cluster size
-		newSize := *found.Spec.Replicas + 1
-		found.Spec.Replicas = &newSize
+	// No need for these checks if pod is not running. These checks will fail if there is any other pod in failed state.
+	if isPodRunningAndReady {
```
tanmayja, line 808: should we get the running and ready status again, instead of using the old variable.

controllers/rack.go
```diff
@@ -908,8 +930,7 @@ func (r *SingleClusterReconciler) rollingRestartRack(found *appsv1.StatefulSet,
 	}
```
tanmayja, line 930: at line 916, is it possible to have a pod with image failed state if Len(failed pods) is 0?

controllers/rack.go
```diff
@@ -86,14 +86,28 @@ func (r *SingleClusterReconciler) reconcileRacks() reconcileResult {
...
+				found, state, ignorablePods, failedPods,
+			); !res.isSuccess {
+				return res
+			}
+
+			r.Log.Info("Reconciled the failed pods in the Rack", "rackID", state.Rack.ID, "failedPods", failedPods)
+		}
+
+		// Use old podList as if there are new failed pods then they will not recover with this rolling restart
+		failedPods, _ = getFailedAndActivePods(podList)
```
tanmayja, line 101: Here we are using old podList, can there be a case where due to scale down happend above, pod got deleted but podList will have the object of same.

controllers/rack.go
tanmayja, line 1: in the definition of rollingRestartPods(). Is it possible now to have failed and active pods both simultaneously in the pod batch. If no, we can modify accordingly.

controllers/rack.go
abhishekdwivedi3060: After every scaledown we requeue the request, so I don't think that above scenario will occur

tanmayja: got it, then there is no reason of fetching `failedPods` again here as `podlist` is same and `getFailedAndActivePods` has all static checks on the list.

controllers/rack.go
```diff
@@ -752,22 +768,25 @@ func (r *SingleClusterReconciler) scaleDownRack(
...
-	// temporarily add the pod to be deleted in ignorablePods slice to avoid setting migrate-fill-delay to pod when
-	// pod is not ready
-	if res := r.setMigrateFillDelay(
-		policy, &rackState.Rack.AerospikeConfig, true,
-		append(ignorablePods, *pod),
-	); !res.isSuccess {
-		return found, res
+		// set migrate-fill-delay to 0 across all nodes of cluster to scale down fast
+		// temporarily add the pod to be deleted in ignorablePods slice to avoid setting migrate-fill-delay to pod when
+		// pod is not ready
```
abhishekdwivedi3060, line 783: Need to update the comment here as we are not setting MFD only in-case of running pod
```suggestion
		// temporarily add the pod to be deleted in ignorablePods slice to avoid setting migrate-fill-delay to pod being deleted
```

controllers/rack.go
abhishekdwivedi3060: This is for scenario CR is updated and pods go in pending/failed state and then again CR is updated to revert the changes. In this case spec and status are same and above flow won't handle failed pod. This condition is added separately outside to handle scale-up even with failed pods.

tanmayja: yes I know, here my point is about calling `failedPods, _ = getFailedAndActivePods(podList)`. Is it ever gonna give you different failedPods than above line 87?

controllers/rack.go
sud82: Added it after scaleup

controllers/rack.go
sud82: Now we don't wait.

controllers/rack.go
sud82: rolling restart should recover failed pods

controllers/rack.go
sud82: Let's keep it this way for now. If we update it then we have to update the upgrade path too in a similar way.

abhishekdwivedi3060 (approved):

tanmayja (approved):

PR #232, opened 2023-07-05
Moving KO-235 changes from main.go to reconciler

tanmayja:


controllers/statefulset.go
```diff
@@ -1081,44 +1081,52 @@ func (r *SingleClusterReconciler) updateContainerImages(statefulset *appsv1.Stat
...
+				r.aeroCluster, container.Name,
 			)
+			if err != nil {
+				return err
+			}
 
-			statefulSet.Spec.Template.Spec.InitContainers[idx].Image = desiredImage
+			if !utils.IsImageEqual(container.Image, desiredImage) {
+				r.Log.Info(
+					"Updating image in statefulset spec", "container",
```
abhishekdwivedi3060, line 1107: Should we add sts name in the info msg?

test/storage_init_test.go
```diff
@@ -338,9 +340,212 @@ var _ = Describe(
...
+							Value:     append(oldInitVol, "ns"),
+						}
+
+						var patches []jsonpatch.PatchOperation
+						patches = append(patches, patch1)
+
+						jsonPatchJSON, err := json.Marshal(patches)
+						Expect(err).ToNot(HaveOccurred())
+
+						constantPatch := crClient.RawPatch(types.JSONPatchType, jsonPatchJSON)
```
abhishekdwivedi3060, line 466: We can modify this test-case as per new flow like add old format pvcname and remove api-version label. Reconcile should add label and migrate volume names. Validate both initvolumes name and label

abhishekdwivedi3060 (changes requested):

PR #233, opened 2023-07-10
KO-165: Added support for simultaneous scale-down and rolling restart SC cluster

abhishekdwivedi3060:
- Added support for simultaneous scale-down and rolling restart SC cluster


controllers/strong_consistency.go
```diff
@@ -30,10 +32,58 @@ func (r *SingleClusterReconciler) validateSCClusterState(policy *as.ClientPolicy
...
+}
+
+func (r *SingleClusterReconciler) getIgnorableNamespaces(allHostConns []*deployment.HostConn) (
+	sets.Set[string], error) {
+	removedNSes, err := r.removedNamespaces(allHostConns)
+	if err != nil {
+		return nil, err
+	}
+
+	addSCNamespaces, err := r.addedSCNamespaces(allHostConns)
```
tanmayja, line 80: Is it ok to get sc namespaces from all the hosts even if there are some entries in `r.aeroCluster.Spec.RosterNodeBlockList`? similarly in `validateSCClusterState`

controllers/strong_consistency.go
```diff
@@ -30,10 +32,58 @@ func (r *SingleClusterReconciler) validateSCClusterState(policy *as.ClientPolicy
...
+		nodeNamespaces := sets.NewString(namespaces...)
+		newAddedSCNamespaces.Insert(specSCNamespaces.Difference(nodeNamespaces).List()...)
+	}
+
+	return newAddedSCNamespaces.List(), nil
+}
+
+func (r *SingleClusterReconciler) getIgnorableNamespaces(allHostConns []*deployment.HostConn) (
+	sets.Set[string], error) {
+	removedNSes, err := r.removedNamespaces(allHostConns)
```
tanmayja, line 75: we call `deployment.GetClusterNamespaces` in both `removedNamespaces` as well as `r.addedSCNamespaces`, see if we can reduce calls here.

controllers/strong_consistency.go
abhishekdwivedi3060: As per discussion, not much useful. So skipping extra checks for RosterNodeBlockList

sud82 (approved):

PR #237, opened 2023-07-19
Allow TLS configuration update wherever possible.

tanmayja:
Allow TLS update in certain scenarios like:
1 - ca-file to ca-path change in existing and in-use TLS config.
2- Adding, removing and modifying TLS config, which is not being used in any network conf.

### DIFF
--- a/api/v1/aerospikecluster_validating_webhook.go
+++ b/api/v1/aerospikecluster_validating_webhook.go
@@ -1451,14 +1451,12 @@ func validateTLSUpdate(oldConf, newConf map[string]interface{}) error {
 
 			if (oldCAFileOK || oldCAPathOK) && !(newCAPathOK || newCAFileOK) {
 				return fmt.Errorf(
-					"cannot remove `ca-file` or `ca-path` from tls",
+					"cannot remove used `ca-file` or `ca-path` from tls",
 				)
 			}
 
-			if oldCAFileOK {
-				if newCAFileOK && newCAFile.(string) != oldCAFile {
-					return fmt.Errorf("cannot change ca-file of used tls")
-				}
+			if oldCAFileOK && newCAFileOK && newCAFile.(string) != oldCAFile {
+				return fmt.Errorf("cannot change ca-file of used tls")
 			}
 		}
 	}
--- a/api/v1/aerospikecluster_validating_webhook.go
+++ b/api/v1/aerospikecluster_validating_webhook.go
@@ -29,6 +29,7 @@ import (
 
 	validate "github.com/asaskevich/govalidator"
 	"github.com/go-logr/logr"
+	boolean "github.com/golangf/extra-boolean"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -834,13 +835,14 @@ func (c *AerospikeCluster) validateNetworkConfig(networkConf map[string]interfac
 				tlsNames[tlsName.(string)] = struct{}{}
 			}
 
-			if _, ok := tlsConf["ca-path"]; ok {
-				if _, ok1 := tlsConf["ca-file"]; ok1 {
-					return fmt.Errorf(
-						"both `ca-path` and `ca-file` cannot be set in `tls`. tlsConf %v",
-						tlsConf,
-					)
-				}
+			_, CAPathOK := tlsConf["ca-path"]
+			_, caFileOK := tlsConf["ca-file"]
+
+			if boolean.Xnor(CAPathOK, caFileOK) {
+				return fmt.Errorf(
+					"only one of (`ca-path` and `ca-file`) must be set in `tls`. tlsConf %v",
+					tlsConf,
+				)
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1396,6 +1392,47 @@ func validateAerospikeConfigUpdate(
...
+		}
+
+		for _, tls := range oldTLS.([]interface{}) {
+			tlsMap := tls.(map[string]interface{})
+
+			oldCAFile, oldCAFileOK := tlsMap["ca-file"]
+			if oldCAFileOK {
+				oldTLSCAFileMap[tlsMap["name"].(string)] = oldCAFile.(string)
+			}
+		}
```
abhishekdwivedi3060, line 1420: I think this oldTLSCAFileMap should be populated for only those TLS which were used earlier in any sub-context. If TLS is just added and not used then we can allow any type of update on that

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1396,6 +1392,47 @@ func validateAerospikeConfigUpdate(
...
+
+		for _, tls := range newTLS.([]interface{}) {
+			tlsMap := tls.(map[string]interface{})
+			if caFile, ok := oldTLSCAFileMap[tlsMap["name"].(string)]; ok {
```
abhishekdwivedi3060, line 1424: Check `usedTLS.Has(tlsMap["name"].(string))` before if condition and continue if not used

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1396,6 +1392,47 @@ func validateAerospikeConfigUpdate(
...
+				newCAFile, newCAFileOK := tlsMap["ca-file"]
+				if newCAFileOK && newCAFile.(string) != caFile && usedTLS.Has(tlsMap["name"].(string)) {
+					return fmt.Errorf("cannot change ca-file of used tls")
+				}
```
abhishekdwivedi3060, line 1428: Check if new TLS has `ca-path` set or not when `newCAFileOK` is false

abhishekdwivedi3060 (changes requested):

test/cluster_test.go
```diff
@@ -331,7 +336,176 @@ func DeployClusterWithSyslog(ctx goctx.Context) {
...
+					Expect(err).ToNot(HaveOccurred())
+
+					network := aeroCluster.Spec.AerospikeConfig.Value["network"].(map[string]interface{})
+					tlsList := network["tls"].([]interface{})
+					usedTLS := tlsList[0].(map[string]interface{})
+					usedTLS["name"] = "aerospike-a-0.test-runner2"
+					tlsList[0] = usedTLS
+					network["tls"] = tlsList
+					aeroCluster.Spec.AerospikeConfig.Value["network"] = network
+					err = updateCluster(k8sClient, ctx, aeroCluster)
```
abhishekdwivedi3060, line 484: We should call `updateClusterWithTO` with small timeout to reduce test time as we are sure test-case is going to fail. Change wherever applicable

test/cluster_test.go
```diff
@@ -331,7 +336,176 @@ func DeployClusterWithSyslog(ctx goctx.Context) {
...
+					network = aeroCluster.Spec.AerospikeConfig.Value["network"].(map[string]interface{})
+					tlsList = network["tls"].([]interface{})
+					usedTLS = tlsList[0].(map[string]interface{})
+					usedTLS["ca-file"] = "/etc/aerospike/secret/fb_cert.pem"
+					tlsList[0] = usedTLS
+					network["tls"] = tlsList
+					aeroCluster.Spec.AerospikeConfig.Value["network"] = network
+					err = updateCluster(k8sClient, ctx, aeroCluster)
+					Expect(err).Should(HaveOccurred())
+				},
```
abhishekdwivedi3060, line 502: Should we add steps to check when both ca-file and ca-path are updated in cluster?

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1396,6 +1390,82 @@ func validateAerospikeConfigUpdate(
...
+
+			_, newCAPathOK := tlsMap["ca-path"]
+			newCAFile, newCAFileOK := tlsMap["ca-file"]
+
+			oldCAFile, oldCAFileOK := oldTLSCAFileMap[tlsMap["name"].(string)]
+			_, oldCAPathOK := oldTLSCAPathMap[tlsMap["name"].(string)]
+
+			if (oldCAFileOK || oldCAPathOK) && !(newCAPathOK || newCAFileOK) {
+				return fmt.Errorf(
+					"cannot remove `ca-file` or `ca-path` from tls",
```
abhishekdwivedi3060, line 1454: nit: 
```suggestion
					"cannot remove used `ca-file` or `ca-path` from tls",
```

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1396,6 +1390,82 @@ func validateAerospikeConfigUpdate(
...
+				)
+			}
+
+			if oldCAFileOK {
+				if newCAFileOK && newCAFile.(string) != oldCAFile {
```
abhishekdwivedi3060, line 1459: Can be clubbed into one

abhishekdwivedi3060 (changes requested):

test/cluster_test.go
tanmayja: When webhook fails(which is this case), updateCluster is not gonna take much time.

abhishekdwivedi3060 (approved):

sud82 (approved):

PR #250, opened 2023-09-13
[KO-259] Changing ports for headless service.

tanmayja:
Changing ports for headless service when moving tls to non-tls and vice versa


controllers/service.go
```diff
@@ -381,7 +392,7 @@ func isServiceTLSChanged(specAeroConf, statusAeroConf *asdbv1.AerospikeConfigSpe
 	specSVCPort := asdbv1.GetServicePort(specAeroConf)
 	statusSVCPort := asdbv1.GetServicePort(statusAeroConf)
 
-	return specTLSName != statusTLSName || specSVCPort != statusSVCPort
+	return specTLSName != statusTLSName || *specSVCPort != *statusSVCPort
```
abhishekdwivedi3060, line 395: Won't this pointer de-reference throw panic when port is removed/added when TLS is already set.

controllers/rack.go
```diff
@@ -646,6 +646,12 @@ func (r *SingleClusterReconciler) upgradeRack(statefulSet *appsv1.StatefulSet, r
 		}
 	}
 
+	err = r.updateSTSHeadlessSvc()
```
abhishekdwivedi3060, line 649: Right now headless svc is created once in `createEmptyRack` func for the 1st rack and skipped for the rest of the racks. As this service is tied to one CR and not rack specific, can we move this existing create code and new update code out of rack processing  and call only once in every reconcile like `createorUpdateHeadlessService`.
This way we don't have to call create and update multiple times at different places

abhishekdwivedi3060 (changes requested):

controllers/service.go
```diff
@@ -21,77 +21,64 @@ func getSTSHeadLessSvcName(aeroCluster *asdbv1.AerospikeCluster) string {
...
-					// deprecates service.alpha.kubernetes.io/tolerate-unready-endpoints as of 1.
-					// 10? see: kubernetes/kubernetes#49239 Fixed in 1.11 as of #63742
-					PublishNotReadyAddresses: true,
-					ClusterIP:                "None",
-					Selector:                 ls,
-				},
-			}
-
-			r.appendServicePorts(service)
+	err := r.updateServicePorts(serviceName, r.aeroCluster.Namespace)
```
abhishekdwivedi3060, line 29: This flow doesn't seem much readable. Instead we can get the service first here and based on that can go with create or update flow similar to old code

controllers/service.go
```diff
@@ -182,6 +146,23 @@ func (r *SingleClusterReconciler) createSTSLoadBalancerSvc() error {
...
+	if err := r.Client.Update(
+		context.TODO(), service, updateOption,
+	); err != nil {
+		return fmt.Errorf(
+			"failed to update service %s: %v", serviceName, err,
+		)
+	}
+
 	r.Log.Info(
 		"LoadBalancer Service already exist for cluster. Using existing service",
```
abhishekdwivedi3060, line 167: Update Info msg saying updated. Also would be better to add similar msg saying 
`LoadBalancer Service already exist for cluster, trying to update if required"` before the update code flow

controllers/service.go
```diff
@@ -182,6 +146,23 @@ func (r *SingleClusterReconciler) createSTSLoadBalancerSvc() error {
 		return err
 	}
 
+	if len(service.Spec.Ports) == 1 && service.Spec.Ports[0].Port == servicePort.Port &&
```
abhishekdwivedi3060, line 149: There are other fields also in LB struct like `LoadBalancerSourceRanges`, `ExternalTrafficPolicy` and `Annotations`. Shouldn't we handle those as well as part of update check?

controllers/service.go
```diff
@@ -190,47 +171,46 @@ func (r *SingleClusterReconciler) createSTSLoadBalancerSvc() error {
...
 
-func (r *SingleClusterReconciler) createPodService(pName, pNamespace string) error {
-	service := &corev1.Service{}
-	if err := r.Client.Get(
-		context.TODO(),
-		types.NamespacedName{Name: pName, Namespace: pNamespace}, service,
-	); err == nil {
-		return errors.NewAlreadyExists(schema.GroupResource{Resource: "Service"}, service.Name)
+func (r *SingleClusterReconciler) createOrUpdatePodService(pName, pNamespace string) error {
+	err := r.updateServicePorts(pName, pNamespace)
```
abhishekdwivedi3060, line 175: Same readable code flow comment here also

abhishekdwivedi3060 (changes requested):

controllers/service.go
tanmayja: validation web hook does not allow any modification in loadbalancer spec.

controllers/service.go
```diff
@@ -270,7 +276,7 @@ func (r *SingleClusterReconciler) updateServicePorts(sName, sNamespace string) e
 		context.TODO(), service, updateOption,
 	); err != nil {
 		return fmt.Errorf(
-			"failed to update service %s: %v", sName, err,
+			"failed to update service %s: %v", service.Name, err,
```
abhishekdwivedi3060, line 279: Can we add logs when we are updating service similar to LB service logs

abhishekdwivedi3060 (approved):
Can we add/modify test-cases for testing the service update related changes?

controllers/service.go
abhishekdwivedi3060: As per discussion, not allowing LB service update through CR for now

tanmayja:
> Can we add/modify test-cases for testing the service update related changes?

Adding tests to validate pod service port and headless service port, to keep it simple.

test/utils.go
```diff
@@ -53,6 +53,9 @@ const aerospikeNs string = "aerospike"
 const zoneKey = "topology.kubernetes.io/zone"
 const regionKey = "topology.kubernetes.io/region"
 
+const tlsPort = 4333
+const nontlsPort = 3000
```
abhishekdwivedi3060, line 57: nit: Better to use service prefix to denote service tls and non-tls Port
Also, use these constants at other places also in the test-suites where hardcoding is done for these values

test/cluster_helper.go
```diff
@@ -381,6 +381,46 @@ func rollingRestartClusterByRemovingNamespaceDynamicallyTest(
 	return updateCluster(k8sClient, ctx, aeroCluster)
 }
 
+func validateServiceUpdate(k8sClient client.Client, ctx goctx.Context,
```
abhishekdwivedi3060, line 384: I think service should be validated twice to check service ports are not overwritten :
1. Validate service when both tls and non-tls are configured in first step 
2. Validate service when final tls/non-tls is configured

Above scenarios are valid for both non-tls to tls migration and vice-versa

abhishekdwivedi3060 (changes requested):

test/cluster_helper.go
```diff
@@ -381,6 +392,45 @@ func rollingRestartClusterByRemovingNamespaceDynamicallyTest(
...
+
+		for _, p := range service.Spec.Ports {
+			if portSet.Has(p.Port) {
+				portSet.Delete(p.Port)
+			}
+		}
+
+		if portSet.Len() > 0 {
+			return fmt.Errorf("service %s port not configured correctly", serviceNamespacesName.Name)
+		}
```
abhishekdwivedi3060, line 428: nit: why not just check and move the portSet outside of for loop
```
if !portSet.Has(p.Port) {
				return fmt.Errorf()
			}
```

abhishekdwivedi3060 (approved):

test/cluster_helper.go
tanmayja: multiple ways to implement this anyway, here just didn't want to return error if any port is also available in port list other than the required ones. I know in test cases it will not happen.

sud82 (approved):

PR #257, opened 2023-11-20
KO-285: Allow Operator operations with failed pods

abhishekdwivedi3060:


test/cluster_test.go
```diff
@@ -45,6 +45,11 @@ var _ = Describe(
 		// 		DeployClusterWithSyslog(ctx)
 		// 	},
 		// )
+		Context(
+			"DeployClusterWithSyslog", func() {
```
abhishekdwivedi3060, line 49: Will fix this typo in next commit.

test/cluster_test.go
```diff
@@ -124,6 +129,92 @@ func ScaleDownWithMigrateFillDelay(ctx goctx.Context) {
...
+					aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+					aeroCluster.Spec.Image = baseImage + ":6.4.0.4"
+					err = updateCluster(k8sClient, ctx, aeroCluster)
+					Expect(err).ToNot(HaveOccurred())
+
+					By("Scale up")
+					aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+					aeroCluster.Spec.IgnorePodList = []string{ignorePodName}
```
tanmayja, line 201: any reason of setting it again after setting it at line 179? AFAIK only user can remove the name from IgnorePodList right?

test/cluster_test.go
```diff
@@ -124,6 +129,92 @@ func ScaleDownWithMigrateFillDelay(ctx goctx.Context) {
 	)
 }
 
+func clusterWithIgnorePodList(ctx goctx.Context) {
```
tanmayja, line 132: can we add a test after removing the name from list and make that pod running to check if updated config is correctly reflected on that pod as well?

test/cluster_test.go
abhishekdwivedi3060: Not required.

controllers/aero_info_calls.go
```diff
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
```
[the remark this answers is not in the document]
abhishekdwivedi3060: golang-set is only used in SC file for management-lib changes. 
As discussed, we won't change existing code to golang-set in this PR, so adding new changes with golang-set was causing confusion. So kept everything on apimachinery set pkg and will move to golang-set in one go.

controllers/pod.go
```diff
@@ -345,9 +350,14 @@ func (r *SingleClusterReconciler) ensurePodsRunningAndReady(podsToCheck []*corev
 	return reconcileRequeueAfter(10)
 }
 
-func getFailedAndActivePods(pods []*corev1.Pod) (failedPods, activePods []*corev1.Pod) {
+func getFailedAndActivePods(pods []*corev1.Pod, ignorablePodNames sets.Set[string],
+) (failedPods, activePods []*corev1.Pod) {
 	for idx := range pods {
 		pod := pods[idx]
+		if ignorablePodNames.Has(pod.Name) {
```
tanmayja, line 357: we calls `getFailedAndActivePods` at 4 places, at 2 places we make sure that there are no ignorable pods present in pods list passed to this function, and in other two places from `reconcileRacks()`, there are chances of having ignorable pods in pods list. Can we write it better and remove check `if ignorablePodNames.Has(pod.Name) {` from `getFailedAndActivePods` if possible.

PR #258, opened 2023-11-23
Add 7.0.0.0 schema and update samples, test-cases as per latest schema

abhishekdwivedi3060:


config/samples/tls_cluster_cr.yaml
```diff
@@ -77,11 +77,9 @@ spec:
 
     namespaces:
       - name: bar
-        memory-size: 3000000000
         replication-factor: 1
         storage-engine:
           type: device
```
tanmayja, line 82: Here `type`: should be `memory`.

abhishekdwivedi3060: I have intentionally removed the data-in-memory wherever the functionality is not related to data-in-memory

PR #262, opened 2023-12-11
Adding support for dynamic config change.

tanmayja:
This PR is to handle aerospike config fields which can be changed dynamically without any rolling restart.
whenever user will change any config in CR, operator will do certain operations:

Compare the spec and status config from CR and fetch different config in a specific format.
Check whether that field is static or dynamic.
If dynamic, construct the set-config command based on the context and sub-context.
Execute the command in all the hosts via asinfo.
If command fails for any reason, fall back to rolling restart.


abhishekdwivedi3060 (changes requested):

controllers/pod.go
```diff
@@ -71,24 +87,39 @@ func (r *SingleClusterReconciler) getRollingRestartTypeMap(
...
 		}
 
-		restartType := r.getRollingRestartTypePod(rackState, pods[idx], confMap, addedNSDevices)
+		restartTypeMap[pods[idx].Name] = r.getRollingRestartTypePod(rackState, pods[idx], confMap, addedNSDevices,
+			len(dynamicConfDiffPerPod[pods[idx].Name]) > 0)
 
-		restartTypeMap[pods[idx].Name] = restartType
+		if podStatus.DynamicConfigFailed {
+			restartTypeMap[pods[idx].Name] = mergeRestartType(restartTypeMap[pods[idx].Name], quickRestart)
+		}
```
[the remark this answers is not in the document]
tanmayja: No, if hash is same, then may be user has reverted the changes after dynamic failure happened. In that case if DynamicConfigFailed is true then instead of no-restart, quick restart should happen.

controllers/pod.go
```diff
@@ -1212,3 +1284,168 @@ func (r *SingleClusterReconciler) getConfigMap(rackID int) (*corev1.ConfigMap, e
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load config map by lib: %v", err)
+	}
+
+	asConf, err := asconfig.NewMapAsConfig(r.Log, rackState.Rack.AerospikeConfig.Value)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load config map by lib: %v", err)
+	}
+
+	asConfSpec, err := getFlatConfig(r.Log, asConf.ToConfFile())
```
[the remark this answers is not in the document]
tanmayja: when we convert confirmation file(annotation) to flat config, output is not similar to rack config, for example node-address-ports field has ip and port separated by space in rack config but converted flat conf from file has ":" as delimiter. So to resolve these things, had to run same steps on both kind of input.

controllers/pod.go
```diff
@@ -71,24 +86,39 @@ func (r *SingleClusterReconciler) getRollingRestartTypeMap(
...
+				}
+			}
+
+			// If dynamic commands have failed in previous retry, then we should not try to update config dynamically.
+			if !podStatus.DynamicConfigFailed {
+				// Fetching all dynamic config change.
+				dynamicConfDiffPerPod[pods[idx].Name], err = r.handleDynamicConfigChange(rackState, pods[idx])
+				if err != nil {
+					return nil, nil, err
+				}
```
abhishekdwivedi3060, line 104: Add a check here or wherever required to not to follow dynamic flow for server versions < 7.0

test/dynamic_config_test.go
```diff
@@ -0,0 +1,264 @@
...
+func validateServerRestart(ctx goctx.Context, cluster *asdbv1.AerospikeCluster, pidMap map[string]podID,
+	shouldRestart bool) {
+	restarted := false
+
+	newPodPidMap, err := getPodIDs(ctx, cluster)
+	Expect(err).ToNot(HaveOccurred())
+
+	for podName, pid := range pidMap {
+		if newPodPidMap[podName].podUID != pid.podUID || newPodPidMap[podName].asdPID != pid.asdPID {
+			restarted = true
```
abhishekdwivedi3060, line 223: Break from here

test/dynamic_config_test.go
```diff
@@ -0,0 +1,264 @@
...
+						Expect(err).ToNot(HaveOccurred())
+
+						pod := aeroCluster.Status.Pods["dynamic-config-test-0-0"]
+
+						conf, err := getAerospikeConfigFromNode(logger, k8sClient, ctx, clusterNamespacedName,
+							"security", &pod)
+						Expect(err).ToNot(HaveOccurred())
+
+						enableQuotas, ok := conf["enable-quotas"].(bool)
+						Expect(ok).ToNot(BeFalse())
```
abhishekdwivedi3060, line 195: This seems redundant as we are fetching info from node and the default value of that field itself is false.
I think its better to set it to true before update or use different field for testing

abhishekdwivedi3060 (approved):

sud82 (changes requested):

abhishekdwivedi3060 (approved):

sud82 (approved):

PR #273, opened 2024-02-19
[KO-290] Allow enabling security in the existing deployed cluster

tanmayja:


controllers/reconciler.go
```diff
@@ -144,16 +172,22 @@ func (r *SingleClusterReconciler) Reconcile() (ctrl.Result, error) {
...
 	); err != nil {
 		r.Log.Error(err, "Failed to check for Quiesced nodes")
 		return reconcile.Result{}, err
 	}
 
+	podList, err := r.getClusterPodList()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
```
abhishekdwivedi3060, line 187: Can we move this inside `validateAndReconcileAccessControl` like if `securityEnabledPods` is nil then get cluster pods list otherwise use the given list?
This way we don't have to get pod list in case of security disabled cluster

controllers/reconciler.go
```diff
@@ -919,3 +960,30 @@ func (r *SingleClusterReconciler) AddAPIVersionLabel(ctx context.Context) error
 
 	return r.Client.Update(ctx, aeroCluster, updateOption)
 }
+
+func (r *SingleClusterReconciler) getsecurityEnabledPods() ([]corev1.Pod, error) {
```
abhishekdwivedi3060, line 964:
```suggestion
func (r *SingleClusterReconciler) getSecurityEnabledPods() ([]corev1.Pod, error) {
```

abhishekdwivedi3060 (approved):

controllers/reconciler.go
tanmayja: we can do that, but passing nil as `securityEnabledPods` argument kind of looks like there is no security enabled pods in cluster even if all pods are security enabled. Just for the readability I did this way.

abhishekdwivedi3060 (approved):

controllers/access_control.go
```diff
@@ -37,41 +37,39 @@ func AerospikeAdminCredentials(
...
+		// It is possible that this is a new cluster and current state is empty.
+		currentSecurityEnabled, currentSecurityErr = asdbv1.IsSecurityEnabled(
+			outgoingVersion, currentState.AerospikeConfig,
+		)
+	} else {
+		currentSecurityErr = outgoingVersionErr
+	}
 
-		enabled, newErr = asdbv1.IsSecurityEnabled(
+	incomingVersion, incomingVersionErr := asdbv1.GetImageVersion(desiredState.Image)
```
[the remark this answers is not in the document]
tanmayja: The idea here was to send Aerospike credentials if security is enabled in either the spec or status. Following that, we check for `AerospikeAccessControl` in the `status` to determine whether to use user-provided credentials or the default admin credentials.

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1346,11 +1347,22 @@ func validateSecurityContext(
...
 
+	isSecurityEnabledPodExist := false
+
+	for pod := range podStatus {
+		if podStatus[pod].IsSecurityEnabled {
+			isSecurityEnabledPodExist = true
+			break
+		}
+	}
+
```
abhishekdwivedi3060, line 1362: I think its better to move this code inside `validateSecurityConfigUpdate` func

abhishekdwivedi3060 (changes requested):

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -119,7 +119,7 @@ func (c *AerospikeCluster) ValidateUpdate(oldObj runtime.Object) (admission.Warn
 	if err := validateAerospikeConfigUpdate(
 		aslog, incomingVersion, outgoingVersion,
 		c.Spec.AerospikeConfig, old.Spec.AerospikeConfig,
-		c.Status.AerospikeConfig,
+		c.Status.AerospikeConfig, c.Status.Pods,
```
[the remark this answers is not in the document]
tanmayja: This function is being called from `validateRackUpdate` where instead of passing `c.Status.AerospikeConfig`, we pass `rackStatusConfig`. So here we can pass the whole status object but in rack-specific validation, it will be difficult.

api/v1/aerospikecluster_validating_webhook.go
tanmayja: did not want to pass podStatus map unnecessarily, hence passing `security` related info only.

abhishekdwivedi3060 (approved):

PR #279, opened 2024-03-25
K0-314: Added support for batch scale-down

abhishekdwivedi3060:


controllers/rack.go
```diff
@@ -940,7 +960,7 @@ func (r *SingleClusterReconciler) scaleDownRack(
 
 	// No need for these checks if pod was not running.
```
tanmayja, line 961: can change comments like pod -> pods etc..

tanmayja (approved):

test/batch_scaledown_pods_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	return waitForClusterScaleDown(
+		k8sClient, ctx, aeroCluster,
+		int(aeroCluster.Spec.Size), retryInterval,
+		getTimeout(aeroCluster.Spec.Size),
+	)
+}
+
+func isBatchScaleDown(aeroCluster *asdbv1.AerospikeCluster, scaleDownBatchSize int) bool {
```
sud82, line 132: Review this.

sud82 (changes requested):

PR #281, opened 2024-03-29
BugFix: Updating read and write quotas in cluster.

tanmayja:


abhishekdwivedi3060 (approved):

controllers/access_control.go
```diff
@@ -576,6 +576,16 @@ func (roleCreate aerospikeRoleCreateUpdate) updateRole(
 		}
 	}
 
+	if role.ReadQuota != roleCreate.readQuota || role.WriteQuota != roleCreate.writeQuota {
```
sud82, line 579: Add a test for this change.

sud82 (changes requested):

controllers/access_control.go
sud82: Setting, Re-setting

tanmayja: test is already there https://github.com/aerospike/aerospike-kubernetes-operator/blob/9d3edf738a8ca65031fd617fdd5a50b4fd036b81/test/access_control_test.go#L1896

PR #282, opened 2024-04-08
KO-319: Add github-action for releasing container image

abhishekdwivedi3060:


.github/workflows/docker-image-release.yaml
```diff
@@ -0,0 +1,60 @@
...
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Build and push image on Dockerhub
+        run: make docker-buildx IMG=${{ env.IMG_BASE }}:${{ env.TAG }} VERSION=${{ env.TAG }}
+
+      - name: Build and push image on Quay
+        run: make docker-buildx IMG=quay.io/${{ env.IMG_BASE }}:${{ env.TAG }} VERSION=${{ env.TAG }}
+
+      - name: Run Snyk image security scan
+        uses: snyk/actions/docker@master
```
sud82, line 44: Shouldn't we use specific version in place of master?

sud82 (approved):

.github/workflows/docker-image-release.yaml
abhishekdwivedi3060: Maybe we can. However, all the examples I've seen uses master tag: https://github.com/snyk/actions/tree/master/docker#snyk-docker-action

PR #283, opened 2024-04-10
Adding negative testcase for dynamic config

tanmayja:


test/dynamic_config_test.go
```diff
@@ -263,3 +516,193 @@ func getPodIDs(ctx context.Context, aeroCluster *asdbv1.AerospikeCluster) (map[s
 
 	return pidMap, nil
 }
+
+func updateValue(val interface{}) interface{} {
```
sud82, line 520: I think 
For bool type value we can flip it to the opposite for testing dynamic update. 
For string, we can try to update same value

test/dynamic_config_test.go
```diff
@@ -263,3 +516,193 @@ func getPodIDs(ctx context.Context, aeroCluster *asdbv1.AerospikeCluster) (map[s
...
+	case uint64:
+		return val2 + 1
+	case int64:
+		return val2 + 1
+	case float64:
+		return val2 + 1
+
+	case lib.Stats:
+
+	default:
```
sud82, line 539: Let's remove the nil type. We would know if there is any other type we are missing.

test/dynamic_config_test.go
```diff
@@ -263,3 +516,193 @@ func getPodIDs(ctx context.Context, aeroCluster *asdbv1.AerospikeCluster) (map[s
...
+) error {
+	newSpec := *flatSpec
+
+	for confKey, val := range *flatServer {
+		if asconfig.ContextKey(confKey) != "xdr" {
+			continue
+		}
+
+		tokens := strings.Split(confKey, ".")
+		if dyn := dynamic.Contains(asconfig.GetFlatKey(tokens)); dyn {
```
sud82, line 687: You can simply check `if dynamic.Contains(asconfig.GetFlatKey(tokens)) {}`
You are not using dyn anywhere. Same in other functions.

sud82 (changes requested):

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1550,13 +1550,8 @@ func validateNsConfUpdate(newConfSpec, oldConfSpec, currentStatus *AerospikeConf
...
 
 			if singleConf["name"] == oldSingleConf["name"] {
-				val, err := lib.CompareVersions(incomingVersion, Version6)
-				if err != nil {
-					return fmt.Errorf("failed to check image version: %v", err)
-				}
-
-				// For versions 6.0 and later, replication-factor is dynamic for AP namespaces (non strong-consistency).
-				if (IsNSSCEnabled(singleConf) || val < 0) && isValueUpdated(
+				// replication-factor update not allowed
```
sud82, line 1553: Add a test for this.

sud82 (approved):

PR #287, opened 2024-05-06
Fixing bugs

tanmayja:
1.  In case of unavailable partition recovery, scale up the cluster with batch and wait for sts after scale up.
2.  In case of security enabled flag:
     * if status.pods is nil, don't allow security update.
     * if isSecurityEnabled(status.aerospikeConfig) is true and spec is false, then don't allow to disable security.
     * remove IsSecurityEnabled flag, instead getting security enabled info by comparing hash.
3. In case of multiPodPerHost, deref the value and then compare
4. Allow EnableDynamicConfigUpdate to be true only if init image is updated to 2.2.0 or above.
5. Add IsClusterReadinessEnabled field in status to determine the cluster readiness status. This is used for PDB reconcile


controllers/reconciler.go
```diff
@@ -1042,3 +1046,30 @@ func (r *SingleClusterReconciler) handleEnableSecurity() error {
...
+		!asdbv1.GetBool(r.aeroCluster.Status.EnableDynamicConfigUpdate) {
+		// Populate Pod's annotation with aerospike config, if empty
+		pods, err := r.getClusterPodList()
+		if err != nil {
+			return err
+		}
+
+		for idx := range pods.Items {
+			pod := &pods.Items[idx]
+			if pod.Annotations == nil {
```
abhishekdwivedi3060, line 1066: We should check for only required annotation key in the annotations map not all

abhishekdwivedi3060 (changes requested):

controllers/rack.go
```diff
@@ -1731,6 +1743,75 @@ func (r *SingleClusterReconciler) getCurrentRackList() (
...
+	}
+
+	confMap, err := r.getConfigMap(rackState.Rack.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	requiredConfHash := confMap.Data[aerospikeConfHashFileName]
+
+	securityEnabledPods := make([]corev1.Pod, 0, len(r.aeroCluster.Status.Pods))
```
abhishekdwivedi3060, line 1800:
```suggestion
	securityEnabledPods := make([]corev1.Pod, 0, len(pods))
```

This should be len of rack pods as status contains pods of all racks

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2301,3 +2306,61 @@ func (c *AerospikeCluster) validateMaxUnavailable() error {
...
+			if val < 0 {
+				minVersion = version
+			}
+		} else {
+			return baseInitVersion, nil
+		}
+	}
+
+	return minVersion, nil
+}
```
abhishekdwivedi3060, line 2366: Instead of adding new var like minVersion and baseInitVersion, how about passing the version to be compared as a param and return true or false based on init tag comparison. Maybe move this code in `validateEnableDynamicConfigUpdate` func itself

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2282,3 +2306,61 @@ func (c *AerospikeCluster) validateMaxUnavailable() error {
...
+		return err
+	}
+
+	val, err := lib.CompareVersions(minInitVersion, minInitVersion4DynamicConf)
+	if err != nil {
+		return fmt.Errorf("failed to check image version: %v", err)
+	}
+
+	if val < 0 {
+		return fmt.Errorf("cannot enable enableDynamicConfigUpdate flag, init container are running version less than %s",
```
abhishekdwivedi3060, line 2330:
```suggestion
		return fmt.Errorf("cannot enable enableDynamicConfigUpdate flag, init containers are running version less than %s",
```
```suggestion
		return fmt.Errorf("cannot enable enableDynamicConfigUpdate flag, init container is running version less than %s",
```

api/v1/aerospikecluster_validating_webhook.go
tanmayja: Both implementations are fine, just wanted to give a generic function to return the minimum init version running in the cluster, which may be helpful in the future.

controllers/rack.go
```diff
@@ -1731,6 +1743,75 @@ func (r *SingleClusterReconciler) getCurrentRackList() (
...
+	}
+
+	confMap, err := r.getConfigMap(rackState.Rack.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	requiredConfHash := confMap.Data[aerospikeConfHashFileName]
+
+	securityEnabledPods := make([]corev1.Pod, 0, len(pods))
```
tanmayja, line 1800: Can we rename this variable, as this function looks generic whether config is updated or not. Func does not check specific about security.

api/v1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: Should we also say to do a cold restart to upgrade init version or point to aerospike doc ?

controllers/reconciler.go
```diff
@@ -446,6 +453,29 @@ func (r *SingleClusterReconciler) setStatusPhase(phase asdbv1.AerospikeClusterPh
 	return nil
 }
 
+func (r *SingleClusterReconciler) getClusterReadinessStatus() (bool, error) {
+	podList, err := r.getClusterPodList()
```
sud82, line 457: > Are we assuming that this function will be called only when podList is not empty? If list is empty then it will return true.

controllers/poddistruptionbudget.go
```diff
@@ -63,24 +63,18 @@ func (r *SingleClusterReconciler) deletePDB() error {
...
-	}
-
-	for podIdx := range podList.Items {
-		pod := &podList.Items[podIdx]
-
-		for containerIdx := range pod.Spec.Containers {
-			if pod.Spec.Containers[containerIdx].Name != asdbv1.AerospikeServerContainerName {
-				continue
-			}
+	// Check for cluster readiness status only when it's false.
```
sud82, line 66: Shouldn't we rely on flag update by updateStatus flow and just make a following check here?
```
if !r.aeroCluster.Status.IsClusterReadinessEnabled {
	r.Log.Info("Pod Readiness is not enabled throughout cluster. Skipping PodDisruptionBudget." +
		" Refer Aerospike documentation for more details.")
	return nil
}
We can rely on

sud82: In that case, PDB won't be updated in the same flow but it will get created in the next reconcile. I think reconcile runs one more time after updating the status.

sud82 (changes requested):

controllers/poddistruptionbudget.go
abhishekdwivedi3060: With that flag check, PBD creation was skipped as that flag is updated in the updateStatus and there is no reconcile after that. So changed that logic and optimised with false check

controllers/reconciler.go
abhishekdwivedi3060: This func is called after all the racks are reconciled and the current readiness check is determined using all the existing pods

PR #288, opened 2024-05-13
Changing 0 rack-id to correct rack id in rackConfig namespaces

tanmayja:


abhishekdwivedi3060 (approved):

controllers/configmap.go
```diff
@@ -106,6 +107,14 @@ func (r *SingleClusterReconciler) createConfigMapData(rack *asdbv1.Rack) (
 
 	confData[aerospikeTemplateConfFileName] = confTemp
 
+	// This field value is rectified in 3.3.0.
```
sud82, line 110: rack-id was historically set to 0 for all namespaces, but since the AKO 3.3.0, it reflects actual values. This change led to hash mismatches during the AKO 3.3.0 upgrade, triggering unnecessary warm restarts. Solution: Replace real rack-id with 0 in hash calculations to avoid this issue.

api/v1/aerospikecluster_mutating_webhook.go
```diff
@@ -411,6 +418,8 @@ func setDefaultNsConf(asLog logr.Logger, configSpec AerospikeConfigSpec,
 								err,
 							)
 						}
+					} else {
+						delete(nsMap, "rack-id")
```
sud82, line 422: Add an explanation of this delete.

api/v1/aerospikecluster_mutating_webhook.go
```diff
@@ -403,6 +403,13 @@ func setDefaultNsConf(asLog logr.Logger, configSpec AerospikeConfigSpec,
 					if rackID != nil {
 						// Add rack-id only in rack specific config, not in global config
 						defaultConfs := map[string]interface{}{"rack-id": *rackID}
+
+						// Delete rack-id from namespace in rack specific config if set to 0
```
sud82, line 407: Add a bit of detailed comment.

sud82:
This is merged in #287

PR #289, opened 2024-05-17
[KO-326] Added support for default-password-file

sud82:


controllers/client_policy.go
```diff
@@ -53,6 +53,55 @@ func (pp fromSecretPasswordProvider) Get(
...
+	defaultPasswordFilePath := asdbv1.GetDefaultPasswordFilePath(spec.AerospikeConfig)
+
+	// No default password file specified. Give default password.
+	if defaultPasswordFilePath == nil {
+		return asdbv1.DefaultAdminPassword
+	}
+
+	// Default password file specified. Get the secret name from the volume
+	volume := spec.Storage.GetVolumeForAerospikePath(*defaultPasswordFilePath)
+	secretName := volume.Source.Secret.SecretName
```
sud82, line 67: Validate if it's secret or not.

test/cluster_helper.go
```diff
@@ -1123,8 +1123,10 @@ func createDummyAerospikeCluster(
 						"proto-fd-max":     defaultProtofdmax,
 						"auto-pin":         "none",
 					},
-					"security": map[string]interface{}{},
-					"network":  getNetworkConfig(),
+					"security": map[string]interface{}{
```
sud82, line 1126: This func is used in pre-7.1.0 clusters as well. This field is not valid for them. Create another func.

test/access_control_test.go
```diff
@@ -2144,6 +2142,69 @@ var _ = Describe(
...
+					clientPolicy := getClientPolicy(aeroCluster, k8sClient)
+					clientPolicy.Password = pass
+
+					client, cerr := getClientWithPolicy(
+						pkgLog, aeroCluster, k8sClient, clientPolicy)
+					if cerr != nil {
+						return cerr
+					}
+
+					nodes := client.GetNodeNames()
```
sud82, line 2197: What if 3 min not enough. Can we use client.IsConnected?

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1813,10 +1816,15 @@ func validateRequiredFileStorageForFeatureConf(
 	// CA cert related fields are not supported with Secret Manager, so check their mount volume
 	allPaths = append(allPaths, caPaths...)
 
+	if defaultPassFilePath != nil {
```
abhishekdwivedi3060, line 1819: We can add a condition here to error out in case of secret agent default password file path

abhishekdwivedi3060 (changes requested):

tanmayja (commented):
change  AKO version from crd annotation in file api/v1/aerospikecluster_types.go:927

abhishekdwivedi3060 (approved):
Are we going to create a separate PR for 3.0.0 tag update across repo?

PR #309, opened 2024-08-07
KO-328: Helm charts for AerospikeBackupService, AerospikeBackup and AerospikeRestore CRs

abhishekdwivedi3060:


helm-charts/aerospike-restore/templates/aerospike-restore-cr.yaml
```diff
@@ -0,0 +1,25 @@
...
+  name: {{ template "aerospike-restore.commonName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "aerospike-restore.labels" . | nindent 4 }}
+    {{- with .Values.customLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  # Aerospike Backup Service reference
+  backupService:
```
tanmayja, line 13: can we use {{ .Values.backupService }} directly instead of breaking into fields, so that for new fields in the future, we don't have to change.

tanmayja (changes requested):

tanmayja (approved):

PR #324, opened 2024-11-12
KO-341: Added event flow for BackupService, Backup and Restore controllers

jwalantmodi05:


tanmayja (commented):
We should enhance event tracking by incorporating additional events, especially for the successful (happy path) scenarios. Currently, only failure and deletion events are being logged, leaving gaps in visibility for standard operations.

internal/controller/restore/reconciler.go
```diff
@@ -41,13 +49,21 @@ func (r *SingleRestoreReconciler) Reconcile() (result ctrl.Result, recErr error)
...
+
 			return res.Result, res.Err
 		}
 
 		return res.Result, nil
 	}
 
 	if err := r.checkRestoreStatus(); err != nil {
+		r.Log.Error(err, "Failed to check restore status")
+		r.Recorder.Eventf(r.aeroRestore, corev1.EventTypeWarning, "CheckStatusFailed",
```
abhishekdwivedi3060, line 64:
```suggestion
		r.Recorder.Eventf(r.aeroRestore, corev1.EventTypeWarning, "RestoreStatusCheckFailed",
```

internal/controller/restore/reconciler.go
```diff
@@ -131,6 +150,8 @@ func (r *SingleRestoreReconciler) checkRestoreStatus() error {
 	}
 
 	r.Log.Info(fmt.Sprintf("Restore status: %+v", restoreStatus))
+	r.Recorder.Eventf(r.aeroRestore, corev1.EventTypeNormal, "RestoreStatus",
+		"Restore status: %+v", restoreStatus)
```
abhishekdwivedi3060, line 154: Seems too much info for event. Better not to add in events.
Instead add Restore Compete event in the main reconcile func at the very end https://github.com/aerospike/aerospike-kubernetes-operator/pull/324/files#diff-47218f09e6b90f3189bfd8f3e13c64f77d5ac7bf48c9039ff21ee499ff944f33R73

abhishekdwivedi3060 (changes requested):

internal/controller/restore/reconciler.go
jwalantmodi05: Okay, I will remove this event and will add restore complete event.

abhishekdwivedi3060 (approved):

PR #330, opened 2024-12-20
Upgrade controller runtime package

JessicaGreben:
The cloud team would like to upgrade the version of the controller-runtime package to `sigs.k8s.io/controller-runtime v0.19.2` because we are encountering this error when we try to build our code with ako in this [PR](https://github.com/Aerospike-DBaaS/cluster-manager/pull/190):
```
# github.com/aerospike/aerospike-kubernetes-operator/api/v1
../../../../pkg/mod/github.com/aerospike/aerospike-kubernetes-operator@v0.0.0-20241016205704-12c90d4db3f3/api/v1/aerospikecluster_webhook.go:38:57: cannot use admission.NewDecoder(mgr.GetScheme()) (value of type admission.Decoder) as *admission.Decoder value in struct literal: admission.Decoder does not implement *admission.Decoder (type *admission.Decoder is pointer to interface, not interface)
```

We tried to downgrade `sigs.k8s.io/controller-runtime v0.17.3`  on our side, but it resulted in errors from other packages we are using (for example `github.com/external-secrets/external-secrets`) that also require a higher version.

Some of the breaking changes in `sigs.k8s.io/controller-runtime` going from `0.17.3` to `0.19.2` are:
1. `admission.Decoder` is now an interface instead of a struct (see [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2736))
2. The `options.AddFrom` method is deleted (ref: [docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.3/pkg/manager#Options.AndFrom)). This change is causing the CI for this PR to fail since the `AddFrom` method is used here: https://github.com/aerospike/aerospike-kubernetes-operator/blob/master/cmd/main.go#L79. It seems like a big undertaking to replace that. I'm not sure whats the best path forward.


JessicaGreben:
closing since an update is in this branch https://github.com/aerospike/aerospike-kubernetes-operator/compare/master...criticalVul

PR #334, opened 2025-01-09
AC-1707 Add support for optional annotations to headless and pod services

joshbench:
This PR will allow users to add custom annotations to the headless and pod services created by AKO.

The specific functionality we are looking for is to use [external-dns](https://github.com/kubernetes-sigs/external-dns) to create DNS records that map to these pod addresses, and we need to add annotations to these services to do so.

These annotations can be added by adding the following to the AerospikeCluster spec:
```
  seedsFinderServices:
    headless:
      annotations:
        foo: bar
    pod:
      annotations:
        foo: bar
```

And validated by checking the generated services:
- pod:
    ```
    ❯ kubectl describe service aerospike-0-0 -n aerospike
    Name:                     aerospike-0-0
    Namespace:                aerospike
    Labels:                   <none>
    Annotations:              foo: bar
    Selector:                 statefulset.kubernetes.io/pod-name=aerospike-0-0
    Type:                     NodePort
    ```
- headless:
    ```
    ❯ kubectl describe service aerospike -n aerospike
    Name:                     aerospike
    Namespace:                aerospike
    Labels:                   aerospike.com/cr=aerospike
                              app=aerospike-cluster
    Annotations:              foo: bar
                              service.alpha.kubernetes.io/tolerate-unready-endpoints: true
    Selector:                 aerospike.com/cr=aerospike,app=aerospike-cluster
    Type:                     ClusterIP
    ```

I have tested this by creating AerospikeClusters, both specifying these annotations and excluding them.


JessicaGreben:
I'm adding the feedback that @sud82 provided about this PR:

> Notes around configurable pod service and headless service annotations:
- The current place doesn't seem to be the right place to add service annotations as it's under SeedFinderServices.
- Instead of just annotations, it's better to add an ObjectMeta structure to cover both labels and annotations.
- There is a separate section in LoadBalancerSpec for LB annotations which means all types of services should ideally have their own configurable objectMeta option. Using common PodSpec annotations for services will be confusing
- PR doesn't handle annotations update flow.
- No test-case added

PR #335, opened 2025-01-10
KO-370: Added custom Service Account support in the backup service

abhishekdwivedi3060:


api/v1beta1/aerospikebackupservice_webhook.go
```diff
@@ -52,6 +53,10 @@ func (r *AerospikeBackupService) Default() {
 	if r.Spec.Resources != nil && r.Spec.PodSpec.ServiceContainerSpec.Resources == nil {
 		r.Spec.PodSpec.ServiceContainerSpec.Resources = r.Spec.Resources
 	}
+
+	if r.Spec.PodSpec.ServiceAccountName == "" {
```
sud82, line 57: Why not make it like GetAerospikeInitContainerImage func

sud82 (changes requested):

api/v1beta1/aerospikebackupservice_webhook.go
abhishekdwivedi3060: Moved to func call in the reconciler for now.

sud82 (approved):

PR #337, opened 2025-01-13
KO-372: Drop support for old server versions and K8s APIs

abhishekdwivedi3060:
- Dropped support for v1beta1 AerospikeCluster API
- Dropped support for server version < 6.0


tanmayja (commented):
we can remove this as well
https://github.com/aerospike/aerospike-kubernetes-operator/blob/f124d00ef1d1b64a5873f51936d7473a623f0e98/internal/controller/cluster/pod.go#L120

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1281,82 +1251,41 @@ func getNamespaceReplicationFactor(nsConf map[string]interface{}) (int, error) {
...
+	return validateSecurityContext(newSpec, oldSpec)
 }
 
-func validateSecurityContext(
-	newVersion, oldVersion string, newSpec, oldSpec *AerospikeConfigSpec) error {
-	ovflag, err := IsSecurityEnabled(oldVersion, oldSpec)
+func validateSecurityContext(newSpec, oldSpec *AerospikeConfigSpec) error {
+	ovflag, err := IsSecurityEnabled(oldSpec)
 	if err != nil {
 		if !errors.Is(err, internalerrors.ErrNotFound) {
```
tanmayja, line 1277: `IsSecurityEnabled` will never return `ErrNotFound` type error, this is being handled in the `IsSecurityEnabled` defination

tanmayja (commented):
Is there an explicit validation in the webhook to ensure the server version is less than 6?

It seems schema validation would catch this, but the error message might not clearly indicate that the failure is due to the minimum version requirement.

abhishekdwivedi3060:
> Is there an explicit validation in the webhook to ensure the server version is less than 6?
> 
> It seems schema validation would catch this, but the error message might not clearly indicate that the failure is due to the minimum version requirement.

There is a explicit min server version check in webhook validation, something like this:
```
admission webhook "vaerospikecluster.kb.io" denied the request: image version 5.7.0.1 not supported. Base version 6.0.0.0
```

sud82 (approved):

PR #339, opened 2025-01-20
KO-371: Added a validation check to ensure that MRT fields are allowed only in SC namespaces

jwalantmodi05:


api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1235,6 +1239,24 @@ func validateNamespaceConfig(
...
+	}
+
+	return nil
+}
+
+func isMRTFieldSet(nsConf map[string]interface{}) bool {
+	_, isMRTDurationSet := nsConf["mrt-duration"]
+	disableMRTWrites, ok := nsConf["disable-mrt-writes"]
+
+	return (ok && disableMRTWrites.(bool)) || isMRTDurationSet
```
tanmayja, line 1257: In AP cluster if user sets `disable-mrt-writes` to false, validation will pass but reconcile will fail, as set-config command will fail.

api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1235,6 +1239,24 @@ func validateNamespaceConfig(
...
+	return nil
+}
+
+func isMRTFieldSet(nsConf map[string]interface{}) bool {
+	_, isMRTDurationSet := nsConf["mrt-duration"]
+	_, isDisableMRTWritesSet := nsConf["disable-mrt-writes"]
+
+	return isDisableMRTWritesSet || isMRTDurationSet
+}
+
```
abhishekdwivedi3060, line 1259: nit:
We can make it generic by making a list of MRT related fields and looping over them

abhishekdwivedi3060 (approved):

PR #345, opened 2025-02-18
[KO-384] Adding blkdiscardWithHeaderCleanup option as cleanup methods for Aerospike devices

tanmayja:


sud82 (changes requested):

api/v1/storage.go
```diff
@@ -440,45 +441,58 @@ func validateStorageVolumeSource(volume *VolumeSpec) error {
 	}
```
abhishekdwivedi3060, line 441: Can we optimise this whole func? Possible optimisiation
1. Too many if conditions. Why not user a source counter to count the number of sources present and return error once based on counter 0 or > 1 value.
2. Define a set of  `source.PersistentVolume.AccessModes` and use that for lookups in a loop

abhishekdwivedi3060 (changes requested):

abhishekdwivedi3060 (approved):

sud82 (approved):

PR #347, opened 2025-02-21
 fix: [KO-392] fix init image version validation during cluster create.

tanmayja:


api/v1/aerospikecluster_validating_webhook.go
```diff
@@ -222,6 +222,25 @@ func (c *AerospikeCluster) validate(aslog logr.Logger) (admission.Warnings, erro
...
+			return warnings, fmt.Errorf("failed to check init image version: %v", err)
+		}
+
+		if val < 0 {
+			return warnings, fmt.Errorf("cannot set enableDynamicConfigUpdate flag, init container version is less"+
+				" than %s. Please visit https://aerospike.com/docs/cloud/kubernetes/operator/Cluster-configuration-settings#spec"+
+				" for more details about enableDynamicConfigUpdate flag",
+				minInitVersionForDynamicConf)
+		}
+	}
```
abhishekdwivedi3060, line 242: See if we can move this common code to a func and reuse that both in create and update

abhishekdwivedi3060 (changes requested):

api/v1/aerospikecluster_validating_webhook.go
tanmayja: Both ways are different, while create, we check init version in spec and while update we check status and get the minimum version running. And apart from this error message is also different in both cases. I don't think so we will be saving much on code duplication here,

sud82 (approved):

abhishekdwivedi3060 (approved):

PR #349, opened 2025-02-24
KO-380: Upgrade aerospike-client-go to v8

jwalantmodi05:


.github/workflows/golangci-lint.yaml
```diff
@@ -17,13 +17,13 @@ jobs:
...
+          go-version: 1.23
       - name: Checkout sources
         uses: actions/checkout@v3
         with:
           submodules: true
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.59.1
+          version: v1.60.1
```
tanmayja, line 28: let the lint version be `v1.60.2`, the same as management-lib.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 )
 
 require (
+	github.com/aerospike/aerospike-client-go/v7 v7.8.0 // indirect
```
tanmayja, line 31: Why is version v7 still appearing as an indirect dependency?

jwalantmodi05: aerospike-backup-service/v3 is depends on it.

sud82:
Let's reopen it when the backup service also starts using v8

tanmayja (approved):

abhishekdwivedi3060 (approved):

PR #358, opened 2025-03-24
WIP: refactor: move aerospike config validation code to management-lib [KO-407]

abhishekdwivedi3060:
- Moved aerospikeConfig related validation code to management-lib
- TODO: Finalise on the aerospike config update code
- Deferred int overflow gosec check as it requires changes at multiple places


abhishekdwivedi3060:
After discussion it was decided to keep the validation code in AKO only for now

PR #360, opened 2025-03-27
refactor: move aerospikeconfig valdidation code to validation pkg [KO-407]

abhishekdwivedi3060:


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -714,206 +706,41 @@ func validateClusterSize(_ logr.Logger, sz int) error {
...
-
-	if nsList, ok1 := nsListInterface.([]interface{}); !ok1 {
-		return fmt.Errorf(
-			"aerospikeConfig.namespace not valid namespace list %v",
-			nsListInterface,
-		)
-	} else if err := validateNamespaceConfig(
-		nsList, storage, clSize,
+	// It validates the aerospikeConfig schema and generic aerospikeConfig fields
+	if err := validation.ValidateAerospikeConfig(
```
tanmayja, line 714: When we started passing a value instead of a configSpec pointer, checking for nil on the map might incorrectly allow an empty map. Additionally, if we type-cast a value from the map without first verifying that the corresponding key exists, this could potentially lead to a runtime failure. Have we considered whether any such scenarios could occur?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -733,206 +725,41 @@ func validateClusterSize(_ logr.Logger, sz int) error {
...
-					)
-				}
-			}
-		}
-	}
-
-	if err := validateTLSClientNames(
-		serviceConf.(map[string]interface{}), clientCert,
-	); err != nil {
+	if err := validateNetworkConfig(configSpec.Value, clientCert); err != nil {
```
tanmayja, line 739: can we rename functions here, like there are 2 functions `validateNetworkConfig` one in aerospikecluster_validating_webhook.go and another in validate.go, check others also.

abhishekdwivedi3060: Actually I intentionality kept the names same to identify the similar functionality across diff pkg.
Eg: `validation.ValidateAerospikeConfig` is validation func for aerospikeConfig specific validation whereas `validateAerospikeConfig` in webhook covers both aerospikeConfig and AKO specific aerospikeConfig validation.
This way it's easier to track. 

Think of it like `validation` is a third party pkg and we are writing some wrapper funcs to add some AKO specific validations.

We can discuss and decide the renaming part if its still confusing.

tanmayja (approved):

PR #368, opened 2025-04-24
fix: [KO-415] Roster is not getting set for ignored pod after recovery

jwalantmodi05:


internal/controller/cluster/reconciler.go
```diff
@@ -274,9 +274,9 @@ func (r *SingleClusterReconciler) Reconcile() (result ctrl.Result, recErr error)
 		return reconcile.Result{}, err
 	}
 
-	// Try to recover pods only when MaxIgnorablePods is set
-	if r.aeroCluster.Spec.RackConfig.MaxIgnorablePods != nil {
-		if res := r.recoverIgnorablePods(); !res.IsSuccess {
+	// Try to recover pods only when length of ignorablePodNames is greater than zero
```
tanmayja, line 277: try to add some "why?" in comments. Current comment is exactly saying what like 278 is doing.

internal/controller/cluster/reconciler.go
```diff
@@ -286,36 +286,39 @@ func (r *SingleClusterReconciler) Reconcile() (result ctrl.Result, recErr error)
...
 			}
 
-			r.Log.Info("Deleted pod", "pod", podList.Items[idx].Name)
+			r.Log.Info("Deleted pod", "pod", pod.Name)
 		}
 	}
 
-	if anyPodFailed {
-		r.Log.Info("Found failed/pending pod(s), requeuing")
+	if anyPodFailed || len(ignorablePodNames) > 0 {
```
tanmayja, line 320: we are calling this function only when ignorablePodNames > 0.
what is the point of having this condition here?

internal/controller/cluster/reconciler.go
```diff
@@ -286,36 +286,39 @@ func (r *SingleClusterReconciler) Reconcile() (result ctrl.Result, recErr error)
...
+
 	var anyPodFailed bool
 	// Try to recover failed/pending pods by deleting them
-	for idx := range podList.Items {
-		if cErr := utils.CheckPodFailed(&podList.Items[idx]); cErr != nil {
+	for idx := range ignorablePodNamesList {
+		pod := &corev1.Pod{}
+		podNamespacedName := types.NamespacedName{Namespace: r.aeroCluster.Namespace, Name: ignorablePodNamesList[idx]}
+
+		if err := r.Client.Get(context.TODO(), podNamespacedName, pod); err != nil {
```
tanmayja, line 300: can we use the older approach of getting the list of all the pods and filter it in code based on ignorablePodNamesList. 
It can reduce the API calls if possible.

tanmayja (changes requested):

internal/controller/cluster/reconciler.go
jwalantmodi05: Will remove it.

abhishekdwivedi3060 (approved):

internal/controller/cluster/reconciler.go
```diff
@@ -319,7 +321,9 @@ func (r *SingleClusterReconciler) recoverIgnorablePods() common.ReconcileResult
 		return common.ReconcileRequeueAfter(0)
 	}
 
-	return common.ReconcileSuccess()
+	r.Log.Info("Found ignorable pod(s), requeuing")
+
+	return common.ReconcileRequeueAfter(0)
```
tanmayja, line 326: The return statement is the same in both lines 321 and 326, so it doesn't need to be duplicated.
If the goal is to differentiate log messages based on conditions, you can handle that separately before a single return.
Or you can place the log message as soon as you get the failed pod under `CheckPodFailed` condition.

tanmayja (approved):

PR #370, opened 2025-05-02
feat: add aerospike_ako_aerospikecluster_phase custom metric [KO-408]

abhishekdwivedi3060:
- Add aerospike_ako_aerospikecluster_phase custom metric  to track the status of all the CRs.
- Sample screenshot of Prometheus showing the metric
Screenshot 2025-05-21 at 6 24 28 PM

### DIFF
--- a/internal/controller/backup-service/reconciler.go
+++ b/internal/controller/backup-service/reconciler.go
@@ -102,11 +102,11 @@ func (r *SingleBackupServiceReconciler) Reconcile() (result ctrl.Result, recErr 
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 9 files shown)


abhishekdwivedi3060:
> the

Metrics at the AKO level will always be accurate irrespective of how quickly the metrics value is switching.
What you'll see in the Prom or Grafana will depend on the scrapping interval, normally its 15s. So yes, transient states like error and all won't be captured within that interval. However, if the error is not recovered within that time, you'll see that in the scrapped config.

tanmayja (dismissed):

tanmayja (approved):

PR #373, opened 2025-05-12
fix: ABS controller recover reconcile flow [KO-426]

abhishekdwivedi3060:
- Fixed ABS controller reconcile flow to recover from pending state in case of new ABS deployment
- Added test-cases

### DIFF
--- a/internal/controller/backup-service/reconciler.go
+++ b/internal/controller/backup-service/reconciler.go
@@ -102,11 +102,11 @@ func (r *SingleBackupServiceReconciler) Reconcile() (result ctrl.Result, recErr 
 		return ctrl.Result{}, err
 	}
 
-	if err := r.reconcileDeployment(); err != nil {
-		r.Log.Error(err, "Failed to reconcile deployment",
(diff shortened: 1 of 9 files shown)


internal/controller/backup-service/reconciler.go
```diff
@@ -348,18 +348,30 @@ func (r *SingleBackupServiceReconciler) getBackupSvcDeployment() (*app.Deploymen
 }
 
 func (r *SingleBackupServiceReconciler) updateBackupSvcConfig() error {
+	if r.aeroBackupService.Status.Config.Raw == nil {
+		r.Log.Info("Skipping backup service config reload as status is empty")
```
tanmayja, line 352: add backup service name and namespace in log message.

tanmayja (approved):

internal/controller/backup-service/reconciler.go
abhishekdwivedi3060: That info is already added in the logger while initialising the reconciler

PR #375, opened 2025-05-19
feat: adding support for dynamic rackid assignment after pod schedule [KO-385]

tanmayja:


test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,420 @@
...
+												},
+											},
+										},
+									},
+								},
+							},
+						}
+
+						err := k8sClient.Create(ctx, ds)
+						Expect(err).NotTo(HaveOccurred())
```
abhishekdwivedi3060, line 107: Shall we wait for the pods to come up to avoid timing related race-conditions?

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,420 @@
...
+									}
+								}
+							},
+						)
+					},
+				)
+			},
+		)
+		Context(
+			"When doing invalid operations", func() {
```
abhishekdwivedi3060, line 292: Can we check the error string here just to confirm the error reasons for different invalid cases?

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,420 @@
...
+						aeroCluster.Spec.RackConfig.RackIDSource = &asdbv1.RackIDSource{FilePath: aerospikePath + "/rackid"}
+
+						Expect(DeployCluster(k8sClient, ctx, aeroCluster)).Should(HaveOccurred())
+					},
+				)
+
+				It(
+					"Should fail if dynamic rackID source volume is not hostpath type", func() {
+						aeroCluster := createDummyRackAwareAerospikeCluster(clusterNamespacedName, 2)
+						emptydirVolume := asdbv1.VolumeSpec{
```
abhishekdwivedi3060, line 360:
```suggestion
						emptyDirVolume := asdbv1.VolumeSpec{
```

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,420 @@
...
+
+						aeroCluster.Spec.Storage.Volumes = append(aeroCluster.Spec.Storage.Volumes, emptydirVolume)
+						aeroCluster.Spec.RackConfig.RackIDSource = &asdbv1.RackIDSource{FilePath: "/opt/empty/rackid"}
+
+						Expect(DeployCluster(k8sClient, ctx, aeroCluster)).Should(HaveOccurred())
+					},
+				)
+
+				It(
+					"Should fail if dynamic rackID source volume is mounted  with readOnly false", func() {
```
abhishekdwivedi3060, line 378:
```suggestion
					"Should fail if dynamic rackID source volume is mounted with readOnly false", func() {
```

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,420 @@
...
+								info, err := requestInfoFromNode(logger, k8sClient, ctx, clusterNamespacedName, "namespace/test", &pod)
+								Expect(err).ToNot(HaveOccurred())
+
+								confs := strings.Split(info["namespace/test"], ";")
+								for _, conf := range confs {
+									if strings.Contains(conf, "rack-id") {
+										keyValue := strings.Split(conf, "=")
+										Expect(keyValue[1]).To(Equal("2"))
+									}
+								}
```
abhishekdwivedi3060, line 222: Can we move this out in a common func something like `validateDynamicRackID`?

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,420 @@
...
+								Template: v1.PodTemplateSpec{
+									ObjectMeta: metav1.ObjectMeta{
+										Labels: map[string]string{"app": "hostpath-writer"}, // must match selector
+									},
+									Spec: v1.PodSpec{
+										Containers: []v1.Container{
+											{
+												Name:    "writer",
+												Image:   "busybox",
+												Command: []string{"sh", "-c", `echo "2" > /tmp/rackid && sleep 3600`},
```
abhishekdwivedi3060, line 72: Can we test with multiple dynamic rackid like 1, 2 ?

abhishekdwivedi3060 (changes requested):

api/v1/aerospikecluster_types.go
```diff
@@ -1100,6 +1111,14 @@ type AerospikeNetworkPolicy struct {
 	CustomTLSFabricNetworkNames []string `json:"customTLSFabricNetworkNames,omitempty"`
 }
 
+// RackIDSource specifies the source from which to read the rack ID.
+type RackIDSource struct {
+	// FilePath specifies the absolute path to a file containing the rack ID mounted in aerospike server.
+	// The file should contain a single integer value.
+	// +optional
+	FilePath string `json:"filePath,omitempty"`
```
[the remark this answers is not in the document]
tanmayja: yeah, added annotation also as an option that's why it was optional. later removed it after discussion.
Will change it to required.

internal/controller/cluster/pod.go
```diff
@@ -137,6 +137,10 @@ func (r *SingleClusterReconciler) getRollingRestartTypeMap(rackState *RackState,
 		if podStatus.DynamicConfigUpdateStatus == asdbv1.PartiallyFailed {
 			restartTypeMap[pods[idx].Name] = mergeRestartType(restartTypeMap[pods[idx].Name], quickRestart)
 		}
+
+		if !reflect.DeepEqual(podStatus.RackIDSource, r.aeroCluster.Spec.RackConfig.RackIDSource) {
+			restartTypeMap[pods[idx].Name] = mergeRestartType(restartTypeMap[pods[idx].Name], quickRestart)
+		}
```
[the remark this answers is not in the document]
tanmayja: dynamicRackIDEnabled flag option won't work, because it won't say if anything else is changed like path or source option in future.
Hash seems to be a better option but it will be different than existing hashes like podspec or networkpolicy. Because rackIDSource is an optional field unlike others, so need special handling at multiple places. Anyway will give it a try.

internal/controller/cluster/statefulset.go
```diff
@@ -1453,10 +1455,19 @@ func getFinalVolumeAttachmentsForVolume(volume *asdbv1.VolumeSpec) (
...
+		readOnlyVolume = true
+	}
+
 	initContainerAttachments = append(
 		initContainerAttachments, asdbv1.VolumeAttachment{
 			ContainerName: asdbv1.AerospikeInitContainerName,
 			Path:          initVolumePath,
+			AttachmentOptions: asdbv1.AttachmentOptions{
+				MountOptions: asdbv1.MountOptions{ReadOnly: ptr.To(readOnlyVolume)},
+			},
```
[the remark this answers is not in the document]
tanmayja: we can use, is user wants to mount any volume as read-only in server we should mount it in init in same way.

internal/controller/cluster/statefulset.go
```diff
@@ -1490,11 +1502,16 @@ func addVolumeMountInContainer(
...
 				} else {
 					volumeMount = corev1.VolumeMount{
-						Name:      volumeName,
-						MountPath: volumeAttachment.Path,
+						Name:             volumeName,
+						MountPath:        volumeAttachment.Path,
+						ReadOnly:         ptr.Deref(volumeAttachment.AttachmentOptions.MountOptions.ReadOnly, false),
+						SubPath:          volumeAttachment.AttachmentOptions.MountOptions.SubPath,
+						SubPathExpr:      volumeAttachment.AttachmentOptions.MountOptions.SubPathExpr,
+						MountPropagation: volumeAttachment.AttachmentOptions.MountOptions.MountPropagation,
```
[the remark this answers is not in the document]
tanmayja: yes, but if user has set them but not getting honored, we would have got a bug already

test/cluster/dynamic_rackid_test.go
tanmayja: There are many places where we should check err message instead of just err occurred to make sure that it is failing at expected stage.
Will add in these tests as of now.

internal/controller/cluster/rack.go
```diff
@@ -1507,7 +1498,7 @@ func (r *SingleClusterReconciler) isVolumeAttachmentAddedOrUpdated(
 		if volumeMount != nil {
 			// Found, check for updated
 			if getOriginalPath(volumeMount.MountPath) != attachment.Path ||
-				volumeMount.ReadOnly != attachment.ReadOnly ||
+				volumeMount.ReadOnly != ptr.Deref(attachment.ReadOnly, false) ||
```
abhishekdwivedi3060, line 1501: We use a wrapper of this func called `GetBool` to avoid passing `false` at multiple places.  Check at other places as well

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,469 @@
...
+										err = k8sClient.Get(
+											ctx, types.NamespacedName{
+												Name:      aeroCluster.Name + "-1-0",
+												Namespace: clusterNamespacedName.Namespace,
+											}, podObject,
+										)
+
+										return isHostPathReadOnly(podObject.Status.ContainerStatuses, aerospikePath)
+									}, time.Minute, time.Second,
+								).Should(BeTrue())
```
abhishekdwivedi3060, line 161: Do we need this check hostPathReadOnly check as we are enforcing this in the webhook.

abhishekdwivedi3060 (changes requested):

PR #377, opened 2025-05-22
fix: Fix integer overflow conversion issues [KO-428]

jwalantmodi05:


test/cluster/autoscaling_test.go
```diff
@@ -85,11 +85,11 @@ func validateScaleSubresourceOperation(currentSize, desiredSize int, clusterName
...
-		func() int32 {
+		func() int {
 			aeroCluster, err := getCluster(k8sClient, goctx.TODO(), clusterNamespacedName)
 			Expect(err).ToNot(HaveOccurred())
 
-			return aeroCluster.Spec.Size
+			return int(aeroCluster.Spec.Size)
 		}, time.Minute, time.Second,
-	).Should(Equal(int32(desiredSize)))
+	).Should(Equal(desiredSize))
```
abhishekdwivedi3060, line 94: Any reason for doing the opposite? int32 -> int conversion.
Why not change `desiredSize` param to int32 to keep size consistent across

test/cluster/poddisruptionbudget_test.go
```diff
@@ -215,14 +215,14 @@ func getPDB(ctx context.Context, aerocluster *asdbv1.AerospikeCluster) (*policyv
 	return pdb, err
 }
 
-func createPDB(ctx context.Context, aerocluster *asdbv1.AerospikeCluster, maxUnavailable int) error {
+func createPDB(ctx context.Context, aerocluster *asdbv1.AerospikeCluster, maxUnavailable int32) error {
```
abhishekdwivedi3060, line 218: If we changing the int type here, we should do the same in `validatePDB` func

internal/webhook/v1/aerospikecluster_mutating_webhook.go
```diff
@@ -714,8 +714,8 @@ func setDefaultsInConfigMap(
 ) error {
 	for k, v := range defaultConfigs {
 		// Special handling.
-		// Older baseValues are parsed to int64 but defaults are in int
-		if newV, ok := v.(int); ok {
+		// Older baseValues are parsed to int64 but defaults are in int32
+		if newV, ok := v.(int32); ok {
```
abhishekdwivedi3060, line 718: Check this change once against rack-id set flow in mutation webhook.
Mutations sets rack-id as int not int32

abhishekdwivedi3060 (changes requested):

test/cluster/poddisruptionbudget_test.go
jwalantmodi05: If we change the int type in validatePDB then we won’t be able to use maxUnavailable variable with string value. Because maxUnavailable.IntValue() returns int type.

test/cluster/autoscaling_test.go
jwalantmodi05: To minimise code changes.

abhishekdwivedi3060 (dismissed):

tanmayja (dismissed):

PR #378, opened 2025-06-02
docs: Add AKS storageClass sample and updated others to csi based provisioner [KO-432]

abhishekdwivedi3060:


tanmayja (approved):

test/deploy-test-operator.sh
```diff
@@ -11,11 +11,11 @@ CATALOG_IMG=$2
 
 # Create storage classes.
 case $(kubectl get nodes -o yaml) in
-  *"attachable-volumes-gce-pd"*)
```
abhishekdwivedi3060, line 14: This condition does't work, so skipped completely in current flow.
Updated the condition

tanmayja (approved):

PR #379, opened 2025-06-02
fix: pod restart if in-memory data-size changes [KO-424]

tanmayja:


internal/controller/cluster/pod.go
abhishekdwivedi3060, line 1: What about test-case for this data-size change?

abhishekdwivedi3060 (changes requested):

test/cluster/dynamic_config_test.go
```diff
@@ -251,8 +256,41 @@ var _ = Describe(
...
 
-						validateServerRestart(ctx, aeroCluster, podPIDMap, true)
+						err = updateCluster(k8sClient, ctx, aeroCluster)
+						Expect(err).ToNot(HaveOccurred())
+
+						By("Fetch and verify static configs")
+
+						pod = aeroCluster.Status.Pods[aeroCluster.Name+"-0-0"]
+
+						info, err := requestInfoFromNode(logger, k8sClient, ctx, clusterNamespacedName, "namespace/test", &pod)
```
abhishekdwivedi3060:
```suggestion
						info, err := requestInfoFromNode(logger, k8sClient, ctx, clusterNamespacedName, "namespace/mem", &pod)
```

abhishekdwivedi3060 (changes requested):

abhishekdwivedi3060 (approved):

PR #383, opened 2025-06-13
feat: Optimise upgrade and rolling restart with dynamic migrate-fill-delay  [KO-427]

abhishekdwivedi3060:


internal/controller/cluster/pod.go
```diff
@@ -268,9 +268,37 @@ func (r *SingleClusterReconciler) rollingRestartPods(
...
 			return res
 		}
+
+		// Set migrate-fill-delay O to immediately start the migration. Will be reverted back to the original value
+		// in the next reconcile.
+		if setMigrateFillDelay {
+			if res := r.setMigrateFillDelay(clientPolicy, &rackState.Rack.AerospikeConfig, true,
+				ignorablePodNames,
+			); !res.IsSuccess {
+				r.Log.Error(res.Err, "Failed to revert migrate-fill-delay after restarting active pods")
```
sud82:
```suggestion
				r.Log.Error(res.Err, "Failed to set migrate-fill-delay to `0` after restarting active pods")
```

internal/controller/cluster/pod.go
```diff
@@ -268,9 +268,37 @@ func (r *SingleClusterReconciler) rollingRestartPods(
...
+		r.Log.Info("Migrate-fill-delay", "needed", setMigrateFillDelay)
+
+		// Revert migrate-fill-delay to the original value before restarting active pods.
+		// This will be a no-op in the first reconcile
+		if setMigrateFillDelay {
+			// Add an optimisation check to only run set MFD in case of cold restart.
+			if res := r.setMigrateFillDelay(clientPolicy, &rackState.Rack.AerospikeConfig, false,
+				ignorablePodNames,
+			); !res.IsSuccess {
+				r.Log.Error(res.Err, "Failed to set migrate-fill-delay before restarting active pods")
```
sud82:
```suggestion
				r.Log.Error(res.Err, "Failed to set migrate-fill-delay to original value before restarting the running pods")
```

sud82: check for upgrade as well.

internal/controller/cluster/pod.go
```diff
@@ -268,9 +268,37 @@ func (r *SingleClusterReconciler) rollingRestartPods(
 			return res
 		}
 
+		clientPolicy := r.getClientPolicy()
+		setMigrateFillDelay := r.shouldSetMigrateFillDelay(rackState, podsToRestart, restartTypeMap)
+
+		r.Log.Info("Migrate-fill-delay", "needed", setMigrateFillDelay)
```
sud82: ... this log doesn't look good. Let's rethink.

sud82 (changes requested):

internal/controller/cluster/pod.go
abhishekdwivedi3060: Added that log for debugging. Replaced it with complete log msg

internal/controller/cluster/pod.go
```diff
@@ -268,9 +268,40 @@ func (r *SingleClusterReconciler) rollingRestartPods(
 			return res
 		}
 
+		clientPolicy := r.getClientPolicy()
+		setMigrateFillDelay := r.shouldSetMigrateFillDelay(rackState, podsToRestart, restartTypeMap)
+
+		r.Log.Info(
```
tanmayja, line 274: do we need this log or can it go in if condition?
same in upgrade

abhishekdwivedi3060: It's in general flow to denote MFD optimisation. True or false denotes whether it's effective or not

tanmayja: MFD optimisation is not very generic flow overall. So no need to print about it in every rolling restart flow.

abhishekdwivedi3060: Leaving it as it is as per our discussion.

PR #386, opened 2025-06-19
feat: adding option ddHeaderOnly as volume initialize method [KO-384]

tanmayja:


api/v1/aerospikecluster_types.go
```diff
@@ -623,6 +623,9 @@ const (
 	// AerospikeVolumeMethodDD specifies the block volume should be zeroed using dd command.
 	AerospikeVolumeMethodDD AerospikeVolumeMethod = "dd"
 
+	// AerospikeVolumeMethodDDHeaderOnly specifies the block volume's header should be zeroed using dd command.
```
abhishekdwivedi3060: - Maybe also mention the 8MiB header size
- Also mention a point/warning around old Aerospike devices usage

api/v1/aerospikecluster_types.go
```diff
@@ -623,6 +623,9 @@ const (
...
+	AerospikeVolumeMethodDDHeaderOnly AerospikeVolumeMethod = "ddHeaderOnly"
```
abhishekdwivedi3060: How about just `ddHeader`?

tanmayja: changed to `headerCleanup`

abhishekdwivedi3060 (dismissed):

PR #388, opened 2025-06-19
Update values.yaml

realmgic:
fix the wrong exporter tag (no `v`)

### DIFF
--- a/helm-charts/aerospike-cluster/values.yaml
+++ b/helm-charts/aerospike-cluster/values.yaml
@@ -66,12 +66,14 @@ aerospikeNetworkPolicy: {}
 podSpec: {}
 ##  Multi pod per host
 #   multiPodPerHost: true
-#    sidecars:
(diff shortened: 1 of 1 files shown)


helm-charts/aerospike-cluster/values.yaml
```diff
@@ -68,7 +68,7 @@ podSpec: {}
 #   multiPodPerHost: true
 #    sidecars:
```
abhishekdwivedi3060: Not related to your tag changes but since we are fixing this side-car section, can you please also fix the indentation of `sidecars`. Should be at the same level as multiPodPerHost

```
   multiPodPerHost: true
   sidecars:
     - name: aerospike-prometheus-exporter
       image: aerospike/aerospike-prometheus-exporter:v1.18.0
       ports:
         - containerPort: 9145
           name: exporter
```

PR #389, opened 2025-06-26
Disable security feat

tanmayja:


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -159,6 +163,22 @@ func (acv *AerospikeClusterCustomValidator) ValidateUpdate(_ context.Context, ol
...
+	oldSpec *asdbv1.AerospikeClusterSpec,
+	newSpec *asdbv1.AerospikeClusterSpec,
+	status *asdbv1.AerospikeClusterStatus,
+) error {
+	// Prevent removal of access control before status is updated
+	if oldSpec.AerospikeAccessControl != nil &&
+		newSpec.AerospikeAccessControl == nil &&
+		status.AerospikeAccessControl == nil {
+		return fmt.Errorf(
+			"cannot remove AerospikeAccessControl: status has not yet been updated with the current configuration")
```
abhishekdwivedi3060: Can we change it to something more meaningful like
```suggestion
			"cannot remove aerospikeAccessControl: security enablement is in progress and status has not yet been updated to reflect the current configuration")
```

internal/controller/cluster/rack.go
```diff
@@ -1807,41 +1806,40 @@ func (r *SingleClusterReconciler) handleEnableSecurity(rackState *RackState, ign
...
-	for idx := range pods {
-		podName := pods[idx].Name
-		podStatus := r.aeroCluster.Status.Pods[podName]
+	for idx := range podList.Items {
+		_, port := r.getResolvedTLSNameAndPort()
 
-		if podStatus.AerospikeConfigHash == requiredConfHash {
-			// Config hash is matching, it means config has been applied
-			updatedPods = append(updatedPods, *pods[idx])
+		if security, ok := securityConfigs[hostID(podList.Items[idx].Status.PodIP, int(*port))]; ok {
```
abhishekdwivedi3060, line 1835: Either we can
1. Move the contents of func`getClusterSecurityConfig` here only (not much useful) and use the same hostConns.ID for lookup here. No need to create a new func and hostID
OR
2. Move this `r.getResolvedTLSNameAndPort()` outside the loop. It'll same for all

internal/controller/cluster/service.go
```diff
@@ -496,6 +496,16 @@ func (r *SingleClusterReconciler) getServiceTLSNameAndPortIfConfigured() (tlsNam
 	return tlsName, port
 }
 
+func (r *SingleClusterReconciler) getResolvedTLSNameAndPort() (tlsName string, port *int32) {
```
abhishekdwivedi3060, line 522: Func name and it's functionality is kind of confusing.
This is not actually returning the TLS and port, instead falling back to non-tls port for client connectivity.
See if we actually need this func. If review comments are incorporated then this might be called at 1 place only

test/cluster/utils.go
```diff
@@ -248,6 +248,8 @@ func (acs *AerospikeConfSpec) setEnableSecurity(enableSecurity bool) {
 	if enableSecurity {
```
abhishekdwivedi3060, line 248: nit: It's better to set the name as `setSecurity` as we are driving set/reset based on param

test/cluster/access_control_test.go
```diff
@@ -1600,6 +1601,199 @@ var _ = Describe(
...
+								)
+								err = testAccessControlReconcile(
+									aeroCluster, ctx,
+								)
+								Expect(err).To(HaveOccurred())
+								Expect(err.Error()).Should(ContainSubstring("SECURITY_NOT_ENABLED"))
+							},
+						)
+
+						FIt(
```
abhishekdwivedi3060: Revert

test/cluster/access_control_test.go
```diff
@@ -1600,6 +1601,199 @@ var _ = Describe(
...
+									aerospikeConfigSpec,
+								)
+								err = testAccessControlReconcile(
+									aeroCluster, ctx,
+								)
+								if err != nil {
+									Fail("Security should be enabled")
+								}
+
+								aerospikeConfigSpec, err = NewAerospikeConfSpec(latestImage)
```
abhishekdwivedi3060: Can we not use the same `aerospikeConfigSpec` var above and manipulate security on it?
We are not changing anything apart from security. See at other places also if applicable

PR #391, opened 2025-07-07
fix: Webhook TLS certs are not reloaded automatically in AKO standby replica pod when rotated [KO-423]

jwalantmodi05:


cmd/main.go
```diff
@@ -49,6 +49,18 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+// nonLeaderCertWatcher is a wrapper around certwatcher.CertWatcher
+// that doesn't require leader election.
+type nonLeaderCertWatcher struct {
```
abhishekdwivedi3060:
```suggestion
type customCertWatcher struct {
```
explain the objective in comments

abhishekdwivedi3060 (approved):

PR #393, opened 2025-07-24
Adding support for admin port configurations. [KO-437]

tanmayja:


api/v1/aerospikecluster_types.go
```diff
@@ -1173,6 +1173,14 @@ type AerospikePodStatus struct { //nolint:govet // for readability
 	// +optional
 	ServicePort int32 `json:"servicePort,omitempty"`
 
+	// PodAdminPort is the port users inside K8s can connect to.
+	// +optional
+	PodAdminPort int32 `json:"podAdminPort,omitempty"`
+
+	// ServiceAdminPort is the port users outside K8s can connect to.
+	// +optional
+	ServiceAdminPort int32 `json:"serviceAdminPort,omitempty"`
```
```suggestion
	// PodAdminPort is the admin port K8s internal Aerospike clients can connect to.
	// +optional
	PodAdminPort int32 `json:"podAdminPort,omitempty"`

	// ServiceAdminPort is the admin port Aerospike clients outside K8s can connect to
	// +optional
	ServiceAdminPort int32 `json:"serviceAdminPort,omitempty"`
```

Maybe add a line or two about the use of this port if possible

test/cluster/dynamic_config_test.go
```diff
@@ -1075,7 +1075,7 @@ func getAerospikeConfigFromNodeAndSpec(aeroCluster *asdbv1.AerospikeCluster) (fl
 
 	pod := aeroCluster.Status.Pods[pods.Items[0].Name]
 
-	host, err := createHost(&pod)
+	host, err := createHost(&pod, "service")
```
abhishekdwivedi3060: Use the already defined constant for "service" wherever applicable

internal/controller/cluster/configmap.go
```diff
@@ -281,6 +283,16 @@ func (r *SingleClusterReconciler) getBaseConfData(rack *asdbv1.Rack) (map[string
...
+	var adminTLSPortParam int32
+	if _, adminTLSPort := asdbv1.GetAdminTLSNameAndPort(asConfig); adminTLSPort != nil {
+		adminTLSPortParam = *adminTLSPort
+	}
+
+	var adminPortParam int32
+	if adminPort := asdbv1.GetAdminPort(asConfig); adminPort != nil {
+		adminPortParam = *adminPort
+	}
+
```
abhishekdwivedi3060: Same here not required

internal/controller/cluster/configmap.go
```diff
@@ -50,6 +50,8 @@ type initializeTemplateInput struct {
 	HeartBeatPort    int32
 	HeartBeatTLSPort int32
 	FabricTLSPort    int32
+	AdminPort        int32
+	AdminTLSPort     int32
```
abhishekdwivedi3060: Not required. This templateInput is for script template substitution. We no longer use it.

internal/controller/cluster/pod.go
```diff
@@ -147,8 +147,19 @@ func (r *SingleClusterReconciler) getRollingRestartTypeMap(rackState *RackState,
...
+				// Fetching all block devices that have been added in namespaces.
+				addedNSDevices, err = r.getNSAddedDevices(rackState)
+				if err != nil {
+					return nil, nil, err
+				}
+			}
 		}
 
+		// Fetching all pods requested for on-demand operations.
+		onDemandQuickRestarts, onDemandPodRestarts := r.podsToRestart()
```
abhishekdwivedi3060: Do we need to calculate this for every pod?
Move it outside the pod loop similar to previous flow

internal/controller/cluster/pod.go
```diff
@@ -1763,3 +1767,19 @@ func (r *SingleClusterReconciler) shouldSetMigrateFillDelay(rackState *RackState
 
 	return false
 }
+
+func (r *SingleClusterReconciler) isAnyPodSpecUpdated(rackState *RackState,
```
abhishekdwivedi3060, line 1771: This is kind of confusing/incomplete. Func says isAnyPodSpecUpdated but we are only passing the server container.

Suggestion:
Can we make it generic and implement it based on separation of concern like pass the required pod and update the whole dummy sts. In a lower level func (maybe isPortSpecChanged) check the port change. This way we'll keep it open for future extension.

internal/webhook/v1/aerospikecluster_mutating_webhook.go
```diff
@@ -595,10 +595,21 @@ func setDefaultNetworkConf(
 	)
 
 	// Fabric section
-	if _, ok := networkConf["fabric"]; !ok {
+	if _, ok = networkConf["fabric"]; !ok {
```
abhishekdwivedi3060, line 598: Just curious: Is there any reason why we are not checking the `map[string]interface{}` typecasting for fabric type?

pkg/validation/validate.go
```diff
@@ -822,13 +822,29 @@ func validateNsConfUpdate(oldConf, newConf map[string]interface{}) error {
...
 func validateNetworkConnectionUpdate(oldConf, newConf map[string]interface{}, connectionType string) error {
-	var networkPorts = []string{
-		"port", "access-port",
-		"alternate-access-port",
+	var (
+		networkPorts = []string{
+			"port", "access-port",
+			"alternate-access-port",
+		}
+		oldConnectionConfig, newConnectionConfig map[string]interface{}
```
abhishekdwivedi3060: Not required. Already declared and initialised below.

pkg/validation/validate.go
```diff
@@ -822,13 +822,29 @@ func validateNsConfUpdate(oldConf, newConf map[string]interface{}) error {
...
+	if !oldOk || !newOk {
+		return fmt.Errorf("invalid network configuration structure")
 	}
 
-	oldConnectionConfig := oldConf["network"].(map[string]interface{})[connectionType].(map[string]interface{})
-	newConnectionConfig := newConf["network"].(map[string]interface{})[connectionType].(map[string]interface{})
+	oldConnectionConfig, oldConnOk := oldNetwork[connectionType].(map[string]interface{})
+	newConnectionConfig, newConnOk := newNetwork[connectionType].(map[string]interface{})
+
+	// If the connectionType is missing in either old or new config, assume it's an admin and skip validation
```
abhishekdwivedi3060:
```suggestion
	// If the connectionType is missing in either old or new config, assume it's an admin and skip validation.
	// Other connectionType are required fields and their existence is checked in mutation.
```

test/cluster/cluster_test.go
```diff
@@ -20,7 +20,12 @@ import (
 	lib "github.com/aerospike/aerospike-management-lib"
 )
 
-var _ = Describe(
+const (
+	clusterNameConfig = "cluster-name"
+	serviceConfig     = "service"
+)
+
+var _ = FDescribe(
```
abhishekdwivedi3060: Revert

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+					},
+				)
+			},
+		)
 	},
 )
 
+func adminPortTests(ctx goctx.Context) {
+	Context(
+		"AdminPort", func() {
```
abhishekdwivedi3060: nit: `AdminPort` context already defined in the caller func

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+			aeroCluster := &asdbv1.AerospikeCluster{}
+			customInitImage := "aerospike-kubernetes-init:2.3.0-dev56"
+
+			AfterEach(
+				func() {
+					Expect(DeleteCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+					Expect(CleanupPVC(k8sClient, aeroCluster.Namespace, aeroCluster.Name)).ToNot(HaveOccurred())
+				},
+			)
+			It(
```
abhishekdwivedi3060: Add test-case for admintPort with MultiPodPerHost: true to check k8s service flow

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+					// Create cluster with admin port configuration
+					aeroCluster = createDummyAerospikeCluster(clusterNamespacedName, 2)
+					aeroCluster.Spec.Image = fmt.Sprintf("%s:%s", baseImage, "8.1.0.0-rc3")
+					aeroCluster.Spec.PodSpec.AerospikeInitContainerSpec.ImageRegistryNamespace = ptr.To("tanmayj10")
+					aeroCluster.Spec.PodSpec.AerospikeInitContainerSpec.ImageNameAndTag = customInitImage
+					aeroCluster.Spec.PodSpec.MultiPodPerHost = ptr.To(false)
+
+					// Add admin port configuration to network
+					networkConf := aeroCluster.Spec.AerospikeConfig.Value["network"].(map[string]interface{})
+					networkConf["admin"] = map[string]interface{}{
```
abhishekdwivedi3060: use constant for `admin`, `network` wherever applicable

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+			It(
+				"Should create cluster with admin port and connect successfully",
+				func() {
+					clusterName := fmt.Sprintf("admin-port-cluster-%d", GinkgoParallelProcess())
+					clusterNamespacedName := test.GetNamespacedName(
+						clusterName, namespace,
+					)
+
+					// Create cluster with admin port configuration
+					aeroCluster = createDummyAerospikeCluster(clusterNamespacedName, 2)
```
abhishekdwivedi3060: How about we write a wrapper for `createDummyAerospikeClusterWithAdminPort`?

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+
+					asinfo, err := getASInfo(logger, k8sClient, ctx, clusterNamespacedName, clusterNamespacedName.Name+"-0-0", "admin")
+					Expect(err).ToNot(HaveOccurred())
+
+					confs, err := getAsConfig(asinfo, "network")
+					Expect(err).ToNot(HaveOccurred())
+
+					// Verify admin port is set to 3003
+					network := confs["network"].(lib.Stats)
+					Expect(network["admin.port"]).To(Equal(int64(3003)))
```
abhishekdwivedi3060: Can be moved to a func called `validateAdminPort`

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+				},
+			)
+		},
+	)
+}
+
+// Network-related validation tests
+func NetworkValidationTest(ctx goctx.Context) {
+	Context(
+		"NetworkValidation", func() {
```
abhishekdwivedi3060: Same here

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+						clusterNamespacedName, 1,
+					)
+					aeroCluster.Spec.AerospikeConfig.
+						Value[serviceConfig].(map[string]interface{})[clusterNameConfig] = clusterNameConfig
+					Expect(DeployCluster(k8sClient, ctx, aeroCluster)).Should(HaveOccurred())
+				},
+			)
+		},
+	)
+}
```
abhishekdwivedi3060: Duplicate test-case. Already present in `negativeDeployClusterValidationTest`. Not required

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
...
+
+					aeroCluster.Spec.AerospikeConfig.
+						Value[serviceConfig].(map[string]interface{})[clusterNameConfig] = clusterNameConfig
+					err = k8sClient.Update(ctx, aeroCluster)
+					Expect(err).Should(HaveOccurred())
+				},
+			)
+		},
+	)
+}
```
abhishekdwivedi3060: Same here

test/cluster/cluster_test.go
```diff
@@ -87,9 +87,511 @@ var _ = Describe(
 				ValidateAerospikeBenchmarkConfigs(ctx)
 			},
 		)
+		// Network-related tests
+		Context(
+			"Network", func() {
+				Context(
+					"UpdateTLSCluster", func() {
+						UpdateTLSClusterTest(ctx)
```
abhishekdwivedi3060, line 96: Did we add any test-case for the case we found?  MultiPodPerHost: false and Tls to both update -> Should result in Pod restart

abhishekdwivedi3060 (changes requested):

internal/webhook/v1/aerospikecluster_mutating_webhook.go
tanmayja: no harm in validating, but in fabric we don't set any default value hence not checking for typecasting here and relying on schema validation.

test/cluster/cluster_test.go
tanmayja: "Should update cluster by removing admin connection"
This test will cover scenario where cluster is MultiPodPerHost: true with admin network.

internal/controller/cluster/pod.go
tanmayja: - yes, we can pass full pod instead of server container for more generic future use.
- I’d prefer not to populate a dummy StatefulSet with all fields unnecessarily, unless we plan to pass the updated StatefulSet further down where it's actually applied to Kubernetes. Introducing that change now would be quite significant.
If we leave optimization aside, we can update the whole dummy sts.

test/cluster/cluster_test.go
tanmayja: These changes will make the PR really big, as it is not in just this file.

test/cluster/cluster_test.go
tanmayja: Instead validating if port is reflected in containers spec.

internal/controller/cluster/pod.go
```diff
@@ -102,15 +102,18 @@ func (r *SingleClusterReconciler) getRollingRestartTypeMap(rackState *RackState,
...
 		if podStatus.AerospikeConfigHash != requiredConfHash {
-			if addedNSDevices == nil {
-				// Fetching all block devices that have been added in namespaces.
-				addedNSDevices, err = r.getNSAddedDevices(rackState)
-				if err != nil {
-					return nil, nil, err
-				}
+			serverContainer := getContainer(pods[idx].Spec.Containers, asdbv1.AerospikeServerContainerName)
+
+			podSpecUpdated, err := r.isAnyPodSpecUpdated(rackState, pods[idx])
```
abhishekdwivedi3060, line 105: nit:
move this above `getContainer`

test/cluster/cluster_helper.go
```diff
@@ -484,6 +488,40 @@ func validateServiceUpdate(k8sClient client.Client, ctx goctx.Context,
...
+	}
+
+	for podName := range aeroCluster.Status.Pods {
+		podNamespaceName := test.GetNamespacedName(podName, aeroCluster.Namespace)
+		pod := &corev1.Pod{}
+
+		err = k8sClient.Get(ctx, podNamespaceName, pod)
+		if err != nil {
+			return err
+		}
```
abhishekdwivedi3060: Any reason for doing like this one by one?
Why not fetch `getClusterPodList` in one go and loop over that.

test/cluster/cluster_helper.go
```diff
@@ -484,6 +488,40 @@ func validateServiceUpdate(k8sClient client.Client, ctx goctx.Context,
...
+
+		portSet := sets.NewInt32(ports...)
+
+		println(fmt.Sprintf("pod %s, ports: %v", podName, pod.Spec.Containers[0].Ports))
```
abhishekdwivedi3060: nit: pkgLog.Info()

test/cluster/cluster_test.go
```diff
@@ -87,9 +86,426 @@ var _ = Describe(
...
+	It(
+		"Should create cluster with admin port and connect successfully",
+		func() {
+			clusterName := fmt.Sprintf("admin-port-cluster-%d", GinkgoParallelProcess())
+			clusterNamespacedName := test.GetNamespacedName(
+				clusterName, namespace,
+			)
+
+			// Create cluster with admin port configuration
+			aeroCluster = createDummyAerospikeClusterWithAdminPort(clusterNamespacedName, 2, 3003)
```
abhishekdwivedi3060: We can define an adminPort constant above

abhishekdwivedi3060 (commented):
LGTM. Will approve once all the final changes (like init tag, server image updates) are done.

PR #399, opened 2025-08-25
enhancement: Adding support for hostpath volume type. [KO-382] 

tanmayja:
The following changes are included in this PR:

1. Added support for hostPath volume mounts as a read-only file system.
2. Enabled proper application of mountOptions in pods, addressing a previous issue where specified mount options in the volume configuration had no effect, and StatefulSets (STS) were not updated accordingly.

-  After upgrading to this AKO version, if mount options have already been specified by the user, they will now be updated in the StatefulSet. However, pods will not undergo a rolling restart automatically—mount options will be applied only after subsequent pod restart.
-  For clusters created after this update, the mount options will be consistently applied to pods as expected.

https://aerospike.atlassian.net/wiki/spaces/KO/pages/4524277974/HostPath+Volume+Support

### DIFF
--- a/test/cluster/storage_test.go
+++ b/test/cluster/storage_test.go
@@ -362,6 +362,24 @@ var _ = Describe(
 							func() {
 								aeroCluster := createDummyAerospikeCluster(clusterNamespacedName, 2)
 
+								mountOptionsForContainers := []asdbv1.MountOptions{
+									{
+										ReadOnly:         ptr.To(true),
+										SubPath:          "subdir",
+										MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
+									},
+									{
+										ReadOnly:         ptr.To(true),
+										SubPath:          "custom-init-subdir",
+										MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
+									},
+									{
+										ReadOnly:         ptr.To(true),
+										SubPath:          "sidecar-subdir",
+										MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
+									},
+								}
+
 								// Add sidecar container to pod spec
 								aeroCluster.Spec.PodSpec.Sidecars = []v1.Container{
 									{
@@ -401,11 +419,7 @@ var _ = Describe(
 									Aerospike: &asdbv1.AerospikeServerVolumeAttachment{
 										Path: "/mnt/hostpath-data",
 										AttachmentOptions: asdbv1.AttachmentOptions{
-											MountOptions: asdbv1.MountOptions{
-												ReadOnly:         ptr.To(true),
-												SubPath:          "subdir",
-												MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
-											},
+											MountOptions: mountOptionsForContainers[0],
 										},
 									},
 									// Init container mount options
@@ -414,11 +428,7 @@ var _ = Describe(
 											ContainerName: "custom-init-container",
 											Path:          "/mnt/hostpath-custom-init",
 											AttachmentOptions: asdbv1.AttachmentOptions{
-												MountOptions: asdbv1.MountOptions{
-													ReadOnly:         ptr.To(true),
-													SubPath:          "custom-init-subdir",
-													MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
-												},
+												MountOptions: mountOptionsForContainers[1],
 											},
 										},
 									},
@@ -428,11 +438,7 @@ var _ = Describe(
 											ContainerName: "sidecar-container",
 											Path:          "/mnt/hostpath-sidecar",
 											AttachmentOptions: asdbv1.AttachmentOptions{
-												MountOptions: asdbv1.MountOptions{
-													ReadOnly:         ptr.To(true),
-													SubPath:          "sidecar-subdir",
-													MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
-												},
+												MountOptions: mountOptionsForContainers[2],
 											},
 										},
 									},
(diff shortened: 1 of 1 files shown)


test/cluster/storage_test.go
```diff
@@ -573,9 +574,19 @@ var _ = Describe(
 									aeroCluster.Spec.Storage.Volumes[0].Sidecars,
 									va,
 								)
+
+								aeroCluster.Spec.Storage.Volumes = append(aeroCluster.Spec.Storage.Volumes,
+									getHostPathStorageVolumeForSidecar("sidecar-volume",
+										"/para/tomcat1", containerName, true))
+
```
abhishekdwivedi3060: Need to add test webhook readOnly: false validation in webhook

internal/controller/cluster/statefulset.go
```diff
@@ -1441,10 +1442,14 @@ func getFinalVolumeAttachmentsForVolume(volume *asdbv1.VolumeSpec) (
 	initContainerAttachments = append(
 		initContainerAttachments, volume.InitContainers...,
 	)
+
 	initContainerAttachments = append(
 		initContainerAttachments, asdbv1.VolumeAttachment{
 			ContainerName: asdbv1.AerospikeInitContainerName,
 			Path:          initVolumePath,
+			AttachmentOptions: asdbv1.AttachmentOptions{
+				MountOptions: volume.Aerospike.MountOptions,
```
abhishekdwivedi3060: Aerospike MountOptions can be nil

internal/controller/cluster/statefulset.go
```diff
@@ -1467,23 +1475,29 @@ func addVolumeMountInContainer(
...
+					mountPath = volumeAttachment.Path
+				}
+
+				volumeMount = corev1.VolumeMount{
+					Name:             volumeName,
+					MountPath:        mountPath,
+					ReadOnly:         asdbv1.GetBool(volumeAttachment.AttachmentOptions.MountOptions.ReadOnly),
+					SubPath:          volumeAttachment.AttachmentOptions.MountOptions.SubPath,
+					SubPathExpr:      volumeAttachment.AttachmentOptions.MountOptions.SubPathExpr,
+					MountPropagation: volumeAttachment.AttachmentOptions.MountOptions.MountPropagation,
```
abhishekdwivedi3060, line 1503: Can we add/update some test-cases to confirm if these values are actually reflect in the pod spec?

test/cluster/storage_test.go
```diff
@@ -352,6 +354,174 @@ var _ = Describe(
...
+								// Validate Aerospike server container mount options
+								aerospikeMountOptions, err := getVolumeMountOptions(podNamespacedName,
+									asdbv1.AerospikeServerContainerName, "hostpath-mount-options-test")
+								Expect(err).ToNot(HaveOccurred())
+
+								Expect(asdbv1.GetBool(aerospikeMountOptions.ReadOnly)).To(BeTrue())
+								Expect(aerospikeMountOptions.SubPath).To(Equal("subdir"))
+								Expect(aerospikeMountOptions.SubPathExpr).To(BeEmpty())
+								Expect(*aerospikeMountOptions.MountPropagation).To(Equal(v1.MountPropagationHostToContainer))
```
abhishekdwivedi3060: Can be moved to a common func like `validateMountOptions` and use current and desired params for validation

test/cluster/storage_test.go
```diff
@@ -352,6 +354,174 @@ var _ = Describe(
...
+									// Aerospike server container mount options
+									Aerospike: &asdbv1.AerospikeServerVolumeAttachment{
+										Path: "/mnt/hostpath-data",
+										AttachmentOptions: asdbv1.AttachmentOptions{
+											MountOptions: asdbv1.MountOptions{
+												ReadOnly:         ptr.To(true),
+												SubPath:          "subdir",
+												MountPropagation: &[]v1.MountPropagationMode{v1.MountPropagationHostToContainer}[0],
+											},
+										},
```
abhishekdwivedi3060: See if we can move it to var or func as desiredMountOptions and use this for validation the mountOptions in the pod

test/cluster/storage_test.go
```diff
@@ -597,6 +768,56 @@ var _ = Describe(
...
+// getVolumeMountOptions retrieves mount options for a specific volume in a container
+//
+//nolint:unparam // generic function
+func getVolumeMountOptions(podNamespacedName types.NamespacedName, containerName,
+	volumeName string) (*asdbv1.MountOptions, error) {
+	updatedPod := &v1.Pod{}
+
+	if err := k8sClient.Get(context.TODO(), podNamespacedName, updatedPod); err != nil {
+		return nil, err
+	}
```
abhishekdwivedi3060: See if we can move this get pod out of this func. We are calling this `getVolumeMountOptions` multiple times for the same pod, resulting in multiple get pod calls. We can pass the pod object as a param

test/cluster/storage_test.go
```diff
@@ -352,6 +354,174 @@ var _ = Describe(
...
+								Expect(asdbv1.GetBool(sidecarMountOptions.ReadOnly)).To(BeTrue())
+								Expect(sidecarMountOptions.SubPath).To(Equal("sidecar-subdir"))
+								Expect(aerospikeMountOptions.SubPathExpr).To(BeEmpty())
+								Expect(*sidecarMountOptions.MountPropagation).To(Equal(v1.MountPropagationHostToContainer))
+
+								volumeIndex := len(aeroCluster.Spec.Storage.Volumes) - 1
+
+								// Test with read-write mount (should fail due to validation)
+								aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+								Expect(err).ToNot(HaveOccurred())
```
abhishekdwivedi3060: We don't need to fetch the cluster again and again. `updateCluster` func already does that internally

abhishekdwivedi3060 (changes requested):

jwalantmodi05 (approved):

PR #404, opened 2025-09-03
Add wait for migration flow before deletion of pod [KO-444]

jwalantmodi05:


internal/controller/cluster/rack.go
```diff
@@ -974,6 +975,42 @@ func (r *SingleClusterReconciler) scaleDownRack(
...
+			ignorablePodNames,
+		); !res.IsSuccess {
+			r.Log.Error(
+				res.Err, "Failed to wait for migration to complete before deleting pods",
+				"rackID", rackState.Rack.ID,
+			)
+
+			return found, res
+		}
+	}
```
abhishekdwivedi3060: In case of failed pod also, I think we should wait for migration to complete. What if the pods recently failed and migration are still going on. We should make sure the migration are complete in any case

internal/controller/cluster/rack.go
```diff
@@ -974,6 +975,42 @@ func (r *SingleClusterReconciler) scaleDownRack(
...
+	for idx := range rackStateList {
+		if rackStateList[idx].Rack.ID == rackState.Rack.ID {
+			newDesiredSize = rackStateList[idx].Size
+			break
+		}
+	}
+	// If scale down is not needed then requeue
+	// This can happen if user reverts the size back to original value before scale down could complete
+	if *found.Spec.Replicas <= newDesiredSize {
+		return found, common.ReconcileRequeueAfter(1)
```
abhishekdwivedi3060: Can you explain what is going on here?
If we have reached this stage in the code flow then it means scale down is required. If user reverts the scale down operation when the migrations are happening, then that operation should be handled in the next reconcile. And in the next reconcile after revert operation, code flow won't reach this stage

abhishekdwivedi3060 (changes requested):
Add test-cases

test/cluster/cluster_helper.go
```diff
@@ -616,6 +617,45 @@ func validateMigrateFillDelay(
...
+	if err != nil {
+		return -1, err
+	}
+
+	maxMigrations := 0
+
+	for idx := range podList.Items {
+		pod := &podList.Items[idx]
+
+		asinfo, err := getASInfo(logger, k8sClient, ctx, clusterNamespacedName, pod.Name, "statistics")
```
tanmayja: cross-check how to call getASInfo() function, last argument "statistics" passsed here is not correct.

internal/controller/cluster/aero_info_calls.go
```diff
@@ -82,6 +82,22 @@ func (r *SingleClusterReconciler) waitForMultipleNodesSafeStopReady(
 	return common.ReconcileSuccess()
 }
 
+// waitForMigrationToComplete waits for the migration to complete on all the nodes in the cluster.
+func (r *SingleClusterReconciler) waitForMigrationToComplete(ignorablePodNames sets.Set[string],
```
abhishekdwivedi3060: We can pass the policy as param similar to setMigrateFillDelay func

test/cluster/cluster_helper.go
```diff
@@ -616,6 +617,45 @@ func validateMigrateFillDelay(
...
+	clusterNamespacedName types.NamespacedName) (int, error) {
+	aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+	if err != nil {
+		return -1, err
+	}
+
+	podList, err := getPodList(aeroCluster, k8sClient)
+	if err != nil {
+		return -1, err
+	}
```
abhishekdwivedi3060: This info can be passed as param since we are already fetching this outside

test/cluster/cluster_test.go
```diff
@@ -767,6 +772,74 @@ func ScaleDownWithMigrateFillDelay(ctx goctx.Context) {
...
+								utils.Len32(podList.Items), migrations))
+						}
+
+						// Success condition: migrations complete and number of pods is equal to spec size
+						if utils.Len32(podList.Items) == aeroCluster.Spec.Size && migrations == 0 {
+							return true
+						}
+						// Keep waiting otherwise
+						return false
+					}, 5*time.Minute, 10*time.Second).Should(BeTrue(), "Pods should only scaledown after migrations complete")
```
abhishekdwivedi3060: Are we sure that this 10s interval is sufficient to check the required condition?
Seems like here there may be race conditions here. Is it possible to check the expected behaviour like 4 nodes + 1 quiesced  + migration non-zeo? This is confirm the behaviour as per expected change

PR #408, opened 2025-09-17
feat: Introduce support for removing a rack from the roster in a running cluster [KO-452]

tanmayja:


api/v1/aerospikecluster_types.go
```diff
@@ -533,6 +533,12 @@ type Rack struct { //nolint:govet // for readability
 	// Effective/operative PodSpec. The resultant is user input if specified else global PodSpec
 	// +optional
 	PodSpec RackPodSpec `json:"effectivePodSpec,omitempty"`
+
+	// ForceBlockFromRoster when set to true, removes all nodes from this rack from the roster
+	// If set to true, the operator will ignore unavailable/dead partitions and go ahead and remove the nodes from roster
+	// Hence, this should be used with caution.
```
```suggestion
	// ForceBlockFromRoster, when set to true, forcefully removes all nodes in this rack from the roster.
   // With this enabled, the operator ignores unavailable or dead partitions and proceeds with node removal.
   // It may result in data loss. Use this with caution.
```

internal/controller/cluster/aero_info_calls.go
```diff
@@ -82,6 +84,54 @@ func (r *SingleClusterReconciler) waitForMultipleNodesSafeStopReady(
...
+		nodeID := pods[idx].Aerospike.NodeID
+		// ensure nodeID is long enough to contain rackID + infix + ordinal
+		if len(nodeID) <= len(rackIDStr) {
+			continue
+		}
+
+		if strings.HasPrefix(nodeID, rackIDStr) {
+			return string(nodeID[len(rackIDStr)]) // middle character
+		}
+	}
```
abhishekdwivedi3060: Couple of cases missed:
1. This prefix check is not full-proof. It will return true even if node is 12a0 and rackID is 1.
2. nodeID in status is in small cases but server shows observed nodes in capital case. See if this is working properly while ignoring the nodes from roster

internal/controller/cluster/aero_info_calls.go
```diff
@@ -82,6 +84,54 @@ func (r *SingleClusterReconciler) waitForMultipleNodesSafeStopReady(
...
+		}
+
+		rackID := rackState.Rack.ID
+		infix := getNodeIDInfixForRack(r.aeroCluster.Status.Pods, rackID)
+
+		for podOrdinal := int32(0); podOrdinal < rackState.Size; podOrdinal++ {
+			nodeID := fmt.Sprintf("%d%s%d", rackID, infix, podOrdinal)
+			nodeIDs = append(nodeIDs, nodeID)
+
+			r.Log.V(1).Info("Added rack node to comprehensive block list",
```
abhishekdwivedi3060: See if it makes sense to print the complete list at the very end instead of pod by pod log just to avoid log clutter
```suggestion
			r.Log.V(1).Info("Added rack node to comprehensive roster block list",
```

internal/controller/cluster/aero_info_calls.go
abhishekdwivedi3060: Think this from a vertical scaling point of view: When the rack migration to the new revision is going on and both old and new rack revisions co-exist. At this point, status will have 2 infix for the same rackID. I think this will work eventually but lets confirm this with the current code flow

internal/controller/cluster/rack.go
```diff
@@ -276,6 +276,23 @@ func (r *SingleClusterReconciler) createEmptyRack(rackState *RackState) (
 	return found, common.ReconcileSuccess()
 }
 
+// getRacksToBeBlockedFromRoster identifies racks that should have their nodes blocked from roster
+// and returns a slice of racks that have ForceBlockFromRoster: true
```
abhishekdwivedi3060, line 230:
```suggestion
// and returns a slice of rackIDs that have ForceBlockFromRoster: true
```

api/v1/aerospikecluster_types.go
```diff
@@ -533,6 +533,12 @@ type Rack struct { //nolint:govet // for readability
 	// Effective/operative PodSpec. The resultant is user input if specified else global PodSpec
 	// +optional
 	PodSpec RackPodSpec `json:"effectivePodSpec,omitempty"`
+
+	// ForceBlockFromRoster when set to true, removes all nodes from this rack from the roster
+	// If set to true, the operator will ignore unavailable/dead partitions and go ahead and remove the nodes from roster
```
abhishekdwivedi3060: Lets also mention a point here that failed/pending pods in this rack are also ignored with this flag

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -510,11 +512,47 @@ func validateRackUpdate(
...
+	}
+
+	for idx := range newObj.Status.RackConfig.Racks {
+		rack := newObj.Status.RackConfig.Racks[idx]
+		if asdbv1.GetBool(rack.ForceBlockFromRoster) {
+			racksBlockedFromRosterInStatus.Insert(rack.ID)
+		}
+	}
+
+	if len(newObj.Status.RackConfig.Racks)-len(racksBlockedFromRosterInSpec) == 1 &&
```
abhishekdwivedi3060: What's the reason for comparing this with status RackConfig instead of spec RackConfig?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -510,11 +512,47 @@ func validateRackUpdate(
...
+		(newObj.Spec.RackConfig.RollingUpdateBatchSize != nil || newObj.Spec.RackConfig.ScaleDownBatchSize != nil) {
+		return fmt.Errorf("with only one rack remaining in roster, cannot use batch update or scale down")
```
abhishekdwivedi3060: Fix the error msg with actual field names like `rollingUpdateBatchSize`

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -510,11 +512,47 @@ func validateRackUpdate(
...
+	}
+
+	desiredRacksBlockedFromRoster := racksBlockedFromRosterInSpec.Difference(racksBlockedFromRosterInStatus)
+
+	if len(desiredRacksBlockedFromRoster) > 1 {
+		return fmt.Errorf("can change only one rack at a time to ForceBlockFromRoster: true")
```
```suggestion
		return fmt.Errorf("Only one rack can be force-blocked from the roster at a time using the forceBlockFromRoster flag")
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -640,6 +680,24 @@ func validateRackConfig(_ logr.Logger, cluster *asdbv1.AerospikeCluster) error {
...
 		migrateFillDelaySet.Insert(migrateFillDelay)
+
+		if asdbv1.GetBool(rack.ForceBlockFromRoster) {
+			racksBlockedFromRoster++
+		}
+	}
+
+	if racksBlockedFromRoster > 0 {
+		if cluster.Spec.RackConfig.MaxIgnorablePods != nil {
+			return fmt.Errorf("ForceBlockFromRoster: true racks cannot be used with MaxIgnorablePods set")
```
abhishekdwivedi3060: Use json tags in the error msg instead of Go field name. Check error msgs at other places as well
```suggestion
			return fmt.Errorf("forceBlockFromRoster cannot be enabled when maxIgnorablePods is set")
```

test/cluster/strong_consistency_test.go
```diff
@@ -109,6 +112,83 @@ var _ = Describe("SCMode", func() {
...
+				Racks:      racks,
+			}
+			aeroCluster.Spec.RackConfig = rackConf
+
+			Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+			validateRoster(k8sClient, ctx, clusterNamespacedName, scNamespace)
+
+			By("Block rack 1 from roster")
+			//		expectedRoster := []string{"2A0@2", "2A1@2"}
```
abhishekdwivedi3060: ??

test/cluster/strong_consistency_test.go
```diff
@@ -109,6 +112,83 @@ var _ = Describe("SCMode", func() {
...
+
+			rosterNodesMap, err = getRoster(hostConns[2], getClientPolicy(aeroCluster, k8sClient), aeroCluster.Namespace)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Roster is in uppercase, whereas nodeID is in lower case in server. Keep it in mind when comparing list
+			rosterStr = rosterNodesMap["roster"]
+			Expect(rosterStr).To(Equal(expectedRoster))
+
+		})
+
```
abhishekdwivedi3060, line 196: Won't this flow make the pod of 2nd rack also unschedulable?

test/cluster/strong_consistency_test.go
```diff
@@ -356,7 +436,80 @@ var _ = Describe("SCMode", func() {
...
+				Racks:      racks,
+			}
+			aeroCluster.Spec.RackConfig = rackConf
+
+			Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+			aeroCluster.Spec.RackConfig.Racks[0].ForceBlockFromRoster = ptr.To(true)
 
+			err := updateCluster(k8sClient, ctx, aeroCluster)
+			Expect(err).ToNot(HaveOccurred())
```
abhishekdwivedi3060: nit: This can be clubbed with below update operation for invalid operations

abhishekdwivedi3060 (changes requested):

internal/controller/cluster/aero_info_calls.go
tanmayja: 1. Valid observation; this will be corrected.
2. The management library performs case-insensitive comparisons internally using Go’s strings.EqualFold function, which allows it to work regardless of case. However, no harm in sending the input in uppercase.

test/cluster/strong_consistency_test.go
tanmayja: why?

internal/controller/cluster/pod.go
```diff
@@ -905,13 +905,17 @@ func (r *SingleClusterReconciler) cleanupDanglingPodsRack(sts *appsv1.StatefulSe
 // 1. From racksToDelete that are currently not running and can be ignored in stability checks.
 // 2. Failed/pending pods from the configuredRacks identified using maxIgnorablePods field and
 // can be ignored from stability checks.
-func (r *SingleClusterReconciler) getIgnorablePods(racksToDelete []asdbv1.Rack, configuredRacks []RackState) (
+func (r *SingleClusterReconciler) getIgnorablePods(racksToDelete []int, configuredRacks []RackState) (
```
abhishekdwivedi3060: We might have to revert this racksToDelete int change. For future changes, we need both id and revision to uniquely identify a rack to delete which is not possible with new changes

test/cluster/strong_consistency_test.go
```diff
@@ -109,6 +112,83 @@ var _ = Describe("SCMode", func() {
...
+
+			By("Block rack 1 from roster")
+			//		expectedRoster := []string{"2A0@2", "2A1@2"}
+			expectedRoster := "2A1@2,2A0@2"
+			aeroCluster.Spec.RackConfig.Racks[0].ForceBlockFromRoster = ptr.To(true)
+
+			Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+			aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
```
jwalantmodi05, line 140: is this `getCluster()` required here?

tanmayja: yes, this getcluster is fetching current CR with updated ACL details in status.

test/cluster/strong_consistency_test.go
```diff
@@ -356,7 +435,81 @@ var _ = Describe("SCMode", func() {
...
+			Expect(err).Should(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(
+				"ForceBlockFromRoster: true racks cannot be used with MaxIgnorablePods set"))
+
+			aeroCluster.Spec.RackConfig.MaxIgnorablePods = nil
+			aeroCluster.Spec.RosterNodeBlockList = []string{"1A0", "1A1"}
+			aeroCluster.Spec.RackConfig.Racks[0].ForceBlockFromRoster = ptr.To(true)
+			err = updateCluster(k8sClient, ctx, aeroCluster)
+			Expect(err).Should(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("ForceBlockFromRoster: true racks cannot be used with RosterNodeBlockList"))
```
abhishekdwivedi3060: Need to change the error msg as per new changes

test/cluster/strong_consistency_test.go
```diff
@@ -356,7 +435,81 @@ var _ = Describe("SCMode", func() {
...
+			aeroCluster.Spec.RackConfig.RollingUpdateBatchSize = &intstr.IntOrString{IntVal: 2}
+
+			err := updateCluster(k8sClient, ctx, aeroCluster)
+			Expect(err).Should(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(
+				"with only one rack remaining in roster, cannot use rollingUpdateBatchSize or scaleDownBatchSize"))
+
+			aeroCluster.Spec.RackConfig.RollingUpdateBatchSize = nil
+			aeroCluster.Spec.RackConfig.MaxIgnorablePods = &intstr.IntOrString{IntVal: 2}
+			aeroCluster.Spec.RackConfig.Racks[0].ForceBlockFromRoster = ptr.To(true)
```
abhishekdwivedi3060: No need to set it again and again

PR #409, opened 2025-09-17
feat: Add support for vertical scaling (storage scaling) [KO-448] [KO-459]

abhishekdwivedi3060:
This PR introduces the Rack Revision feature that enables safe, cost-effective storage updates for Aerospike clusters by implementing a gradual pod-by-pod migration strategy instead of the old add new rack approach. Old approach required rack-aware clients to change their configuration whenever there is storage update change by adding a new rack.

Before: Storage updates required bringing up an entirely new rack before deleting the old one, leading to:
🔴 Double infrastructure cost during migration
🔴 Rack-id changes
🔴 Client configuration changes for rack-aware clients

After: Gradual pod-by-pod migration with:
✅ Gradual rollout of new changes without changing the rack-id
✅ No client side changes

RackRevision Field
```
spec:
  rackConfig:
    racks:
    - id: 1
      revision: "v2"  # New field for versioned storage updates
      storage:
        # Updated storage configuration
```

### DIFF
--- a/internal/webhook/v1/aerospikecluster_validating_webhook.go
+++ b/internal/webhook/v1/aerospikecluster_validating_webhook.go
@@ -1094,7 +1094,6 @@ func validateNsConfUpdateFromStatus(newConfSpec, currentStatus *asdbv1.Aerospike
 	newConf := newConfSpec.Value
 	newNsConfList := newConf["namespaces"].([]interface{})
 
-	// TODO: Do we need to skip this validation for Revision update?
 	return validateStorageEngineDeviceListUpdate(newNsConfList, statusNsConfList)
 }
 
--- a/test/cluster/rack_revision_test.go
+++ b/test/cluster/rack_revision_test.go
@@ -14,7 +14,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
@@ -369,9 +368,6 @@ func createDummyClusterWithRackRevision(
 	clusterNamespacedName types.NamespacedName, revision string, size int32,
 ) *asdbv1.AerospikeCluster {
 	aeroCluster := createDummyAerospikeCluster(clusterNamespacedName, size)
-
-	aeroCluster.Spec.PodSpec.AerospikeInitContainerSpec.ImageRegistryNamespace = ptr.To("aerospike")
-	aeroCluster.Spec.PodSpec.AerospikeInitContainerSpec.ImageNameAndTag = "aerospike-kubernetes-init:2.4.0-dev1"
 
 	racks := []asdbv1.Rack{
 		{ID: 1, Revision: revision},
--- a/.github/workflows/docker-image-release.yaml
+++ b/.github/workflows/docker-image-release.yaml
@@ -2,7 +2,7 @@ name: Release Container Image
 on:
   push:
     tags:
-      - 'v*.*.*'
+      - 'v[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
   build-and-push:
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,7 +23,7 @@ pipeline {
 
         AEROSPIKE_CUSTOM_INIT_REGISTRY="568976754000.dkr.ecr.ap-south-1.amazonaws.com"
         AEROSPIKE_CUSTOM_INIT_REGISTRY_NAMESPACE="aerospike"
-        AEROSPIKE_CUSTOM_INIT_NAME_TAG="aerospike-kubernetes-init:2.4.0-dev1"
+        AEROSPIKE_CUSTOM_INIT_NAME_TAG="aerospike-kubernetes-init:2.4.0-dev2"
     }
 
     stages {
--- a/api/v1/utils.go
+++ b/api/v1/utils.go
@@ -81,7 +81,7 @@ const (
 	AerospikeInitContainerNameTagEnvVar            = "AEROSPIKE_KUBERNETES_INIT_NAME_TAG"
 	AerospikeInitContainerDefaultRegistry          = "docker.io"
 	AerospikeInitContainerDefaultRegistryNamespace = "aerospike"
-	AerospikeInitContainerDefaultNameAndTag        = "aerospike-kubernetes-init:2.4.0-dev1"
+	AerospikeInitContainerDefaultNameAndTag        = "aerospike-kubernetes-init:2.4.0-dev2"
 	AerospikeAppLabel                              = "app"
 	AerospikeAppLabelValue                         = "aerospike-cluster"
 	AerospikeCustomResourceLabel                   = "aerospike.com/cr"
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -71,7 +71,7 @@ spec:
           value: aerospike
         - name: AEROSPIKE_KUBERNETES_INIT_NAME_TAG
           # this is the name and tag of aerospike-init image
-          value: aerospike-kubernetes-init:2.4.0-dev1
+          value: aerospike-kubernetes-init:2.4.0-dev2
       volumes: []
       serviceAccountName: controller-manager
 
--- a/helm-charts/aerospike-kubernetes-operator/values.yaml
+++ b/helm-charts/aerospike-kubernetes-operator/values.yaml
@@ -44,7 +44,7 @@ aerospikeKubernetesInitRegistry: "docker.io"
 aerospikeKubernetesInitRegistryNamespace: "aerospike"
 
 # Name and tag of aerospike-init image
-aerospikeKubernetesInitNameTag: "aerospike-kubernetes-init:2.4.0-dev1"
+aerospikeKubernetesInitNameTag: "aerospike-kubernetes-init:2.4.0-dev2"
 
 ## Resources - limits / requests
 resources:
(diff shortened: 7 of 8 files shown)


internal/controller/cluster/rack.go
```diff
@@ -35,35 +39,27 @@ func (r *SingleClusterReconciler) reconcileRacks() common.ReconcileResult {
 		res                common.ReconcileResult
 	)
 
-	rackStateList := getConfiguredRackStateList(r.aeroCluster)
-
-	racksToDelete, err := r.getRacksToDelete(rackStateList)
+	configuredRacks, renamedRacks, racksToDelete, err := r.categoriseRacks()
 	if err != nil {
-		return common.ReconcileError(err)
+		return common.ReconcileResult{}
```
tanmayja: why not return the err?

api/v1/aerospikecluster_types.go
```diff
@@ -490,6 +490,13 @@ type Rack struct { //nolint:govet // for readability
 	// Identifier for the rack
 	ID int `json:"id"`
 
+	// Revision is a version identifier for this rack's specification, used to trigger controlled migrations
+	// when rack configuration changes require new StatefulSets. Change this field when making changes
+	// that cannot be applied in-place, such as storage updates that require pod recreation.
+	// The revision is appended to the rack ID for Kubernetes resource naming (e.g., <cluster>-<rackID>-<revision>).
```
jwalantmodi05:
```suggestion
	// The revision is appended to the rack ID for Kubernetes resource naming (e.g., <cluster-name>-<rackID>-<revision>).
```

internal/controller/cluster/statefulset.go
```diff
@@ -1065,24 +1066,24 @@ func updateSTSContainers(
 }
 
 func (r *SingleClusterReconciler) waitForAllSTSToBeReady(ignorablePodNames sets.Set[string]) error {
-	r.Log.Info("Waiting for cluster to be ready")
+	r.Log.Info("Waiting for cluster STS to be ready")
```
jwalantmodi05:
```suggestion
	r.Log.Info("Waiting for all cluster STSs to be ready")
```

internal/controller/cluster/statefulset.go
```diff
@@ -1065,24 +1066,24 @@ func updateSTSContainers(
...
 
-	allRackIDs := sets.NewInt()
+	allRackIdentifier := sets.NewString()
```
jwalantmodi05: Can we call this allRackIdentifier var as allRackIdentifiers (plural form)

pkg/utils/pod.go
```diff
@@ -123,22 +123,43 @@ func IsPodTerminating(pod *corev1.Pod) bool {
...
-func GetRackIDFromPodName(podName string) (*int, error) {
-	parts := strings.Split(podName, "-")
-	if len(parts) < 3 {
-		return nil, fmt.Errorf("failed to get rackID from podName %s", podName)
+// GetRackIDAndRevisionFromPodName returns the rack id and revision from a given pod name.
+func GetRackIDAndRevisionFromPodName(clusterName, podName string) (rackID int, rackRevision string, err error) {
+	prefix := clusterName + "-"
+
+	rackAndPodIndexPart := strings.TrimPrefix(podName, prefix)
+	// parts contain only the rack-id, rack-revision (optional) pod-index.
```
jwalantmodi05:
```suggestion
	// parts contain only the rack-id, rack-revision (optional), and pod-index.
```

internal/controller/cluster/rack.go
```diff
@@ -27,6 +26,11 @@ type scaledDownRack struct {
 	rackState *RackState
 }
 
+type RenamedRack struct {
```
tanmayja: we don't have any name for the rack in CR. saying it renamed doesn't seem right. Can we think of something else, may be around version?

internal/controller/cluster/rack.go
```diff
@@ -1980,3 +1942,234 @@ func chunkBy[T any](items []*T, chunkSize int) (chunks [][]*T) {
...
+			"Old rack STS not found. Ensuring new rack is at the target size.",
+			"oldStsName", utils.GetNamespacedNameForSTSOrConfigMap(r.aeroCluster,
+				utils.GetRackIdentifier(oldRack.Rack.ID, oldRack.Rack.Revision),
+			),
+		)
+
+		oldSts = nil
+	}
+
+	newSts, err := r.getSTS(newRack)
```
tanmayja, line 1971: any issue if we get new sts before old sts above 1958?
If we do so, then  if oldSts == nil  logic will be in continuation with oldsts code.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -502,11 +510,34 @@ func validateRackUpdate(
...
 
-					// Volume storage update is not allowed but cascadeDelete policy is allowed
-					if err := validateStorageSpecChange(&oldStorage, &newStorage); err != nil {
-						return fmt.Errorf(
-							"rack storage config cannot be updated: %v", err,
-						)
+					// Storage update is allowed only when rack revision is changed.
+					// In case of the same rack revision, check for storage update
+					if oldRack.Revision == newRack.Revision {
+						if err := validateStorageSpecChange(&oldStorage, &newStorage); err != nil {
```
tanmayja, line 570: What will be the flow when user revert just the revision just after updating the revision and storage spec in previous edit?
eg: 1- revision x -> y, storage1 -> storage2
2- revision y -> x.       soon after running step 1

abhishekdwivedi3060: This is already handled in current validation.

internal/controller/cluster/rack.go
abhishekdwivedi3060: Changed it to `revisionChangedRack`

jwalantmodi05 (approved):

test/cluster/rack_revision_test.go
```diff
@@ -0,0 +1,486 @@
+package cluster
+
+import (
+	goctx "context"
+	"fmt"
+	"time"
+
+	lib "github.com/aerospike/aerospike-management-lib"
```
tanmayja: check ordering

test/cluster/rack_revision_test.go
```diff
@@ -0,0 +1,486 @@
...
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/pkg/utils"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test"
+)
+
+var _ = Describe(
+	"Revision", func() {
```
tanmayja: make it RackRevision

test/cluster/rack_revision_test.go
```diff
@@ -0,0 +1,486 @@
...
+								By("Waiting for the migration to complete")
+								err = waitForAerospikeCluster(
+									k8sClient, ctx, aeroCluster, int(aeroCluster.Spec.Size), retryInterval,
+									getTimeout(aeroCluster.Spec.Size),
+									[]asdbv1.AerospikeClusterPhase{asdbv1.AerospikeClusterCompleted},
+								)
+								Expect(err).ToNot(HaveOccurred())
+
+								// Validate final state
+								err = validateRackEnabledCluster(k8sClient, ctx, clusterNamespacedName)
```
tanmayja: where are we validation whether updated volume is visible in pod after successful operation?

test/cluster/rack_revision_test.go
```diff
@@ -0,0 +1,486 @@
...
+										Namespace: clusterNamespacedName.Namespace,
+									},
+								}
+								Expect(DeleteCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+								Expect(CleanupPVC(k8sClient, aeroCluster.Namespace, aeroCluster.Name)).ToNot(HaveOccurred())
+							},
+						)
+
+						It(
+							"Should successfully migrate rack to new revision with storage update", func() {
```
tanmayja: can we have a test to update the storage with no revision initially to some revision. This will test vertical scaling on existing clusters.

test/cluster/rack_revision_test.go
```diff
@@ -0,0 +1,486 @@
...
+
+		if err := k8sClient.Get(ctx, nsNm, cm); err == nil {
+			return fmt.Errorf("ConfigMap with revision %s still exists after cleanup: %s", cleanedRevision, nsNm.Name)
+		}
+	}
+
+	return nil
+}
+
+func deleteOldStatefulSet(
```
tanmayja: you can rename it more generic, as you are passing which revision you want to delete. No need to mention Old in name.

internal/controller/cluster/pod.go
```diff
@@ -950,14 +960,44 @@ func (r *SingleClusterReconciler) getIgnorablePods(racksToDelete []asdbv1.Rack,
...
+	for rackID, revisionChangedRackInfo := range revisionChangedRacks {
+		oldRackState := revisionChangedRackInfo.oldRack
+		r.Log.Info("Checking old revision failed pods for revision-changed rack",
+			"rackID", rackID, "oldRevision", oldRackState.Rack.Revision)
+
+		oldPodList, err := r.getRackPodList(oldRackState.Rack.ID, oldRackState.Rack.Revision)
+		if err != nil {
+			r.Log.Error(err, "Failed to list pods for old revision rack",
+				"rackID", oldRackState.Rack.ID, "oldRevision", oldRackState.Rack.Revision)
+			continue
```
tanmayja: check why we are not returning error here.

test/cluster/rack_revision_test.go
abhishekdwivedi3060: Added storage update check

jwalantmodi05 (approved):

PR #410, opened 2025-10-06
fix: Fix race condition in updating ACL status of Aerospike CR [KO-457]

tanmayja:


test/cluster/access_control_test.go
```diff
@@ -1974,6 +1974,154 @@ var _ = Describe(
...
+									},
+
+									{
+										Name:       "userToDrop1",
+										SecretName: test.AuthSecretName,
+										Roles: []string{
+											"profiler",
+										},
+									},
+								}
```
abhishekdwivedi3060: We can use the slicing to remove the last user to avoid code duplication

PR #413, opened 2025-10-17
feat: Handle pod eviction api to safely evict the pod. [KO-456]

tanmayja:
This PR includes:

1. Safe Pod Eviction Handling — Introduces a webhook to intercept Aerospike Pod eviction API calls (e.g., during node drain) and safely manage Pod deletions by handling Aerospike node operations such as quiesce, cluster stability, and migrations. Controlled via the environment variable ENABLE_SAFE_POD_EVICTION (default: false).
2. Webhook Test Flow in GitHub Actions — Adds a framework to run webhook-related tests as part of the GitHub Actions pipeline using envtest. 
3. Scoped Webhook Configuration — Updates webhook configurations installed by Helm to watch resources only within the WATCH_NAMESPACE.

### DIFF
--- a/test/deploy-test-operator.sh
+++ b/test/deploy-test-operator.sh
@@ -49,7 +49,7 @@ done
 operator-sdk run bundle "$BUNDLE_IMG"  --namespace=test --install-mode MultiNamespace=$(echo "$namespaces" | tr " " ",") --timeout=10m0s
 
 echo "Enabling safe pod eviction in operator by patching the subscription"
-kubectl patch subscription $(kubectl get subscriptions -n operators -o jsonpath='{.items[?(@.metadata.name contains "aerospike")].metadata.name}') -n operators --type='merge' -p '{"spec":{"config":{"env":[{"name":"ENABLE_SAFE_POD_EVICTION","value":"true"}]}}}'
+kubectl patch subscription $(kubectl get subscriptions -n test -o jsonpath='{.items[?(@.metadata.name contains "aerospike")].metadata.name}') -n test --type='merge' -p '{"spec":{"config":{"env":[{"name":"ENABLE_SAFE_POD_EVICTION","value":"true"}]}}}'
 
 for namespace in $namespaces; do
   ATTEMPT=0


helm-charts/aerospike-kubernetes-operator/templates/aerospikecluster-clusterrole.yaml
```diff
@@ -31,3 +31,4 @@ rules:
     - get
     - list
     - update
+    - patch
```
abhishekdwivedi3060: Not required

internal/webhook/general/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+		log.V(1).Info("Allowing eviction of non-Aerospike pod", "pod", eviction.Name)
+		return nil // Allow eviction
+	}
+
+	// Block Aerospike pod eviction
+	log.Info("Blocking eviction of Aerospike pod", "pod", eviction.Name)
+
+	// Set annotation asynchronously (non-blocking)
+	// TODO: do we really want async here?
+	go ew.setEvictionBlockedAnnotationAsync(pod)
```
abhishekdwivedi3060: Doing async operation is not safe:
1. Not handling the error encountered if any
2. Not waiting for the operation to complete, can fail
3. Not achieving much here

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+	webhookServer := mgr.GetWebhookServer()
+	webhookServer.Register("/validate-eviction", http.HandlerFunc(ew.Handle))
+
+	return nil
+}
+
+//nolint:lll // for readability
+// +kubebuilder:webhook:path=/validate-eviction,mutating=false,failurePolicy=ignore,sideEffects=None,groups="",resources=pods/eviction, verbs=create,versions=v1,name=veviction.kb.io,admissionReviewVersions={v1}
+
+func (ew *EvictionWebhook) Handle(w http.ResponseWriter, r *http.Request) {
```
abhishekdwivedi3060, line 120: I think it would be better if we can add some metric around this like number of eviction calls blocked etc

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	// Create base response
+	response := admissionv1.AdmissionResponse{
+		UID:     admissionReview.Request.UID,
+		Allowed: true,
+	}
+
+	// Check if webhook is enabled
+	if !ew.isWebhookEnabled() {
```
abhishekdwivedi3060, line 135: Also, instead of reading the env every time for every request, can we read this once at startup and set the intended behavior in the EvictionWebhook struct?

.github/workflows/webhook-tests.yml
```diff
@@ -0,0 +1,32 @@
+name: Eviction Webhook Test Suite
+
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+  pull_request:
+    branches:
```
abhishekdwivedi3060, line 10: Shall we add a conditional test run condition for code file path changes?

internal/controller/cluster/pod.go
```diff
@@ -1746,6 +1760,31 @@ func (r *SingleClusterReconciler) podsToRestart() (quickRestarts, podRestarts se
...
 }
 
+// getEvictionBlockedPods returns pods that have eviction-blocked annotation and need restart
+func (r *SingleClusterReconciler) getEvictionBlockedPods() sets.Set[string] {
+	evictionBlockedPods := make(sets.Set[string])
+
+	// List all pods in the cluster namespace
+	pods, err := r.getClusterPodList()
+	if err != nil {
+		r.Log.Error(err, "Failed to list pods for eviction-blocked check")
```
abhishekdwivedi3060, line 1850: Any reason for not returning error from here?
In case of any failure, there won't be any requeue and subsequent pod eviction calls for this pod will not trigger any reconcile event as eviction-blocked annotation would already be present.

internal/webhook/general/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+		pod.Annotations = make(map[string]string)
+	}
+
+	pod.Annotations[EvictionBlockedAnnotation] = time.Now().Format(time.RFC3339)
+
+	return ew.Client.Patch(ctx, pod, patch)
+}
+
+// SetupEvictionWebhookWithManager registers the eviction webhook with the manager
+func SetupEvictionWebhookWithManager(mgr ctrl.Manager) error {
```
abhishekdwivedi3060: We can remove the error return type. There is no error condition in the func

internal/webhook/general/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+// getPodForEviction retrieves the pod that is being evicted
+func (ew *EvictionWebhook) getPodForEviction(eviction *policyv1.Eviction, namespace string) (*corev1.Pod, error) {
+	podKey := types.NamespacedName{
+		Name:      eviction.Name,
+		Namespace: namespace,
+	}
+
+	pod := &corev1.Pod{}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
abhishekdwivedi3060: Instead of defining a new ctx, we can use the main context from the request `r`

internal/webhook/general/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+	if err := ew.Client.Get(ctx, podKey, pod); err != nil {
+		return nil, fmt.Errorf("failed to get pod %s: %w", podKey.String(), err)
+	}
+
+	return pod, nil
+}
+
+// setEvictionBlockedAnnotationAsync sets the eviction blocked annotation asynchronously
+func (ew *EvictionWebhook) setEvictionBlockedAnnotationAsync(pod *corev1.Pod) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
abhishekdwivedi3060: We are defining these kind of timeout at multiple places. With this approach, eviction blocking will not work in slow clusters (API server is running slow) as these requests will timeout even though they are within webhooks timeout range. We need to re-think the overall approach

internal/webhook/general/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+	// Set annotation asynchronously (non-blocking)
+	// TODO: do we really want async here?
+	go ew.setEvictionBlockedAnnotationAsync(pod)
+
+	return &admissionv1.AdmissionResponse{
+		UID:     admissionReview.Request.UID,
+		Allowed: false,
+		Result: &metav1.Status{
+			Status: metav1.StatusFailure,
+			Message: fmt.Sprintf("Eviction of Aerospike pod %s/%s is blocked by admission webhook",
```
abhishekdwivedi3060: Can we make it more descriptive with more details like AKO will take care of pod migration?

internal/webhook/general/eviction_webhook.go
```diff
@@ -0,0 +1,286 @@
...
+		log.Error(err, "Failed to parse eviction object")
+
+		return &admissionv1.AdmissionResponse{
+			UID:     admissionReview.Request.UID,
+			Allowed: false,
+			Result: &metav1.Status{
+				Status:  metav1.StatusFailure,
+				Message: fmt.Sprintf("Failed to parse eviction object: %v", err),
+				Code:    http.StatusBadRequest,
+			},
```
abhishekdwivedi3060: Can move a func called `getDeniedResponse` or something similar

internal/webhook/general/eviction_webhook.go
abhishekdwivedi3060: I think we should deny here as well in case of failure while setting annotation

internal/controller/cluster/aerospikecluster_controller.go
```diff
@@ -37,6 +40,52 @@ type AerospikeClusterReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager
 func (r *AerospikeClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Create predicate for pods with eviction-blocked annotation
+	podPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Check if the annotation was added in this update
```
jwalantmodi05:
```suggestion
			// Check if the eviction-blocked annotation was added in this update
```

internal/controller/cluster/aerospikecluster_controller.go
```diff
@@ -37,6 +40,52 @@ type AerospikeClusterReconciler struct {
...
+			if !ok {
+				return false
+			}
+
+			newPod, ok := e.ObjectNew.(*corev1.Pod)
+			if !ok {
+				return false
+			}
+
+			// Check if eviction-blocked annotation was added
```
jwalantmodi05: this comment not required, as it already mentioned above.

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+// isWebhookEnabled checks if the eviction webhook is enabled
+func (ew *EvictionWebhook) isWebhookEnabled() bool {
+	return ew.Enable
+}
+
+// recordMetric records an eviction webhook metric if the webhook is enabled
+func (ew *EvictionWebhook) recordMetric(namespace, decision string) {
+	if ew.Enable {
```
jwalantmodi05:
```suggestion
	if isWebhookEnabled() {
```

test/cluster/eviction_webhook_test.go
```diff
@@ -0,0 +1,233 @@
...
+						Namespace: namespace,
+					},
+				}
+
+				Expect(DeleteCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+				Expect(CleanupPVC(k8sClient, aeroCluster.Namespace, aeroCluster.Name)).ToNot(HaveOccurred())
+			},
+		)
+
+		Context(
```
jwalantmodi05, line 46: Add one more test case to verify the combined behavior of the eviction logic when both Aerospike and non-Aerospike pods are present.

test/envtests/eviction_webhook_test.go
```diff
@@ -0,0 +1,181 @@
...
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
+)
+
+var _ = Describe("Pod eviction webhook", func() {
```
jwalantmodi05, line 18: why we added this separate envtests?

test/envtests/eviction_webhook_test.go
```diff
@@ -0,0 +1,181 @@
...
+	})
+
+	Context("When webhook is enabled", func() {
+		BeforeEach(func() {
+			evictionWebhook.Enable = true
+		})
+
+		Context("Aerospike pods", func() {
+			It("should block eviction of Aerospike pod", func() {
+				evictionWebhook.Enable = true
```
jwalantmodi05: not required, as mentioned in BeforeEach.

test/envtests/eviction_webhook_test.go
```diff
@@ -0,0 +1,181 @@
...
+					ObjectMeta:    metav1.ObjectMeta{Name: podName, Namespace: testNs},
+					DeleteOptions: &metav1.DeleteOptions{},
+				}
+
+				By("attempting eviction")
+				err := clientSet.CoreV1().Pods(testNs).EvictV1(ctx, eviction)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("blocked by admission webhook"))
+
+				By("verifying annotation is set")
```
jwalantmodi05:
```suggestion
				By("verifying eviction-blocked annotation is set")
```

test/envtests/eviction_webhook_test.go
```diff
@@ -0,0 +1,181 @@
...
+	})
+
+	Context("When webhook is disabled", func() {
+		BeforeEach(func() {
+			evictionWebhook.Enable = false
+		})
+
+		It("should allow eviction of a pod", func() {
+			// Create a simple pod
+			pod := getDummyPodWithAerospikeLabels(podName, testNs)
```
jwalantmodi05: shall we put this, getDummyPodWithAerospikeLabels func in BeforeEach? so, we don't need to call this func in every It().

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,352 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
+)
+
+const (
+	// EvictionBlockedAnnotation is the annotation set on pods when eviction is blocked
+	EvictionBlockedAnnotation = "aerospike.com/eviction-blocked"
```
tanmayja: can we have just 1 defination for this constant, 1 is already in utils.go

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,352 @@
...
+// as it is used only for temporary operations and does not need to be deeply copied.
+type EvictionWebhook struct {
+	Client    client.Client // Cache-backed client (fast for Aerospike pods)
+	APIReader client.Reader // Direct API server reader (fallback for non-Aerospike pods)
+	Log       logr.Logger
+	Enable    bool
+}
+
+// isAerospikePod checks if the given pod is an Aerospike pod
+func (ew *EvictionWebhook) isAerospikePod(pod *corev1.Pod) bool {
```
tanmayja: why this function has to be under (ew *EvictionWebhook)?
can we make it generic?

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,352 @@
...
+	// Check for Aerospike-specific labels
+	appLabel, hasAppLabel := labels[asdbv1.AerospikeAppLabel]
+	_, hasCustomResourceLabel := labels[asdbv1.AerospikeCustomResourceLabel]
+
+	// Pod is considered an Aerospike pod if it has both required labels
+	return hasAppLabel && appLabel == asdbv1.AerospikeAppLabelValue && hasCustomResourceLabel
+}
+
+// setEvictionBlockedAnnotation sets an annotation on the pod indicating eviction was blocked
+func (ew *EvictionWebhook) setEvictionBlockedAnnotation(ctx context.Context, pod *corev1.Pod) error {
```
tanmayja, line 72: Suggestion:
Similarly we can have a generic function of setting any annotation to a pod, by passing the key/value.

internal/webhook/eviction/eviction_webhook.go
```diff
@@ -0,0 +1,352 @@
...
+
+// processEvictionRequest processes the eviction request and returns the response
+func (ew *EvictionWebhook) processEvictionRequest(ctx context.Context, admissionReview *admissionv1.AdmissionReview,
+	log logr.Logger) *admissionv1.AdmissionResponse {
+	// Parse eviction object
+	eviction, err := ew.parseEvictionObject(admissionReview.Request.Object.Raw)
+	if err != nil {
+		log.Error(err, "Failed to parse eviction object")
+
+		return getFailureResponse(
```
tanmayja, line 184: have we decided to return failure response in all places, irrespective of aerospike pod or non-aerospike pod?

test/envtests/eviction_webhook_test.go
abhishekdwivedi3060: To test the enable and disable behaviour of eviction webhook. The same is not possible or complex with our standard OLM based AKO deployment testing

test/cluster/eviction_webhook_test.go
abhishekdwivedi3060: We are already testing both aerospike and non-aerospike individually. I don't see any advantage of adding a separate test-case to test both pods together. 
Do you see a test scenario that requires testing both of them together?

internal/webhook/eviction/eviction_webhook.go
abhishekdwivedi3060: Yes, due to data safety reason. Better to block than allow unsafe operation.

internal/webhook/eviction/eviction_webhook.go
abhishekdwivedi3060: To make it truly generic, we need to compare both the key and value to avoid redundant update calls. But in case of eviction annotation we can't compare the value because its a time value. It'll be different in every comparison even if the key is same.
Leaving it as it for now.

jwalantmodi05 (approved):

Issue #400, opened 2025-09-01
AKO pod reconciliation loop interferes with cluster autoscaling


rodrigorfk:
## Problem
The Aerospike Kubernetes Operator (AKO) repeatedly deletes and recreates pods flagged as `Unschedulable` by the Kubernetes scheduler. This behavior prevents cluster autoscalers, such as Karpenter, from provisioning new nodes, leading to an infinite loop of pod creation and termination.

## Context
In a cost-efficient cluster design, nodes are provisioned on demand rather than pre-provisioned as warm spare capacity. Cluster autoscalers like Karpenter or the Cluster Autoscaler are responsible for adding new nodes when they detect pods in a pending state. Karpenter, for example, typically provisions new nodes only after the Kubernetes scheduler marks a pod as `Unschedulable`, allowing a configurable [--batch-idle-duration](https://karpenter.sh/docs/reference/settings/#:~:text=%2D%2Dbatch%2Didle%2Dduration) to consolidate multiple pending pods before acting.

## Current Behavior
The AKO's reconciliation logic proactively deletes Aerospike pods almost immediately after the Kubernetes scheduler marks them as `Unschedulable`. This happens before a cluster autoscaler has sufficient time to observe the pending pod and provision new capacity. As a result, the autoscaler never sees a stable `Unschedulable` pod to trigger a scale-up event. This creates a destructive feedback loop where AKO continuously deletes pods, which are then immediately recreated by the StatefulSet controller, only to be deleted again by AKO.

## Details
When creating a new Aerospike clusters, pods are created and deleted endlessly as following:
```
NAME               READY   STATUS    RESTARTS   AGE
firstcluster-0-0   0/1     Pending   0          0s
firstcluster-0-0   0/1     Pending   0          0s
firstcluster-0-2   0/1     Pending   0          0s
firstcluster-0-1   0/1     Pending   0          0s
firstcluster-0-2   0/1     Pending   0          0s
firstcluster-0-1   0/1     Pending   0          0s
firstcluster-0-2   0/1     Terminating   0          4s
firstcluster-0-2   0/1     Terminating   0          4s
firstcluster-0-1   0/1     Terminating   0          4s
firstcluster-0-1   0/1     Terminating   0          4s
firstcluster-0-0   0/1     Terminating   0          4s
firstcluster-0-0   0/1     Terminating   0          4s
firstcluster-0-2   0/1     Pending       0          0s
[...]

PR #415, opened 2025-10-23
feat: Add support for grace-period for failed/pending pods [KO-455]

abhishekdwivedi3060:
Fixes: https://github.com/aerospike/aerospike-kubernetes-operator/issues/400

- Added Configurable Grace Period for failed/pending pod handling. It is set via env `FAILED_POD_GRACE_PERIOD_SECONDS`
- Added integration and unit test-cases
- Added github actions for unit test-cases
- Added testing readme for testing overview

### DIFF
--- a/api/v1/access_control_validate.go
+++ b/api/v1/access_control_validate.go
@@ -89,11 +89,6 @@ func IsAerospikeAccessControlValid(aerospikeClusterSpec *AerospikeClusterSpec) (
 		return false, err
 	}
 
-	if !enabled && aerospikeClusterSpec.AerospikeAccessControl != nil {
-		// Security is disabled however access control is specified.
-		return false, fmt.Errorf("security is disabled but access control is specified")
-	}
-
 	if !enabled {
 		return true, nil
 	}
--- a/internal/controller/cluster/access_control.go
+++ b/internal/controller/cluster/access_control.go
@@ -37,26 +37,8 @@ func AerospikeAdminCredentials(
 	desiredState, currentState *asdbv1.AerospikeClusterSpec,
 	passwordProvider AerospikeUserPasswordProvider,
 ) (user, pass string, err error) {
-	var (
-		enabled            bool
-		currentSecurityErr error
-		desiredSecurityErr error
-	)
-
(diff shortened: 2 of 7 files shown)


internal/controller/cluster/pod.go
```diff
@@ -548,22 +557,30 @@ func (r *SingleClusterReconciler) ensurePodsRunningAndReady(podsToCheck []*corev
...
-func getFailedAndActivePods(pods []*corev1.Pod) (failedPods, activePods []*corev1.Pod) {
+func getFailedAndActivePods(
+	pods []*corev1.Pod, withGracePeriod bool) (failedPods, failedWithinGracePeriodPods, activePods []*corev1.Pod,
+) {
 	for idx := range pods {
 		pod := pods[idx]
 
-		if err := utils.CheckPodFailed(pod); err != nil {
-			failedPods = append(failedPods, pod)
+		inGracePeriod, err := utils.CheckPodFailedWithGrace(pod, withGracePeriod)
```
tanmayja: The error returned by a function typically indicates that something went wrong during its execution. However, in this case, the function is returning information through the err variable even though it’s not an actual failure. We should check if there’s a better way to handle or represent this.

tanmayja (commented):
Should we impose any limits on the grace period that users can configure? Could an extremely large or small grace period affect AKO’s internal timeouts, especially if we have hardcoded values in certain places?

internal/controller/cluster/reconciler.go
```diff
@@ -590,37 +608,58 @@ func (r *SingleClusterReconciler) isNewCluster() (bool, error) {
...
+	}
+
+	// If we reach here, all pods are failed
+	if inGracePeriod {
+		// Return grace period state
+		return failed, inGracePeriod, nil
+	}
+
+	if r.IsStatusEmpty() {
+		return true, inGracePeriod, nil
```
tanmayja, line 659: inGracePeriod will always be false here, right?
If so, let’s return false explicitly for better clarity, similar to what we did on line 662.

abhishekdwivedi3060: Returning the vars everywhere instead of explicit values

jwalantmodi05 (approved):

PR #417, opened 2025-11-10
feat: Add support for ABS v3.4.0 [KO-460]

abhishekdwivedi3060:
- Added support for ABS v3.4.0
- Added test-cases for TLS backup and restore

### DIFF
--- a/test/cluster/rack_revision_test.go
+++ b/test/cluster/rack_revision_test.go
@@ -105,7 +105,7 @@ var _ = Describe(
 
 								Eventually(func() bool {
 									return checkBothRevisionsExist(k8sClient, ctx, clusterNamespacedName, versionV1, versionV2)
-								}, 5*time.Minute, 10*time.Second).Should(BeTrue())
+								}, 10*time.Minute, 10*time.Second).Should(BeTrue())
 
 								// Manually delete old StatefulSet
 								By("Manually deleting old StatefulSet during migration")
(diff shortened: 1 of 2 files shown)


test/cluster/rack_revision_test.go
```diff
@@ -105,7 +105,7 @@ var _ = Describe(
 
 								Eventually(func() bool {
 									return checkBothRevisionsExist(k8sClient, ctx, clusterNamespacedName, versionV1, versionV2)
-								}, 5*time.Minute, 10*time.Second).Should(BeTrue())
+								}, 10*time.Minute, 10*time.Second).Should(BeTrue())
```
jwalantmodi05, line 108: why increased timeout? and is 10s interval not high for this?

abhishekdwivedi3060: Because of 8 thread parallel test runs, it takes more than 5 minutes to satisfy this check. 
Retry interval is high to reduce the API calls during the wait period

test/backup/test_utils.go
```diff
@@ -124,6 +133,27 @@ func getBackupConfigInMap(prefix string) map[string]interface{} {
...
 }
 
+func getBackupConfigWithTLSInMap(prefix string) map[string]interface{} {
+	backupConfig := getBackupConfigInMap(prefix)
+
+	aerospikeClusterMap := backupConfig[asdbv1beta1.AerospikeClusterKey].(map[string]interface{})
+	aerospikeCluster := aerospikeClusterMap[fmt.Sprintf("%s-%s", prefix, "test-cluster")].(map[string]interface{})
+
+	aerospikeCluster["tls"] = map[string]interface{}{
+		"ca-path":   "/etc/aerospike/secret/cacerts",
```
tanmayja, line 143: Check if we have aliases for these strings

jwalantmodi05 (approved):

PR #420, opened 2025-11-27
feat: Add support for FIPS (Federal Edition) [KO-453]

jwalantmodi05:
This PR includes:

- Updated the mount logic for the work-directory volume — instead of mounting directly at /opt/aerospike, it now uses a subdirectory-based mounting approach.
- Fips related files will be added by the init container at /etc/aerospike. [Here](https://github.com/aerospike/aerospike-kubernetes-init/pull/59) is PR for the same.
- Introduced support for generic PKI-based authentication.
- Added admin TLS certificates (CN='admin') for PKI-based authentication.
- Added validations.

### DIFF
--- a/api/v1/access_control_validate.go
+++ b/api/v1/access_control_validate.go
@@ -110,7 +110,7 @@ func IsAerospikeAccessControlValid(aerospikeClusterSpec *AerospikeClusterSpec) (
 
 	// Validate users.
 	_, err = isUserSpecValid(
-		aerospikeClusterSpec.AerospikeAccessControl.Users, roleMap,
+		aerospikeClusterSpec.Image, aerospikeClusterSpec.AerospikeAccessControl.Users, roleMap,
 	)
 	if err != nil {
 		return false, err
@@ -397,10 +397,11 @@ func subset(first, second []string) bool {
 
 // isUserSpecValid indicates if input user specification is valid.
 func isUserSpecValid(
-	users []AerospikeUserSpec, roles map[string]AerospikeRoleSpec,
+	image string, users []AerospikeUserSpec, roles map[string]AerospikeRoleSpec,
 ) (bool, error) {
 	requiredRolesUserFound := false
 	seenUsers := map[string]bool{}
+	isFederal := IsFederal(image)
 
 	for _, userSpec := range users {
 		_, isSeen := seenUsers[userSpec.Name]
@@ -456,6 +457,10 @@ func isUserSpecValid(
 				return false, fmt.Errorf("user %s cannot set secretName when authMode is PKIOnly", userSpec.Name)
 			}
 		default:
+			if isFederal {
+				return false, fmt.Errorf("authMode for all users must be PKI with Federal Edition")
+			}
+
 			// TODO: We should validate actual password here but we cannot read the secret here.
 			// Will have to be done at the time of creating the user!
 			// Internal or empty authMode requires a secret for password
--- a/internal/webhook/v1/aerospikecluster_validating_webhook.go
+++ b/internal/webhook/v1/aerospikecluster_validating_webhook.go
@@ -750,18 +750,7 @@ func validateAccessControl(_ logr.Logger, cluster *asdbv1.AerospikeCluster) erro
 		return err
 	}
 
-	if asdbv1.IsFederal(cluster.Spec.Image) {
-		users := asdbv1.GetUsersFromSpec(&cluster.Spec)
-		for _, user := range users {
-			if user.AuthMode != asdbv1.AerospikeAuthModePKIOnly {
-				return fmt.Errorf("authMode for all users must be PKI with Federal Edition")
-			}
-		}
-	} else {
-		return validatePKIAuthSupportForEE(&cluster.Spec)
-	}
-
-	return nil
+	return validatePKIAuthSupportForEE(&cluster.Spec)
 }
 
 func validatePodSpecResourceAndLimits(_ logr.Logger, cluster *asdbv1.AerospikeCluster) error {
@@ -1945,20 +1934,11 @@ func validateUsersAuthModeUpdate(oldUsers, newUsers []asdbv1.AerospikeUserSpec) 
 }
 
 func validatePKIAuthSupportForEE(spec *asdbv1.AerospikeClusterSpec) error {
-	if spec.AerospikeAccessControl == nil {
+	if !isEnterprise(spec.Image) {
 		return nil
 	}
 
-	requiresPKIOnlyAuth := false
-
-	for _, userSpec := range spec.AerospikeAccessControl.Users {
-		if userSpec.AuthMode == asdbv1.AerospikeAuthModePKIOnly {
-			requiresPKIOnlyAuth = true
-			break
-		}
-	}
-
-	if !requiresPKIOnlyAuth {
+	if !hasPKIOnlyUser(spec) {
 		return nil
 	}
 


api/v1/aerospikecluster_types.go
```diff
@@ -58,6 +58,17 @@ const (
...
 
+// +kubebuilder:validation:Enum=Internal;PKI
+type AerospikeAuthMode string
+
+const (
+	// AerospikeAuthModeInternal indicates that Internal auth mode is enabled for the user.
+	AerospikeAuthModeInternal AerospikeAuthMode = "Internal"
+	// AerospikeAuthModePKIOnly indicates that PKI auth mode is enabled for the user.
+	// once set, it is not allowed to reverse back to Internal auth mode.
+	AerospikeAuthModePKIOnly AerospikeAuthMode = "PKI"
```
abhishekdwivedi3060: Lets make it "PKIOnly"

api/v1/aerospikecluster_types.go
```diff
@@ -578,6 +589,11 @@ type AerospikeRoleSpec struct {
 
 // AerospikeUserSpec specifies an Aerospike database user, the secret name for the password and, associated roles.
 type AerospikeUserSpec struct {
+	// AerospikeAuthMode specifies an authentication mode (Internal or PKI) enabled for the user.
+	// +kubebuilder:default:=Internal
+	// +optional
+	AerospikeAuthMode AerospikeAuthMode `json:"aerospikeAuthMode,omitempty"`
```
abhishekdwivedi3060: Lets keep the field name as `authMode` to align with other tool's naming convention `auth-mode`

config/samples/secrets/admin_key.pem
abhishekdwivedi3060, line 1: these files are not related to just federal. They are applicable even with Enterprise with PKI support. Rename all the admin related TLS files

internal/webhook/v1/utils.go
```diff
@@ -6,6 +6,7 @@ const (
 	baseVersion                  = "6.0.0.0"
 	baseInitVersion              = "1.0.0"
 	minInitVersionForDynamicConf = "2.2.0"
+	minVersionForPKIOnlyAuthMode = "8.1.0.0"
```
abhishekdwivedi3060:
```suggestion
	minVersionForEnterprisePKIOnlyAuthMode = "8.1.0.0"
```

api/v1/aerospikecluster_types.go
```diff
@@ -578,12 +589,19 @@ type AerospikeRoleSpec struct {
 
 // AerospikeUserSpec specifies an Aerospike database user, the secret name for the password and, associated roles.
 type AerospikeUserSpec struct {
+	// AerospikeAuthMode specifies an authentication mode (Internal or PKIOnly) enabled for the user.
+	// +kubebuilder:default:=Internal
```
abhishekdwivedi3060: Just confirm whether this defaulting here will update the CR value, resulting in rolling restart for existing clusters when AKO is upgraded

api/v1/aerospikecluster_types.go
```diff
@@ -58,6 +58,17 @@ const (
...
 )
 
+// +kubebuilder:validation:Enum=Internal;PKIOnly
+type AerospikeAuthMode string
+
+const (
+	// AerospikeAuthModeInternal indicates that Internal auth mode is enabled for the user.
+	AerospikeAuthModeInternal AerospikeAuthMode = "Internal"
+	// AerospikeAuthModePKIOnly indicates that PKIOnly auth mode is enabled for the user.
+	// once set, it is not allowed to reverse back to Internal auth mode.
```
```suggestion
	// AerospikeAuthModePKIOnly indicates that PKIOnly auth mode is enabled for the user. This mode disables password based authentication for the user
	// Once set, it is not allowed to revert back to Internal auth mode.
```

api/v1/aerospikecluster_types.go
```diff
@@ -58,6 +58,17 @@ const (
 	Empty           DynamicConfigUpdateStatus = ""
 )
 
+// +kubebuilder:validation:Enum=Internal;PKIOnly
+type AerospikeAuthMode string
+
+const (
+	// AerospikeAuthModeInternal indicates that Internal auth mode is enabled for the user.
```
abhishekdwivedi3060, line 65:
```suggestion
	// AerospikeAuthModeInternal indicates that Internal auth mode is enabled for the user.
	// This mode allows user to use password or PKI based authentication.

```

api/v1/aerospikecluster_types.go
```diff
@@ -578,12 +589,19 @@ type AerospikeRoleSpec struct {
...
+	// +optional
+	AerospikeAuthMode AerospikeAuthMode `json:"aerospikeAuthMode,omitempty"`
+
 	// Name is the user's username.
 	Name string `json:"name"`
 
 	// SecretName has secret info created by user. User needs to create this secret from password literal.
 	// eg: kubectl create secret generic dev-db-secret --from-literal=password='password'
-	SecretName string `json:"secretName"`
+	// required when AerospikeAuthMode is Internal
```
```suggestion
     // SecretName has secret info created by user. User needs to create this secret from `password` literal.
	// It is a required field in case of Internal AerospikeAuthMode.
	// eg: kubectl create secret generic dev-db-secret --from-literal=password='password'
```

```

File: api/v1/access_control_validate.go [hunk 8]
```diff
@@ -434,9 +452,14 @@ func isUserSpecValid(
 			}
 		}
 
+		if userSpec.AerospikeAuthMode == AerospikeAuthModePKIOnly && strings.TrimSpace(userSpec.SecretName) != "" {
+			return false, fmt.Errorf(
+				"user %s doesn't require password when authmode is PKI", userSpec.Name,
```
abhishekdwivedi3060:
```suggestion
				"user %s can't set password when authMode is PKIOnly", userSpec.Name,
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1337,9 +1414,32 @@ func validateRequiredFileStorageForAerospikeConfig(
...
-// isEnterprise indicates if aerospike image is enterprise
-func isEnterprise(image string) bool {
-	return strings.Contains(strings.ToLower(image), "enterprise")
+// isEnterpriseOrFederal indicates if aerospike image is enterprise or federal
+func isEnterpriseOrFederal(image string) bool {
+	return strings.Contains(strings.ToLower(image), "enterprise") || strings.Contains(strings.ToLower(image), "federal")
+}
+
+// isFederal indicates if aerospike image is federal
+func isFederal(image string) bool {
```
abhishekdwivedi3060: Create a separate func for enterprise `isEnterprise` similar to federal.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -382,6 +402,18 @@ func validateAccessControlUpdate(
...
 
+func validateImageUpdate(oldImage, newImage string) error {
+	if !asdbv1.FederalImage(oldImage) && asdbv1.FederalImage(newImage) {
+		return fmt.Errorf("enterprise image cannot be updated to federal image")
+	}
+
+	if asdbv1.FederalImage(oldImage) && !asdbv1.FederalImage(newImage) {
+		return fmt.Errorf("federal image cannot be updated to enterprise image")
+	}
+
```
abhishekdwivedi3060: For better readability and clear separation 
```suggestion
	if asdbv1.isEnterprise(oldImage) && asdbv1.FederalImage(newImage) {
		return fmt.Errorf("Enterprise to Federal edition upgrade is not supported")
	}

	if asdbv1.FederalImage(oldImage) && asdbv1.isEnterprise(newImage) {
		return fmt.Errorf("Federal to Enterprise edition upgrade is not supported")
	}

```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1337,9 +1414,32 @@ func validateRequiredFileStorageForAerospikeConfig(
 	return nil
 }
 
-// isEnterprise indicates if aerospike image is enterprise
-func isEnterprise(image string) bool {
-	return strings.Contains(strings.ToLower(image), "enterprise")
+// isEnterpriseOrFederal indicates if aerospike image is enterprise or federal
+func isEnterpriseOrFederal(image string) bool {
+	return strings.Contains(strings.ToLower(image), "enterprise") || strings.Contains(strings.ToLower(image), "federal")
```
abhishekdwivedi3060: Instead of having similar checks here and in federal. Create a separate func for enterprise `isEnterprise` instead of this and use them wherever needed

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -133,6 +133,10 @@ func (acv *AerospikeClusterCustomValidator) ValidateUpdate(_ context.Context, ol
 		return warnings, fmt.Errorf("failed to start upgrade: %v", err)
 	}
 
+	if err := validateImageUpdate(oldObject.Spec.Image, aerospikeCluster.Spec.Image); err != nil {
```
abhishekdwivedi3060: Move this func above `IsValidUpgrade`

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1840,3 +1940,63 @@ func getMinRunningInitVersion(pods map[string]asdbv1.AerospikePodStatus) (string
...
+}
+
+func validatePKIAuthSupport(spec *asdbv1.AerospikeClusterSpec) error {
+	if asdbv1.FederalImage(spec.Image) {
+		return nil
+	}
+
+	requiresPKIOnlyAuth := false
+
+	for _, userSpec := range spec.AerospikeAccessControl.Users {
```
abhishekdwivedi3060: This may panic. Add AerospikeAccessControl nil check

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -160,6 +164,19 @@ func (acv *AerospikeClusterCustomValidator) ValidateUpdate(_ context.Context, ol
...
+		oldUsers = oldObject.Spec.AerospikeAccessControl.Users
+	}
+
+	if aerospikeCluster.Spec.AerospikeAccessControl != nil {
+		newUsers = aerospikeCluster.Spec.AerospikeAccessControl.Users
+	}
+
+	if err := validateUsersAuthModeUpdate(oldUsers, newUsers); err != nil {
+		return warnings, err
+	}
```
abhishekdwivedi3060: Shouldn't this code block be moved to `validateAccessControlUpdate` as this is all related to ACL update?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1840,3 +1940,63 @@ func getMinRunningInitVersion(pods map[string]asdbv1.AerospikePodStatus) (string
...
+	oldUsersMap := make(map[string]asdbv1.AerospikeUserSpec)
+	for _, userSpec := range oldUsers {
+		oldUsersMap[userSpec.Name] = userSpec
+	}
+
+	for _, userSpec := range newUsers {
+		if oldUserSpec, found := oldUsersMap[userSpec.Name]; found {
+			if oldUserSpec.AerospikeAuthMode == asdbv1.AerospikeAuthModePKIOnly &&
+				userSpec.AerospikeAuthMode == asdbv1.AerospikeAuthModeInternal {
+				return fmt.Errorf("user %s cannot be allowed to update auth mode from PKI to Internal", userSpec.Name)
```
```suggestion
				return fmt.Errorf("auth mode update from PKI to Internal is not allowed for user %s", userSpec.Name)
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1337,9 +1414,32 @@ func validateRequiredFileStorageForAerospikeConfig(
...
+	if isFederal(spec.Image) {
+		users := asdbv1.GetUsersFromSpec(spec)
+		for _, user := range users {
+			if user.AerospikeAuthMode != asdbv1.AerospikeAuthModePKIOnly {
+				return fmt.Errorf("AuthMode for user %s is required to be PKI with Federal image", user.Name)
+			}
+		}
+	}
+
+	return validatePKIAuthSupport(spec)
```
abhishekdwivedi3060: Here we are validating the ACL. This shouldn't be part of `validateImage` func. Instead it should be moved to validataAccessControl func

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1337,9 +1414,32 @@ func validateRequiredFileStorageForAerospikeConfig(
...
+	// Validate image type. Only enterprise and federal image allowed for now.
+	if !isEnterpriseOrFederal(spec.Image) {
+		return fmt.Errorf("CommunityEdition Cluster not supported")
+	}
+
+	if isFederal(spec.Image) {
+		users := asdbv1.GetUsersFromSpec(spec)
+		for _, user := range users {
+			if user.AerospikeAuthMode != asdbv1.AerospikeAuthModePKIOnly {
+				return fmt.Errorf("AuthMode for user %s is required to be PKI with Federal image", user.Name)
```
```suggestion
				return fmt.Errorf("authMode must be PKIOnly for user %s with Federal Edition", user.Name)
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: validatePKIAuthSupport should be in Else statement as it'll be either Enterprise or Federal.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1840,3 +1940,63 @@ func getMinRunningInitVersion(pods map[string]asdbv1.AerospikePodStatus) (string
...
+		}
+	}
+
+	return nil
+}
+
+func validatePKIAuthSupport(spec *asdbv1.AerospikeClusterSpec) error {
+	if asdbv1.FederalImage(spec.Image) {
+		return nil
+	}
```
abhishekdwivedi3060: Not required if we move this func call to `else` block

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1840,3 +1940,63 @@ func getMinRunningInitVersion(pods map[string]asdbv1.AerospikePodStatus) (string
...
+
+	val, err := lib.CompareVersions(version, minVersionForPKIOnlyAuthMode)
+	if err != nil {
+		return fmt.Errorf("failed to compare Aerospike server version: %w", err)
+	}
+
+	if val < 0 {
+		return fmt.Errorf(
+			"PKI authentication requires Aerospike server version %s or later (cluster image %s)",
```
```suggestion
			"PKIOnly authMode can be used with Enterprise Edition version %s or later (cluster image %s)",
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -487,7 +519,13 @@ func validateClientCertSpec(cluster *asdbv1.AerospikeCluster) error {
...
 
+	adminUser := asdbv1.GetAdminUserFromSpec(&cluster.Spec)
+
 	switch {
+	case reflect.DeepEqual("false", tlsAuthenticateClientConfig) && adminUser != nil &&
+		adminUser.AerospikeAuthMode == asdbv1.AerospikeAuthModePKIOnly:
+		return fmt.Errorf("tls-authenticate-client cannot be false when auth mode for admin user is PKI, " +
+			"it should be any or specific client name")
 	case reflect.DeepEqual("false", tlsAuthenticateClientConfig):
 		return nil
```
abhishekdwivedi3060: We don't need a separate case for this.  This can merged in in the existing `	case reflect.DeepEqual("false", tlsAuthenticateClientConfig):` case with some conditional statement

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -517,6 +555,45 @@ func validateClientCertSpec(cluster *asdbv1.AerospikeCluster) error {
...
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse operator client certificate: %v", err)
+	}
+
+	if cert.Subject.CommonName != "admin" {
+		return fmt.Errorf("operator client certificate CN is '%s', must be 'admin'", cert.Subject.CommonName)
+	}
```
abhishekdwivedi3060: We need to validate against `clientCertSpec.SecretCertSource` instead of CertPathInOperator

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -517,6 +555,45 @@ func validateClientCertSpec(cluster *asdbv1.AerospikeCluster) error {
...
+	if block == nil {
+		return fmt.Errorf("failed to decode PEM block from operator client cert file")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse operator client certificate: %v", err)
+	}
+
+	if cert.Subject.CommonName != "admin" {
```
abhishekdwivedi3060: Use admin constant

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -517,6 +555,45 @@ func validateClientCertSpec(cluster *asdbv1.AerospikeCluster) error {
...
+	clientCertSpec := cluster.Spec.OperatorClientCertSpec
+
+	var certBytes []byte
+
+	var err error
+
+	if clientCertSpec == nil || clientCertSpec.CertPathInOperator == nil ||
+		clientCertSpec.CertPathInOperator.ClientCertPath == "" {
+		return nil
+	}
```
abhishekdwivedi3060: We need to error out when Client certs are not given for admin in PKI mode

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -329,6 +345,10 @@ func validate(aslog logr.Logger, cluster *asdbv1.AerospikeCluster) (admission.Wa
 		return warnings, err
 	}
 
+	if err := validateAdminCerts(cluster); err != nil {
```
abhishekdwivedi3060: Again this should be moved into `validateClientCertSpec` func

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -517,6 +555,45 @@ func validateClientCertSpec(cluster *asdbv1.AerospikeCluster) error {
 	return nil
 }
 
+func validateAdminCerts(cluster *asdbv1.AerospikeCluster) error {
```
abhishekdwivedi3060: Don't we need a check for mTLS somewhere in validation? For PKIOnly auth-mode, the cluster must be mTLS enabled

api/v1/aerospikecluster_types.go
```diff
@@ -58,6 +58,17 @@ const (
 	Empty           DynamicConfigUpdateStatus = ""
 )
 
+// +kubebuilder:validation:Enum=Internal;PKIOnly
+type AerospikeAuthMode string
```
abhishekdwivedi3060, line 62: Check if we can use the already defined enum in the client-go for the auth-mode.

api/v1/access_control_validate.go
```diff
@@ -145,6 +145,24 @@ func GetUsersFromSpec(spec *AerospikeClusterSpec) map[string]AerospikeUserSpec {
 	return users
 }
 
+// FederalImage indicates if the image is a federal image.
```
tanmayja: - functions returning bool should usually read like a question.
  something like "IsFederalImage"
- Let's have "federal" and "enterprise" as constants somewhere and use.

api/v1/access_control_validate.go
```diff
@@ -145,6 +145,24 @@ func GetUsersFromSpec(spec *AerospikeClusterSpec) map[string]AerospikeUserSpec {
...
+func FederalImage(image string) bool {
+	return strings.Contains(strings.ToLower(image), "federal")
+}
+
+// GetAdminUserFromSpec returns admin user from the spec.
+func GetAdminUserFromSpec(spec *AerospikeClusterSpec) *AerospikeUserSpec {
```
tanmayja, line 149: For better clarity and decoupling, can we pass AerospikeAccessControl pointer directly instead of full spec pointer?

api/v1/access_control_validate.go
```diff
@@ -434,9 +452,14 @@ func isUserSpecValid(
...
 
+		if userSpec.AerospikeAuthMode == AerospikeAuthModePKIOnly && strings.TrimSpace(userSpec.SecretName) != "" {
+			return false, fmt.Errorf(
+				"user %s doesn't require password when authmode is PKI", userSpec.Name,
+			)
+		}
 		// TODO We should validate actual password here but we cannot read the secret here.
 		// Will have to be done at the time of creating the user!
-		if strings.TrimSpace(userSpec.SecretName) == "" {
+		if strings.TrimSpace(userSpec.SecretName) == "" && userSpec.AerospikeAuthMode != AerospikeAuthModePKIOnly {
```
tanmayja: check if we can write these 2 conditions (line 455 and 462) better, as both relies on same arguments and repeated calls like TrimSpace.

api/v1/aerospikecluster_types.go
tanmayja: +1

api/v1/aerospikecluster_types.go
tanmayja: The change may start reflecting after any CR update performed post operator upgrade; however, this needs to be confirmed.

For my understanding: if we already have an enum validation and a default value set to Internal, what problem does omitempty solve in this case?

internal/controller/cluster/client_policy.go
```diff
@@ -176,6 +176,17 @@ func (r *SingleClusterReconciler) getClientPolicy() *as.ClientPolicy {
 	policy.User = user
 	policy.Password = pass
 
+	adminUser := asdbv1.GetAdminUserFromSpec(statusToSpec)
+	if adminUser != nil {
+		if adminUser.AerospikeAuthMode == asdbv1.AerospikeAuthModePKIOnly {
+			policy.AuthMode = as.AuthModePKI
```
abhishekdwivedi3060: Can we not assign `policy.AuthMode = as.AuthModePKI` instead of checking?

internal/controller/cluster/client_policy.go
```diff
@@ -176,6 +176,17 @@ func (r *SingleClusterReconciler) getClientPolicy() *as.ClientPolicy {
 	policy.User = user
 	policy.Password = pass
 
+	adminUser := asdbv1.GetAdminUserFromSpec(statusToSpec)
```
abhishekdwivedi3060, line 183: Add a comment here

internal/controller/cluster/access_control.go
```diff
@@ -164,7 +174,7 @@ func (r *SingleClusterReconciler) reconcileUsers(
 	desired map[string]asdbv1.AerospikeUserSpec,
 	current map[string]asdbv1.AerospikeUserSpec,
 	passwordProvider AerospikeUserPasswordProvider, client *as.Client,
-	adminPolicy as.AdminPolicy,
+	adminPolicy as.AdminPolicy, image string,
```
abhishekdwivedi3060: Why do we need to pass the image as a func param?
All the spec info is already present in `r` receiver

internal/controller/cluster/access_control.go
```diff
@@ -194,14 +204,37 @@ func (r *SingleClusterReconciler) reconcileUsers(
 
 	for userName := range desired {
 		userSpec := desired[userName]
+		currUserSpec, found := current[userName]
+
+		var password *string
+
+		if (!found || currUserSpec.AerospikeAuthMode == asdbv1.AerospikeAuthModeInternal) &&
```
abhishekdwivedi3060: Add comment for this block

internal/controller/cluster/access_control.go
```diff
@@ -194,14 +204,37 @@ func (r *SingleClusterReconciler) reconcileUsers(
...
+			} else {
+				nop := "nopassword"
+				password = &nop
+			}
+		}
 
-		password, err := passwordProvider.Get(userName, &userSpec)
-		if err != nil {
-			return err
+		if userSpec.AerospikeAuthMode == asdbv1.AerospikeAuthModeInternal {
```
tanmayja: If my understanding is incorrect, please feel free to add clarifying comments.

This part seems a bit confusing to me. Is it possible that, in the case of AerospikeAuthModePKIOnly and AerospikeAuthModeInternal, the password could be overwritten? If so, is this the expected behavior? Has this flow been tested?

internal/controller/cluster/access_control.go
tanmayja: If this section of code is just about PKIOnly, then place this check first in if condition for clarity and better Short-circuiting.

userSpec.AerospikeAuthMode == asdbv1.AerospikeAuthModePKIOnly

internal/controller/cluster/access_control.go
```diff
@@ -194,14 +204,37 @@ func (r *SingleClusterReconciler) reconcileUsers(
...
+			nop, err := passwordProvider.Get(userName, &userSpec)
+			if err != nil {
+				return err
+			}
+
+			password = &nop
 		}
 
+		r.Log.Info("user: ", "userName", userName, "password", password)
```
abhishekdwivedi3060: 1. Add comment for this block
2. Simplify the overall nested block for better readability. Current implementation is not clear
3. Don't need to check the federal image in very iteration. Can be extracted outside loop

internal/controller/cluster/rack.go
```diff
@@ -1344,20 +1344,43 @@ func (r *SingleClusterReconciler) isRackStorageUpdatedInAeroCluster(
 		containerAttachments = append(containerAttachments, volume.Sidecars...)
 
 		if volume.Aerospike != nil {
-			containerAttachment := asdbv1.VolumeAttachment{
-				ContainerName: asdbv1.AerospikeServerContainerName,
-				Path:          volume.Aerospike.Path,
-			}
+			if volume.Aerospike.Path == asdbv1.GetWorkDirectory(*r.aeroCluster.Spec.AerospikeConfig) {
```
tanmayja: add comments

internal/controller/cluster/statefulset.go
```diff
@@ -821,10 +821,11 @@ func (r *SingleClusterReconciler) updateSTSNonPVStorage(
 		addVolumeMountInContainer(
 			volume.Name, initContainerAttachments,
 			st.Spec.Template.Spec.InitContainers, initContainerVolumePathPrefix,
+			r.aeroCluster.Spec.AerospikeConfig,
```
abhishekdwivedi3060: We should use the rack level aerospikeConfig, not the spec level.
Fix this wherever aerospikeConfig is passed

internal/controller/cluster/statefulset.go
```diff
@@ -1501,6 +1502,32 @@ func addVolumeMountInContainer(
 					mountPath = volumeAttachment.Path
 				}
 
+				if container.Name == asdbv1.AerospikeServerContainerName && mountPath == asdbv1.GetWorkDirectory(*aeroConf) {
```
abhishekdwivedi3060: 1. Call this GetWorkDirectory once outside the loop instead of calling in the loop
2. Now we have 3 conditions, can we convert this to switch for better readability?
3. Also try to move this subPath related code to a func `getWorkDirMounts`

internal/controller/cluster/rack.go
abhishekdwivedi3060: 1. Add comments for this flow
2. Did we test these new changes with `validationPolicy.skipWorkDirValidate` flag true?

internal/controller/cluster/statefulset.go
```diff
@@ -1501,6 +1502,32 @@ func addVolumeMountInContainer(
...
+					volumeMount1 := corev1.VolumeMount{
+						Name:             volumeName,
+						MountPath:        mountPath + "/smd",
```
abhishekdwivedi3060: Define constants for these "smd" and "usr" path

internal/controller/cluster/client_policy.go
tanmayja, line 183: If we can get this information from the spec, we may not need to check the federal image condition separately. Just wanted to confirm if this is the right thing to do here.

internal/controller/cluster/rack.go
```diff
@@ -1505,36 +1528,49 @@ func (r *SingleClusterReconciler) isVolumeAttachmentAddedOrUpdated(
...
 
-			continue
+			if (attachment.Path == asdbv1.GetWorkDirectory(*r.aeroCluster.Spec.AerospikeConfig)+"/smd" ||
+				attachment.Path == asdbv1.GetWorkDirectory(*r.aeroCluster.Spec.AerospikeConfig)+"/usr") &&
+				volumeMount.MountPath == asdbv1.GetWorkDirectory(*r.aeroCluster.Spec.AerospikeConfig) {
+				// Accept direct workdir mount as compatible for workdir sub-directories attachments
+				found = true
+				break
+			}
+
```
abhishekdwivedi3060: Did you test the upgrade path of AKO with these changes? 
This condition seems wrong. What are we trying to do here?

internal/webhook/v1/aerospikecluster_validating_webhook.go
tanmayja: 1. From our samples/secrets, which file should have common name as "admin"?
2. Can we have a sample yaml to. create a cluster using pki in our config/samples folder?

internal/controller/cluster/access_control.go
```diff
@@ -55,6 +60,11 @@ func AerospikeAdminCredentials(
 		)
 	}
 
+	// If authMode PKIOnly is set for admin user in status, return empty user and empty password.
```
tanmayja, line 63: Similarly if we check this from spec, do we need to check for federal image?

api/v1/access_control_validate.go
abhishekdwivedi3060: Let's keep this in sync with existing funcs `GetUsersFromSpec` and `GetRolesFromSpec`

api/v1/access_control_validate.go
abhishekdwivedi3060: With latest refactor, these keywords are only used once in this file

internal/webhook/v1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: I think we won't be able to validate the user CN name from the secret cert as we don't make any API calls in the webhook. Will have to rely on the user for this.

api/v1/aerospikecluster_types.go
abhishekdwivedi3060: Removing the defaulting marker for now as we don't use that marker anywhere in our code.

internal/controller/cluster/client_policy.go
```diff
@@ -176,14 +176,14 @@ func (r *SingleClusterReconciler) getClientPolicy() *as.ClientPolicy {
 	policy.User = user
 	policy.Password = pass
 
+	//
```
tanmayja: nit

api/v1/aerospikecluster_types.go
abhishekdwivedi3060: Direct enums usage from the go-client is not possible as it's int enum.

internal/controller/cluster/client_policy.go
abhishekdwivedi3060: Can't rely on spec to determine the authMode especially due to EE edition.
In case of new EE with PKI clusters, the authMode must be Internal for the first time even though the spec says PKIOnly

internal/controller/cluster/statefulset.go
```diff
@@ -1455,24 +1456,31 @@ func getFinalVolumeAttachmentsForVolume(volume *asdbv1.VolumeSpec) (
...
-			ContainerName: asdbv1.AerospikeServerContainerName,
-			Path:          volume.Aerospike.Path,
-		}
-
-		// Mount options (ReadOnly, SubPath, SubPathExpr, MountPropagation...) are only
-		// applicable and enforced for hostPath volumes in aerospike containers.
-		if volume.Source.HostPath != nil {
-			containerAttachment.AttachmentOptions = asdbv1.AttachmentOptions{
-				MountOptions: volume.Aerospike.MountOptions,
+		// Special handling for workdir mount in aerospike server container for FIPs compliance.
```
tanmayja: FIPS (all caps)

internal/controller/cluster/client_policy.go
abhishekdwivedi3060: Added comment

test/cluster/access_control_test.go
```diff
@@ -2001,7 +1990,6 @@ var _ = Describe(
 
 										{
 											Name:       "userToDrop",
-											AuthMode:   asdbv1.AerospikeAuthModeInternal,
```
tanmayja: we can remove authmode from 1973 and 1983 line also

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1989,3 +1994,22 @@ func hasPKIOnlyUser(spec *asdbv1.AerospikeClusterSpec) bool {
...
+// validateTLSAndPKIOnlyUpdate blocks upgrading to TLS and PKIOnly in a single update.
+// When enabling TLS with PKIOnly, ACL reconciliation happens during rolling restart
+// before all pods have TLS. If admin is set to PKIOnly (nopassword), the operator cannot connect
+// to non-TLS pods because PKI auth requires TLS certificates.
+func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec) error {
+	oldTLSName, _ := asdbv1.GetServiceTLSNameAndPort(oldSpec.AerospikeConfig)
+	newTLSName, _ := asdbv1.GetServiceTLSNameAndPort(newSpec.AerospikeConfig)
+
+	// Check if TLS is being enabled
+	if oldTLSName == "" && newTLSName != "" {
```
tanmayja, line 1991: check status also

test/cluster/access_control_test.go
```diff
@@ -1967,6 +1970,7 @@ var _ = Describe(
 									Users: []asdbv1.AerospikeUserSpec{
 										{
 											Name:       "admin",
+											AuthMode:   asdbv1.AerospikeAuthModeInternal,
```
abhishekdwivedi3060, line 1973: Intentionally set this authMode at 1-2 places randomly to confirm no-op when authMode is omitted vs Internal.

tanmayja (approved):

tanmayja (approved):

PR #421, opened 2025-12-02
fix: error out during update if LB service is created from outside [KO-462]

abhishekdwivedi3060:


jwalantmodi05 (dismissed):

internal/controller/cluster/service.go
```diff
@@ -177,6 +177,18 @@ func (r *SingleClusterReconciler) reconcileSTSLoadBalancerSvc() error {
 	r.Log.Info("LoadBalancer service already exist for cluster, checking for update",
 		"name", utils.NamespacedName(service.Namespace, service.Name))
 
+	if !utils.IsOwnedBy(service, r.aeroCluster) {
+		r.Log.Info(
```
tanmayja: Do we need this separate log statement? The error is returned on the next line, so it will already appear in the logs.

internal/controller/cluster/service.go
```diff
@@ -177,6 +177,18 @@ func (r *SingleClusterReconciler) reconcileSTSLoadBalancerSvc() error {
...
+			"LoadBalancer service is not created/owned by operator",
+			"name", utils.NamespacedName(service.Namespace, service.Name),
+		)
+
+		return fmt.Errorf(
```
tanmayja, line 181: Is this going to end up in a continuous loop since we’re returning an error here for an LB service that wasn’t created by the operator?

internal/controller/cluster/service.go
abhishekdwivedi3060: We are doing something similar for PDB. Kept it in sync

internal/controller/cluster/service.go
abhishekdwivedi3060: Yes, this is only for case where we need to create a LB but couldn't because there is already a service with same name but not created by us

PR #431, opened 2026-01-29
Feature: Dynamically allocate rack ID after pod scheduling [KO-385]

tanmayja:


api/v1/utils.go
```diff
@@ -84,15 +84,16 @@ const (
...
+	AerospikeInitContainerDefaultRegistryNamespace = "tanmayj10"
+	AerospikeInitContainerDefaultNameAndTag        = "aerospike-kubernetes-init:2.5.0-dev6"
 	AerospikeAppLabel                              = "app"
 	AerospikeAppLabelValue                         = "aerospike-cluster"
 	AerospikeCustomResourceLabel                   = "aerospike.com/cr"
 	AerospikeRackIDLabel                           = "aerospike.com/rack-id"
 	AerospikeRackRevisionLabel                     = "aerospike.com/rack-revision"
 	AerospikeAPIVersionLabel                       = "aerospike.com/api-version"
 	EvictionBlockedAnnotation                      = "aerospike.com/eviction-blocked"
+	EffectiveRackIDAnnotation                      = "aerospike.com/effective-rack-id"
```
abhishekdwivedi3060: Can we call it something with "override" keyword ? "effective" key word doesn't make it clear

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+		Namespace: aeroCluster.Namespace,
+	}, pod)
+	Expect(err).ToNot(HaveOccurred())
+
+	originalAnnotation := pod.Annotations[asdbv1.EffectiveRackIDAnnotation]
+
+	aerospikeConf := pod.Annotations["aerospikeConf"]
+	re := regexp.MustCompile(`^\s*rack-id\s+(.+)$`)
+
+	println(aerospikeConf)
```
abhishekdwivedi3060: Remove the println or use pkg.Logger where required

api/v1/aerospikecluster_types.go
```diff
@@ -191,6 +191,15 @@ type AerospikeClusterSpec struct { //nolint:govet // for readability
 	// +kubebuilder:validation:MaxItems:=1
 	// +optional
 	Operations []OperationSpec `json:"operations,omitempty"`
+
+	// EnableDynamicRackID enables dynamic allocation of rack IDs to pods after they get scheduled.
+	// When enabled, the operator will watch for changes to the aerospike.com/effective-rack-id annotation on pods
```
```suggestion
	// When enabled, the operator will use the rack-id from the aerospike.com/effective-rack-id annotation on pods
```

api/v1/aerospikecluster_types.go
```diff
@@ -355,6 +364,10 @@ type AerospikePodSpec struct { //nolint:govet // for readability
 	Sidecars []corev1.Container `json:"sidecars,omitempty"`
 
 	// InitContainers to add to the pods.
+	// To control the position of custom init containers relative to aerospike-init, include a placeholder
+	// container with name "aerospike-init" at the desired position. Only the name field is required for the placeholder;
+	// all other fields will be ignored and replaced with the actual aerospike-init container configuration.
+	// If no placeholder is found, all custom init containers will be placed after aerospike-init (backward compatible).
```
```suggestion
	// To control the sequence of custom init containers relative to "aerospike-init" init container , include a placeholder
	// container with name "aerospike-init" at the desired position. Only the name field is required for the placeholder;
	// all other fields will be ignored and replaced with the actual "aerospike-init" init container configuration.
	// If no placeholder is found, all custom init containers will be placed after "aerospike-init" (backward compatible).
```

api/v1/aerospikecluster_types.go
```diff
@@ -191,6 +191,15 @@ type AerospikeClusterSpec struct { //nolint:govet // for readability
...
 	Operations []OperationSpec `json:"operations,omitempty"`
+
+	// EnableDynamicRackID enables dynamic allocation of rack IDs to pods after they get scheduled.
+	// When enabled, the operator will watch for changes to the aerospike.com/effective-rack-id annotation on pods
+	// and reconcile when this annotation value changes. This allows rack IDs to be dynamically assigned to pods
+	// based on their scheduling location or other external factors.
+	// This feature requires a single rack configuration (multiple racks are not allowed when enabled).
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enable Dynamic Rack ID"
+	// +optional
+	EnableDynamicRackID *bool `json:"enableDynamicRackID,omitempty"`
```
abhishekdwivedi3060: Just thinking out loud here:
1. User might get confuse this with dynamic rack-id change via asinfo call. Can we use a prefix/suffix that says override something like `EnableRackIDOverride`?
2. Shall we add a point here that in this case, node distribution across rack is not handled by AKO, so there can be skewed racks. Use this with caution?

internal/controller/cluster/aerospikecluster_controller.go
```diff
@@ -57,10 +56,23 @@ func (r *AerospikeClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-			return !hadAnnotation && hasAnnotation
+			// Check if the eviction-blocked annotation was added in this update
+			_, hadEvictionBlocked := oldAnnotations[asdbv1.EvictionBlockedAnnotation]
+			_, hasEvictionBlocked := newAnnotations[asdbv1.EvictionBlockedAnnotation]
+			evictionBlockedAdded := !hadEvictionBlocked && hasEvictionBlocked
+
+			// Check if the effective-rack-id annotation value changed
+			// When a pod is deleted, oldAnnotations is nil. As a result, oldEffectiveRackID and
+			// newEffectiveRackID will not match, which triggers reconciliation even if the
+			// aerospike.com/effective-rack-id value itself has not actually changed.
```
abhishekdwivedi3060: I think this is not just limited to pod restart. This will happen even for fresh cluster deployment. Extra reconciles will come even the normal deployment is going on. Let's update this comment accordingly.

internal/controller/cluster/pod.go
```diff
@@ -257,6 +263,44 @@ func (r *SingleClusterReconciler) getRollingRestartTypePod(
...
+		val, err := lib.CompareVersions(version, minInitVersionForDynamicRackID)
+		if err != nil {
+			return restartType, err
+		}
+
+		if val < 0 {
+			restartType = mergeRestartType(restartType, podRestart)
+		} else {
+			restartType = mergeRestartType(restartType, quickRestart)
+		}
```
abhishekdwivedi3060: Add a comment for this init based check for future ref

internal/controller/cluster/pod.go
```diff
@@ -257,6 +263,44 @@ func (r *SingleClusterReconciler) getRollingRestartTypePod(
...
+			return restartType, err
+		}
+
+		val, err := lib.CompareVersions(version, minInitVersionForDynamicRackID)
+		if err != nil {
+			return restartType, err
+		}
+
+		if val < 0 {
+			restartType = mergeRestartType(restartType, podRestart)
```
abhishekdwivedi3060, line 293: Add a log here also for pod restart due to old init version

pkg/utils/utils.go
```diff
@@ -272,3 +278,15 @@ func IsOwnedBy(obj, parent client.Object) bool {
 
 	return false
 }
+
+// GetContainerByName finds a container by name in a slice of containers.
+// Returns the container pointer if found, nil otherwise.
+func GetContainerByName(containers []corev1.Container, name string) *corev1.Container {
```
abhishekdwivedi3060: This func is only used in the test files at the moment. Why not move it to test utils?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -300,38 +300,46 @@ func validate(aslog logr.Logger, cluster *asdbv1.AerospikeCluster) (admission.Wa
...
 	// Validate access control
-	if err := validateAccessControl(aslog, cluster); err != nil {
+	err = validateAccessControl(aslog, cluster)
+	if err != nil {
 		return warnings, err
 	}
 
 	// Validate rackConfig
-	if err := validateRackConfig(aslog, cluster); err != nil {
+	warns, err = validateRackConfig(aslog, cluster)
```
abhishekdwivedi3060: Is it possible to move this func all above rack config validation loop just to keep all rack config related validation close to each other?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -812,12 +820,35 @@ func validateResourceAndLimits(
...
+	// If EnableDynamicRackID is enabled, only single rack is allowed
+	if asdbv1.GetBool(cluster.Spec.EnableDynamicRackID) {
+		rackCount := len(cluster.Spec.RackConfig.Racks)
+		if rackCount > 1 {
+			return warnings, fmt.Errorf(
+				"enableDynamicRackID requires a single rack configuration, but %d racks are specified",
+				rackCount,
+			)
+		}
+		// rackCount == 0 means default rack (single rack), which is allowed
```
abhishekdwivedi3060: rackCount will never be 0 as mutation adds default rackId 0. So there will always be 1 rack.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -812,12 +820,35 @@ func validateResourceAndLimits(
...
+		// rackCount == 1 means one explicit rack, which is allowed
+
+		// Warn if namespaces is empty when EnableDynamicRackID is set
+		if len(cluster.Spec.RackConfig.Namespaces) == 0 {
+			warnings = append(warnings,
+				"enableDynamicRackID is set but rackConfig.namespaces is empty. "+
+					"Consider specifying namespaces for better rack management.",
```
```suggestion
					"Consider specifying namespaces for namespace rack-aware setup.",
```

internal/controller/cluster/statefulset.go
```diff
@@ -1011,60 +1011,59 @@ func (r *SingleClusterReconciler) updateSTSFromPodSpec(
 	r.updateReservedContainers(st)
 }
 
+// updateSTSContainers updates containers with custom containers from spec, maintaining order.
 func updateSTSContainers(
```
abhishekdwivedi3060, line 1015: This new flow is lil bit difficult to understand. Can we add comments for different flows?

test/cluster/override_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+								Name:      clusterName,
+								Namespace: namespace,
+							},
+						}
+						Expect(DeleteCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+						Expect(CleanupPVC(k8sClient, aeroCluster.Namespace, aeroCluster.Name)).ToNot(HaveOccurred())
+					},
+				)
+
+				It(
```
abhishekdwivedi3060, line 64: Can we add similar test-case for update flow:
- Should reject EnableDynamicRackID update in a running cluster with multiple racks

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+		ctx := goctx.Background()
+		clusterName := fmt.Sprintf("dynamic-rack-id-%d", GinkgoParallelProcess())
+		clusterNamespacedName := test.GetNamespacedName(
+			clusterName, namespace,
+		)
+
+		Context(
+			"When doing invalid operations around EnableDynamicRackID feature", func() {
+
+				AfterEach(
```
abhishekdwivedi3060: Can we move this AfterEach outside of each Context block at the Describe level?

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+	scanner := bufio.NewScanner(strings.NewReader(aerospikeConf))
+
+	var effectiveRackID string
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		println(line)
+
+		if m := re.FindStringSubmatch(line); len(m) == 2 {
+			effectiveRackID = m[1]
```
abhishekdwivedi3060: Break from here

test/cluster/override_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+								validateDynamicRackIDInConfig(ctx, aeroCluster, true)
+
+								By("Updating effective-rack-id annotation")
+								pod := &v1.Pod{
+									ObjectMeta: metav1.ObjectMeta{
+										Name:      aeroCluster.Name + "-1-0",
+										Namespace: aeroCluster.Namespace,
+									},
+								}
```
abhishekdwivedi3060, line 148: How about we define this pod's types.NsNm once in IT and use that for validateDynamicRackIDInConfig and at other places to confirm we are using the same pod everywhere

test/cluster/dynamic_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+								pod := &v1.Pod{
+									ObjectMeta: metav1.ObjectMeta{
+										Name:      aeroCluster.Name + "-1-0",
+										Namespace: aeroCluster.Namespace,
+									},
+								}
+								err = k8sClient.Delete(ctx, pod)
+								Expect(err).ToNot(HaveOccurred())
+
+								started := waitForPod(types.NamespacedName{Namespace: namespace, Name: pod.Name})
```
abhishekdwivedi3060: 1. Instead of this, shouldn't we check for waitForAerospikeCluster? This will ensure that reconcile happened and CR reached completed state.
2. For SC namespaces, we should check the roster to confirm new rackID is effective in roster

test/cluster/override_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+								Expect(started).To(BeTrue(), "Pod did not start in time")
+
+								// Note: The predicate should still trigger reconciliation,
+								// but the restart logic should not trigger based on DynamicRackID
+								// since the feature is disabled
+								validateDynamicRackIDInConfig(ctx, aeroCluster, false)
+							},
+						)
+
+						It(
```
abhishekdwivedi3060, line 221: Need to add test-cases for:
- Should allow disabling EnableDynamicRackID in a running cluster.
- Test case around init version  < 2.5.0 and EnableDynamicRackID is enabled, should result in cold restart
- Should fail when EnableDynamicRackID is enabled but no annotations in the pod metadata

test/cluster/override_rackid_test.go
```diff
@@ -0,0 +1,224 @@
...
+									clusterNamespacedName, 2,
+								)
+
+								aeroCluster.Spec.EnableDynamicRackID = ptr.To(false)
+								Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+							},
+						)
+
+						It(
+							"Should not trigger restart when annotation changes if feature is disabled", func() {
```
abhishekdwivedi3060, line 184: Pleas check if above comments are applicable here and for below test-cases as well.

test/cluster/podspec_test.go
```diff
@@ -218,66 +222,286 @@ var _ = Describe(
...
+									Expect(initContainers).To(HaveLen(3),
+										"Should have 3 init containers: custom-init-1, custom-init-2, aerospike-init")
+
+									// First should be custom-init-1
+									Expect(initContainers[0].Name).To(Equal("custom-init-1"))
+									// Second should be custom-init-2
+									Expect(initContainers[1].Name).To(Equal("custom-init-2"))
+									// Third should be aerospike-init
+									Expect(initContainers[2].Name).To(Equal(asdbv1.AerospikeInitContainerName))
+								}
```
abhishekdwivedi3060: I think we can merge this in above test-case

test/cluster/podspec_test.go
```diff
@@ -218,66 +222,286 @@ var _ = Describe(
...
+								Expect(err).ToNot(HaveOccurred())
+								for _, sts := range stsList.Items {
+									initContainers := sts.Spec.Template.Spec.InitContainers
+									Expect(initContainers).To(HaveLen(3))
+									Expect(initContainers[0].Name).To(Equal(asdbv1.AerospikeInitContainerName))
+									Expect(initContainers[1].Name).To(Equal("custom-init-1"))
+									Expect(initContainers[2].Name).To(Equal("custom-init-2"))
+								}
+							},
+						)
```
abhishekdwivedi3060, line 482: This again seems like a duplicate test. What extra are we testing here?

test/cluster/podspec_test.go
```diff
@@ -49,19 +51,17 @@ var _ = Describe(
...
 
-		// initCont2 := corev1.Container{
-		// 	Name:    "init-mydb",
-		// 	Image:   "busybox:1.28",
-		// 	Command: []string{"sh", "-c",
-		//	"until nslookup mydb.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local;
-		//	do echo waiting for mydb; sleep 2; done"},
-		// }
+		customInit2 := corev1.Container{
+			Name:    "custom-init-2",
```
abhishekdwivedi3060: Lets create a constant for these names and use in the following test-cases

test/cluster/podspec_test.go
tanmayja: This tests adding custom init continers and their ordering, previous tests are playing around existing init containers and their ordering.

api/v1/aerospikecluster_types.go
```diff
@@ -191,6 +191,16 @@ type AerospikeClusterSpec struct { //nolint:govet // for readability
 	// +kubebuilder:validation:MaxItems:=1
 	// +optional
 	Operations []OperationSpec `json:"operations,omitempty"`
+
+	// EnableRackIDOverride enables dynamic allocation of rack IDs to pods after they get scheduled.
+	// When enabled, the operator will watch for changes to the aerospike.com/override-rack-id annotation on pods
+	// and reconcile when this annotation value changes. This allows rack IDs to be dynamically assigned to pods
+	// based on their scheduling location or other external factors.
+	// AKO does not manage node distribution across racks, which may result in uneven rack placement. Use with caution.
```
```suggestion
	// AKO does not manage pod distribution across racks, which may result in skewed rack. Use with caution.
```

internal/controller/cluster/pod.go
```diff
@@ -263,31 +263,42 @@ func (r *SingleClusterReconciler) getRollingRestartTypePod(
...
 
-		val, err := lib.CompareVersions(version, minInitVersionForDynamicRackID)
+		val, err := lib.CompareVersions(version, minInitVersionForOverrideRackID)
 		if err != nil {
 			return restartType, err
 		}
 
 		if val < 0 {
+			r.Log.Info(
+				"Init container version is older than minimum required, full pod restart needed",
```
abhishekdwivedi3060: "Init container version is older than minimum required for enableRackIDOverride support ,

internal/controller/cluster/pod.go
```diff
@@ -263,31 +263,42 @@ func (r *SingleClusterReconciler) getRollingRestartTypePod(
...
 			restartType = mergeRestartType(restartType, quickRestart)
 		}
 
 		r.Log.Info(
-			"EnableDynamicRackID changed. Need rolling restart",
-			"specEnableDynamicRackID", specEnableDynamicRackID,
-			"podStatusDynamicRackID", podStatusDynamicRackID,
+			"EnableRackIDOverride changed. Need rolling restart",
+			"enableRackIDOverride", specEnableRackIDOverride,
+			"overrideRackID", podStatusOverrideRackID,
```
abhishekdwivedi3060: "rackIDOverridden"

test/cluster/dynamic_rackid_test.go
```diff
@@ -92,42 +127,87 @@ var _ = Describe(
...
+
+								hostConns, err := newAllHostConn(logger, aeroCluster, k8sClient)
+								Expect(err).ToNot(HaveOccurred())
+
+								expectedRoster := fmt.Sprintf("1A1@%s,1A0@%s", secondPodRackAnnotation, firstPodRackAnnotation)
+								isEqual, currentRoster, err := compareRoster(hostConns[0], expectedRoster, aeroCluster)
+								Expect(err).ToNot(HaveOccurred())
+								Expect(isEqual).To(BeTrue(), fmt.Sprintf(
+									"expected %v to equal %v", currentRoster, expectedRoster,
+								))
```
abhishekdwivedi3060: Can we move these checks in the validateDynamicRackIDInConfig func?

test/cluster/dynamic_rackid_test.go
```diff
@@ -138,44 +218,189 @@ var _ = Describe(
...
+
+								By("Verifying restart occurred")
+								restartedPod := &v1.Pod{}
+								err = k8sClient.Get(ctx, types.NamespacedName{
+									Name:      aeroCluster.Name + "-1-1",
+									Namespace: aeroCluster.Namespace,
+								}, restartedPod)
+								Expect(err).ToNot(HaveOccurred())
+
+								Expect(string(restartedPod.UID)).ToNot(Equal(podPIDMap[aeroCluster.Name+"-1-1"].podUID))
```
abhishekdwivedi3060: Shouldn't this go under Eventually block?

api/v1/aerospikecluster_types.go
```diff
@@ -191,6 +191,16 @@ type AerospikeClusterSpec struct { //nolint:govet // for readability
 	// +kubebuilder:validation:MaxItems:=1
 	// +optional
 	Operations []OperationSpec `json:"operations,omitempty"`
+
+	// EnableRackIDOverride enables dynamic allocation of rack IDs to pods after they get scheduled.
+	// When enabled, the operator will watch for changes to the aerospike.com/override-rack-id annotation on pods
+	// and reconcile when this annotation value changes. This allows rack IDs to be dynamically assigned to pods
+	// based on their scheduling location or other external factors.
```
abhishekdwivedi3060: We are saying this in our description but this is not actually true. If annotation is changed directly via kubectl then AKO reconciles but doesn't do anything apart from roster set. We need to set the expectations right like this is only considered at the time of pod start.

abhishekdwivedi3060 (approved):

PR #437, opened 2026-02-12
enhancement: Deprecate cross-namespace secret reference and skipXdrDlogFileValidate [KO-473] [KO-478]

abhishekdwivedi3060:
- Deprecate `secretNamespace` field in AerospikeOperatorCertSource struct
- Deprecate `skipXdrDlogFileValidate` field in ValidationPolicySpec struct
- Ignore system namespace in the eviction webhook configuration


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -502,6 +505,60 @@ func validateOperatorClientCert(clientCert *asdbv1.AerospikeOperatorClientCertSp
...
+// warnOperatorClientCertSecretNamespace returns admission warnings when the deprecated
+// secretNamespace field is used in operator client cert or CA cert source.
+func warnOperatorClientCertSecretNamespace(spec *asdbv1.AerospikeOperatorClientCertSpec) admission.Warnings {
+	if spec == nil {
+		return nil
+	}
+
+	var warnings admission.Warnings
+
+	if spec.SecretCertSource == nil {
```
tanmayja: we can club it with 511 condition

PR #440, opened 2026-02-24
enhancement: Remove spec level global storage dependency when rack level storage is used

tanmayja:


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -65,13 +66,156 @@ func (acv *AerospikeClusterCustomValidator) ValidateCreate(_ context.Context, ae
...
+	// • For each rack the projected pod name is computed with a placeholder
+	//   of max(len(revision), minRevisionReservation) chars, MaxRackID, and
+	//   the max pod ordinal. Racks with revision ≤ minRevisionReservation chars
+	//   have a forward guarantee; longer revisions are validated at their actual
+	//   length. Future revision changes are caught by the actual-pod-name check
+	//   in ValidateUpdate.
+	//
+	// Note: revision character validity (IsDNS1123Label) is checked in
+	// validateRackConfig, which runs on both CREATE and UPDATE, because an
+	// invalid character would silently corrupt the StatefulSet name at runtime.
```
abhishekdwivedi3060, line 72: nit: These comments and comments on the func definition are kind of same/duplicate. Let's keep only the required comments here

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -65,13 +66,156 @@ func (acv *AerospikeClusterCustomValidator) ValidateCreate(_ context.Context, ae
...
+	// as DNS-1035, rack.Revision as DNS-1123
+	// label, and rackID/ordinal are always integers. The
+	// constructed pod name therefore always contains valid DNS label characters
+	// and never starts or ends with a hyphen — length is the only failure mode.
+	if len(worstCasePodName) > k8svalidation.DNS1123LabelMaxLength {
+		return fmt.Errorf(
+			"cluster %q: pod name %q (computed at max rack ID %d, revision length %d, max pod ordinal %d) "+
+				"exceeds the maximum DNS label length of %d characters; "+
+				"shorten the cluster name or rack revision",
+			cluster.Name, worstCasePodName, asdbv1.MaxRackID, maxPlaceholderLen,
```
abhishekdwivedi3060: I'm not sure if adding info like worstCasePodName with "aaa", "1000000" and "255" in error response will be useful for the user. I think this will confuse the user more as none of these values are inputs for the user.
Can we rephrase in a way where a user understands exactly what needs to be done

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -65,13 +66,156 @@ func (acv *AerospikeClusterCustomValidator) ValidateCreate(_ context.Context, ae
...
+		}
+	}
+
+	// A pure length check is sufficient: cluster.Name is validated as DNS-1035,
+	// rack.Revision as DNS-1123 label, and rackID/ordinal are always integers.
+	// Only length can cause the constructed pod name to be invalid.
+	if len(longestPodName) > k8svalidation.DNS1123LabelMaxLength {
+		return fmt.Errorf(
+			"pod name %q exceeds the maximum DNS label length of %d characters",
+			longestPodName, k8svalidation.DNS1123LabelMaxLength)
```
abhishekdwivedi3060: Here the longestPodName might not be actually present in the cluster, so will be difficult for a user to understand where it's coming from.
Can we rephrase to convey something like the rackConfig/szie results in new pods names that will exceed the DNS label length

PR #469, opened 2026-02-27
enhancement: Change replication-factor dynamically for AP namespaces [KO-261] [KO-485]

tanmayja:


pkg/validation/validate.go
```diff
@@ -773,20 +773,10 @@ func validateNsConfUpdate(oldConf, newConf map[string]interface{}) error {
...
-			if singleConf["name"] == oldSingleConf["name"] {
-				// replication-factor update not allowed
-				if isValueUpdated(
-					oldSingleConf, singleConf, "replication-factor",
-				) {
-					return fmt.Errorf(
-						"replication-factor cannot be updated. old nsconf %v, new nsconf %v",
-						oldSingleConf, singleConf,
-					)
-				}
```
abhishekdwivedi3060, line 785: I think we need to keep the SC related RF blocker here for the cloud.
In our case, object spec validation above handles this SC RF change but for cloud UI it won't work

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -189,6 +193,12 @@ func (acv *AerospikeClusterCustomValidator) ValidateUpdate(_ context.Context, ol
 		return warnings, err
 	}
 
+	// Validate replication-factor update restrictions
+	// Check both: spec-to-spec changes and status-to-spec changes
+	if err := validateReplicationFactorUpdateRestrictions(oldObject, aerospikeCluster); err != nil {
```
abhishekdwivedi3060, line 181: Shouldn't this be part of validateAerospikeConfigUpdate func?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2108,239 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal new spec: %v", err)
+	}
+
+	patch, err := jsonpatch.CreatePatch(oldJSON, newJSON)
+	if err != nil {
+		return fmt.Errorf("failed to create JSON patch: %v", err)
+	}
+
+	// Filter only spec changes (ignore status and metadata changes)
```
abhishekdwivedi3060: what is the comment for? This is already getting filtered by just comparing the spec

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2108,239 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+				// Extract namespace info from path
+				// Path format: /rackConfig/racks/1/effectiveAerospikeConfig/namespaces/1/replication-factor
+				nsName, scEnabled := extractNamespaceInfoFromPath(op.Path, newObj)
+				if scEnabled {
+					return fmt.Errorf(
+						"cannot update replication-factor for namespace %s with Strong Consistency enabled",
+						nsName,
+					)
+				}
+
```
abhishekdwivedi3060: See if this is still required if we keep the SC related check in validateNsConfUpdate

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2108,239 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+			}
+
+			// Extract namespace index
+			if idx, err := strconv.Atoi(parts[i+1]); err == nil {
+				nsIndex = idx
+			}
+
+			break
+		}
+	}
```
abhishekdwivedi3060: In this check, everything is in reference to "namespaces" keyword and rest of the indexing is hardcoded.
Can we not leverage the direct index of rackIndex and nsIndex ? Is there anycase where the path will be different?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2108,239 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+
+				// Check if this namespace already has a replication-factor set
+				if existingRF, exists := namespaceReplicationFactors[nsName]; exists {
+					// Compare replication-factor values
+					if existingRF != currentRF {
+						return fmt.Errorf(
+							"namespace '%s' has different replication-factor values across racks. "+
+								"Replication-factor must be the same for the same namespace across all racks. "+
+								"Found: %d and %d",
+							nsName, existingRF, currentRF,
```
abhishekdwivedi3060: nit: 
```suggestion
						return fmt.Errorf(
							"namespace '%s' has different replication-factor values across racks. "+
								"Replication-factor must be same for a namespace across all racks. "+
								"Found: %d and %d",
							nsName, existingRF, currentRF,
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2223,38 +2217,21 @@ func validateReplicationFactorUpdateComparison(oldSpec *asdbv1.AerospikeClusterS
...
+	var nsName string
 
-			// Extract namespace index
-			if idx, err := strconv.Atoi(parts[i+1]); err == nil {
-				nsIndex = idx
-			}
-
-			break
-		}
+	rackIndex, err := strconv.Atoi(parts[3])
```
abhishekdwivedi3060: To be on safe side to avoid panic, we can have add len checks for both these indexes

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2108,212 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+				if nsName != "" {
+					replicationFactorPatches.Insert(nsName)
+				}
+			}
+		} else {
+			if !strings.HasSuffix(op.Path, "/enableDynamicConfigUpdate") {
+				otherSpecChanged = true
+			}
+		}
+	}
```
abhishekdwivedi3060, line 2250: Did we check this flow with namespace addition/removal? I guess these checks might block those along with some other field change. Confirm this once

internal/webhook/v1/aerospikecluster_validating_webhook.go
tanmayja: In that case, this validation will run multiple times for all racks and global config which is not needed.

internal/webhook/v1/aerospikecluster_validating_webhook.go
tanmayja: That is handled already as adding/removing an element from list is a whole patch.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2151,238 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+	rfChangedNamespaces := nsNamesFromRFChanges(actualRFChanges)
+
+	// Rule: only one namespace's replication-factor may be changed per update.
+	if len(rfChangedNamespaces) > 1 {
+		return fmt.Errorf(
+			"cannot update replication-factor for multiple namespaces at the same time; "+
+				"changed namespaces: %v",
+			rfChangedNamespaces.UnsortedList(),
+		)
+	}
```
abhishekdwivedi3060, line 2277: I think we can move this check above statusNsNames to fail fast

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2151,238 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+
+	var otherSpecChanged bool
+
+	for _, op := range patch {
+		if strings.HasSuffix(op.Path, "/"+asdbv1.ConfKeyReplicationFactor) {
+			// RF changes are detected by direct value comparison below; they are never
+			// counted as "other spec changed" regardless of where in the path they appear.
+		} else if !strings.HasSuffix(op.Path, "/enableDynamicConfigUpdate") {
+			otherSpecChanged = true
+		}
```
abhishekdwivedi3060: Lets change it to single if statement as we are not doing anything in the if statement

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2151,238 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+
+	// No genuine RF value changes – nothing further to validate.
+	if actualRFChanges.Len() == 0 {
+		return nil
+	}
+
+	// If a deployed-status set was provided, verify that every RF-changed namespace already
+	// exists there. An RF change on a namespace that hasn't been deployed yet is not allowed.
+	if statusNsNames != nil && !statusNsNames.HasAll(actualRFChanges.UnsortedList()...) {
+		otherSpecChanged = true
```
abhishekdwivedi3060: For this type of scenario, below error string doesn't make sense.
It should explicitly say that RF can not be updated at the time namespace addition rollout is inprogress

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -2093,3 +2151,238 @@ func validateTLSAndPKIOnlyUpdate(oldSpec, newSpec *asdbv1.AerospikeClusterSpec,
...
+
+		// Normalize to int for comparison
+		rfInt, err := validation.GetIntType(rf)
+		if err != nil {
+			// If we can't parse it, return 0 to trigger an error
+			return 0
+		}
+
+		return rfInt
+	}
```
abhishekdwivedi3060: We can use the func `validation.GetNamespaceReplicationFactor` instead

PR #477, opened 2026-03-12
Test: Dynamic RF Update TCs

rpandey-stack:
This PR is created to add 'Dynamic RF Update' TCs in test/envtests

**TCs Added:**
**Deploy validation — spec.aerospikeConfig (replication-factor)**
  negative: rejects RF > cluster size for SC namespace
  positive: allows AP namespace with RF greater than cluster size on deploy

**Deploy validation — spec.rackConfig (replication-factor)**
  negative: rejects rack-aware create when RF values are mismatched across racks
  positive: allows rack-aware deploy with replication-factor 256 in aerospikeConfig and all racks (schema max)

**Update validation — spec.aerospikeConfig (replication-factor)**
  negative: rejects RF change attempted for SC namespace
  negative: rejects RF change combined with unrelated spec changes
  negative: rejects RF update when enableDynamicConfigUpdate is false
  negative: rejects invalid RF value (zero or negative)
  negative: rejects RF update in spec.aerospikeConfig.namespaces (above schema max 256)
  positive: allows basic RF increase for AP namespace
  positive: allows basic RF decrease for AP namespace
  positive: allows no-op RF update (same RF value)
  positive: allows RF increase greater than cluster size for AP namespace
  positive: allows RF set to minimum value (RF=1)
  positive: allows RF set to maximum allowed (RF = cluster size)
  positive: allow RF remove from spec.aerospikeConfig for AP namespace

**Update validation — spec.rackConfig (replication-factor)**
  negative: rejects RF update if RF is inconsistent across racks for same namespace)
  negative: rejects simultaneous RF update across multiple namespaces
  negative: rejects RF update if RF is inconsistent in spec.aerospikeConfig.namespaces and spec.rackConfig.aerospikeConfig for the same namespace
  negative: rejects RF update in rackConfig for value above schema max 256)
  positive: allows rack-aware RF increase when consistent across racks
  positive: allows rack-aware RF increase greater than cluster size
  positive: allow adding namespace to rackConfig.namespaces with RF
  positive: allow removing replication-factor from namespace in rackConfig.aerospikeConfig

### DIFF
--- a/test/envtests/cluster/access_control_webhook_test.go
+++ b/test/envtests/cluster/access_control_webhook_test.go
@@ -89,7 +89,11 @@ var _ = Describe("AerospikeCluster access control validation (envtests)", func()
 					Namespace: clusterNamespacedName.Namespace,
 				},
 			}
-			_ = envtests.K8sClient.Delete(ctx, aeroCluster)
+			Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+			Expect(testCluster.CleanupPVC(
+				envtests.K8sClient,
+				clusterNamespacedName.Namespace,
+				clusterNamespacedName.Name)).ToNot(HaveOccurred())
 		})
 
 		Context("spec.aerospikeAccessControl (validation)", func() {
@@ -196,7 +200,11 @@ var _ = Describe("AerospikeCluster access control validation (envtests)", func()
 					Namespace: clusterNamespacedName.Namespace,
 				},
 			}
-			_ = envtests.K8sClient.Delete(ctx, aeroCluster)
+			Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+			Expect(testCluster.CleanupPVC(
+				envtests.K8sClient,
+				clusterNamespacedName.Namespace,
+				clusterNamespacedName.Name)).ToNot(HaveOccurred())
 		})
 		Context("spec.aerospikeAccessControl (users)", func() {
 			Context("negative", func() {
--- a/test/envtests/cluster/access_control_webhook_test.go
+++ b/test/envtests/cluster/access_control_webhook_test.go
@@ -76,11 +76,10 @@ func adminOperatorCertForTest() *asdbv1.AerospikeOperatorClientCertSpec {
 var _ = Describe("AerospikeCluster access control validation (envtests)", func() {
 	const (
 		clusterName = "access-control-webhook-cluster"
-		testNs      = "default"
 	)
 
 	ctx := context.TODO()
-	clusterNamespacedName := test.GetNamespacedName(clusterName, testNs)
+	clusterNamespacedName := test.GetNamespacedName(clusterName, testutil.DefaultNamespace)
 
 	Context("Deploy validation", func() {
 		AfterEach(func() {
--- a/test/envtests/cluster/access_control_webhook_test.go
+++ b/test/envtests/cluster/access_control_webhook_test.go
@@ -90,10 +90,6 @@ var _ = Describe("AerospikeCluster access control validation (envtests)", func()
 				},
 			}
 			Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
-			Expect(testCluster.CleanupPVC(
-				envtests.K8sClient,
-				clusterNamespacedName.Namespace,
-				clusterNamespacedName.Name)).ToNot(HaveOccurred())
 		})
 
 		Context("spec.aerospikeAccessControl (validation)", func() {
@@ -201,10 +197,6 @@ var _ = Describe("AerospikeCluster access control validation (envtests)", func()
 				},
 			}
 			Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
-			Expect(testCluster.CleanupPVC(
-				envtests.K8sClient,
-				clusterNamespacedName.Namespace,
-				clusterNamespacedName.Name)).ToNot(HaveOccurred())
 		})
 		Context("spec.aerospikeAccessControl (users)", func() {
 			Context("negative", func() {
--- a/test/envtests/cluster/dynamic_rf_webhook_test.go
+++ b/test/envtests/cluster/dynamic_rf_webhook_test.go
@@ -47,8 +47,8 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 	// apNamespaceConfig returns an AP namespace config for envtest.
 	apNamespaceConfig := func(name string, rf int, device string) map[string]interface{} {
 		return map[string]interface{}{
-			"name":               name,
-			"replication-factor": rf,
+			asdbv1.ConfKeyName:              name,
+			asdbv1.ConfKeyReplicationFactor: rf,
 			asdbv1.ConfKeyStorageEngine: map[string]interface{}{
 				"type":    "device",
 				"devices": []interface{}{device},
@@ -59,15 +59,15 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 	// scNamespaceConfig returns an SC namespace config for envtest.
 	scNamespaceConfig := func(name string, rf int) map[string]interface{} {
 		cfg := apNamespaceConfig(name, rf, testutil.DefaultDevicePath)
-		cfg["strong-consistency"] = true
+		cfg[asdbv1.ConfKeyStrongConsistency] = true
 
 		return cfg
 	}
 
 	inMemoryNS := func(name string, rf int) map[string]interface{} {
 		return map[string]interface{}{
-			"name":               name,
-			"replication-factor": rf,
+			asdbv1.ConfKeyName:              name,
+			asdbv1.ConfKeyReplicationFactor: rf,
 			asdbv1.ConfKeyStorageEngine: map[string]interface{}{
 				"type":      "memory",
 				"data-size": 1073741824,
@@ -76,13 +76,14 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 	}
 
 	// updateNamespaceRF creates a cluster with one namespace (AP or SC), updates its RF to newRF,
-	// and returns the update error. namespaceType is "AP" or "SC".
-	updateNamespaceRF := func(namespaceType string, size int32, initialRF, newRF int) error {
+	// and returns the Update error. If rackCount is 0, the RF change is applied to spec.aerospikeConfig
+	// only. If rackCount > 0, RackConfig is populated with that many racks (each with InputAerospikeConfig
+	// for the namespace), and the RF change is applied only to each rack's InputAerospikeConfig.
+	updateNamespaceRF := func(isSCNamespace bool, size int32, initialRF, newRF int, rackCount int) error {
 		var nsConfig map[string]interface{}
-		switch namespaceType {
-		case "SC":
+		if isSCNamespace {
 			nsConfig = scNamespaceConfig(namespaceName, initialRF)
-		default:
+		} else {
 			nsConfig = apNamespaceConfig(namespaceName, initialRF, testutil.DefaultDevicePath)
 		}
 
@@ -90,15 +91,53 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 		aeroCluster.Spec.EnableDynamicConfigUpdate = ptr.To(true)
 		aeroCluster.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = []interface{}{nsConfig}
 
+		if rackCount > 0 {
+			racks := make([]asdbv1.Rack, rackCount)
+			for i := 0; i < rackCount; i++ {
+				var rackNS map[string]interface{}
+				if isSCNamespace {
+					rackNS = scNamespaceConfig(namespaceName, initialRF)
+				} else {
+					rackNS = apNamespaceConfig(namespaceName, initialRF, testutil.DefaultDevicePath)
+				}
+				racks[i] = asdbv1.Rack{
+					ID: i + 1,
+					InputAerospikeConfig: &asdbv1.AerospikeConfigSpec{
+						Value: map[string]interface{}{
+							asdbv1.ConfKeyNamespace: []interface{}{rackNS},
+						},
+					},
+				}
+			}
+			aeroCluster.Spec.RackConfig = asdbv1.RackConfig{
+				Namespaces: []string{namespaceName},
+				Racks:      racks,
+			}
+		}
+
 		err := envtests.K8sClient.Create(ctx, aeroCluster)
 		Expect(err).ToNot(HaveOccurred())
 
 		current, err := testCluster.GetCluster(envtests.K8sClient, ctx, clusterNamespacedName)
 		Expect(err).ToNot(HaveOccurred())
 
+		if rackCount > 0 {
+			for i := range current.Spec.RackConfig.Racks {
+				rack := &current.Spec.RackConfig.Racks[i]
+				Expect(rack.InputAerospikeConfig).ToNot(BeNil())
+				nsList := rack.InputAerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
+				nsConf := nsList[0].(map[string]interface{})
+				nsConf[asdbv1.ConfKeyReplicationFactor] = newRF
+				nsList[0] = nsConf
+				rack.InputAerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
+			}
+
+			return envtests.K8sClient.Update(ctx, current)
+		}
+
 		nsList := current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
 		nsConf := nsList[0].(map[string]interface{})
-		nsConf["replication-factor"] = newRF
+		nsConf[asdbv1.ConfKeyReplicationFactor] = newRF
 		nsList[0] = nsConf
 		current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
 
@@ -112,7 +151,11 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 				Namespace: clusterNamespacedName.Namespace,
 			},
 		}
-		_ = envtests.K8sClient.Delete(ctx, aeroCluster)
+		Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+		Expect(testCluster.CleanupPVC(
+			envtests.K8sClient,
+			clusterNamespacedName.Namespace,
+			clusterNamespacedName.Name)).ToNot(HaveOccurred())
 	})
 
 	Context("Deploy validation", func() {
@@ -243,7 +286,7 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 		Context("spec.aerospikeConfig (replication-factor)", func() {
 			Context("negative", func() {
 				It("rejects RF change attempted for SC namespace", func() {
-					err := updateNamespaceRF("SC", 3, 2, 3)
+					err := updateNamespaceRF(true, 3, 2, 3, 0)
 					Expect(err).To(HaveOccurred())
 					envtests.NewStatusErrorMatcher().
 						WithMessageSubstrings(
@@ -268,7 +311,7 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 
 					nsList := current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
 					nsConf := nsList[0].(map[string]interface{})
-					nsConf["replication-factor"] = 3
+					nsConf[asdbv1.ConfKeyReplicationFactor] = 3
 					nsList[0] = nsConf
 					current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
 					current.Spec.Size = 4
@@ -300,7 +343,7 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 
 					nsList := current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
 					nsConf := nsList[0].(map[string]interface{})
-					nsConf["replication-factor"] = 3
+					nsConf[asdbv1.ConfKeyReplicationFactor] = 3
 					nsList[0] = nsConf
 					current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
 
@@ -319,7 +362,7 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 				It("rejects RF update in spec.aerospikeConfig.namespaces (above schema max 256)", func() {
 					// Pre-conditions: AP namespace; RF=2; enableDynamicConfigUpdate=true;
 					// update to RF=257 (above schema max 256)
-					err := updateNamespaceRF("AP", 3, 2, rfAboveSchemaMax)
+					err := updateNamespaceRF(false, 3, 2, rfAboveSchemaMax, 0)
 					Expect(err).To(HaveOccurred())
 
 					envtests.NewStatusErrorMatcher().
@@ -351,8 +394,8 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation", func(
 					// Update RF for test2 and test3 in one update
 					for i := range nsList {
 						nsConf := nsList[i].(map[string]interface{})
-						if nsConf["name"] == namespaceName2 || nsConf["name"] == namespaceName3 {
-							nsConf["replication-factor"] = 3
+						if nsConf[asdbv1.ConfKeyName] == namespaceName2 || nsConf["name"] == namespaceName3 {
+							nsConf[asdbv1.ConfKeyReplicationFactor] = 3
 							nsList[i] = nsConf
 						}
 					}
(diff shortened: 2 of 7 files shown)


test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+	"k8s.io/utils/ptr"
+
+	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test"
+	testCluster "github.com/aerospike/aerospike-kubernetes-operator/v4/test/cluster"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test/envtests"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test/testutil"
+)
+
+var _ = Describe("AerospikeCluster dynamic replication-factor validation (envtests)", func() {
```
abhishekdwivedi3060:
```suggestion
var _ = Describe("AerospikeCluster dynamic replication-factor validation", func() {
```

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+	const (
+		clusterName       = "dynamic-rf-webhook-cluster"
+		testNs            = "default"
+		maxSchemaRF       = 256
+		rfAboveSchemaMax  = 257
+		defaultDevicePath = "/test/dev/xvdf"
+		namespaceName     = "test"
+		namespaceName2    = "test2"
+		namespaceName3    = "test3"
+	)
```
abhishekdwivedi3060, line 42: I think its better to move some of these constants like namespaces, devicePath outside at the `cluster` pkg level so that these can be reused in other tests

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		nsList := current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
+		nsConf := nsList[0].(map[string]interface{})
+		nsConf["replication-factor"] = newRF
+		nsList[0] = nsConf
+		current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
+
+		return envtests.K8sClient.Update(ctx, current)
+	}
```
abhishekdwivedi3060: This seems duplicate of updateAPNamespaceRF func expect the SC namespace part. See if we can have a common func called `updateNamespaceRF` which AP/SC namespace as a param based on test

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+				Namespace: clusterNamespacedName.Namespace,
+			},
+		}
+		_ = envtests.K8sClient.Delete(ctx, aeroCluster)
+	})
+
+	Context("Deploy validation", func() {
+		Context("spec.aerospikeConfig (replication-factor)", func() {
+			Context("negative", func() {
+				It("rejects RF > cluster size for SC namespace", func() {
```
abhishekdwivedi3060: This test-case already present in cluster env file cluster_webhook_test.go

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+						Validate(err)
+				})
+			})
+
+			Context("positive", func() {
+				It("allows AP namespace with RF greater than cluster size on deploy", func() {
+					// AP mode: replication-factor may exceed cluster size (unlike SC); see validateNamespaceReplicationFactor.
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(clusterNamespacedName, 3)
+					ns := apNamespaceConfig(namespaceName, 4, defaultDevicePath)
+					ns["replication-factor"] = 4
```
abhishekdwivedi3060: nit: Not required. Already setting it to 4

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+						).
+						Validate(err)
+				})
+
+				It("rejects RF change combined with unrelated spec changes", func() {
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(clusterNamespacedName, 3)
+					aeroCluster.Spec.EnableDynamicConfigUpdate = ptr.To(true)
+					aeroCluster.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = []interface{}{
+						apNamespaceConfig(namespaceName, 2, defaultDevicePath),
+					}
```
abhishekdwivedi3060, line 300: CreateDummyAerospikeCluster func already returns namepsaces with SC. Any reason for doing it again here? Check all other places as well

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					err := envtests.K8sClient.Create(ctx, aeroCluster)
+					Expect(err).To(HaveOccurred())
+
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"vaerospikecluster.kb.io",
+							"replication-factor can not be zero or negative",
+						).
+						Validate(err)
+				})
```
abhishekdwivedi3060: This should go in Deploy validation test

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+
+				It("allows basic RF decrease for AP namespace", func() {
+					// Pre-conditions: AP namespace; RF initially 3; enableDynamicConfigUpdate=true
+					updateAPNamespaceRF(3, 3, 2)
+				})
+
+				It("allows no-op RF update (same RF value)", func() {
+					// Pre-conditions: AP namespace; RF=3; enableDynamicConfigUpdate=true; update to same RF
+					updateAPNamespaceRF(3, 3, 3)
+				})
```
abhishekdwivedi3060, line 426: This seems redundant

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					// Pre-conditions: size ≥ 3; AP namespace; RF=2; enableDynamicConfigUpdate=true
+					updateAPNamespaceRF(3, 2, 4)
+				})
+
+				It("allows RF set to minimum value (RF=1)", func() {
+					updateAPNamespaceRF(2, 2, 1)
+				})
+
+				It("allows RF set to maximum allowed (RF = cluster size)", func() {
+					const clusterSize = 3
```
abhishekdwivedi3060: No need of the constant just for this test-case

abhishekdwivedi3060: IMO, these RF test-cases are kind of duplicate.
Won't having below 2 test-cases cover all scenearios?
1. It("allows RF set to minimum value (RF=1)"
2. It("allows RF set to maximum allowed (256)"

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+						WithMessageSubstrings(
+							"vaerospikecluster.kb.io",
+							"replication-factor Must be less than or equal to 256 ",
+						).
+						Validate(err)
+				})
+			})
+
+			Context("positive", func() {
+				It("allows basic RF increase for AP namespace", func() {
```
abhishekdwivedi3060: All the test-cases are for AP. In some test-cases we are explicitly saying AP namespace, keep them consistent across all

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).To(HaveOccurred())
+
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"vaerospikecluster.kb.io",
+							"replication-factor Must be less than or equal to 256 ",
+						).
+						Validate(err)
```
abhishekdwivedi3060, line 369: See if we can leverage the updateAPNamespaceRF func for this. Everything is same except the error part. Try returning error or something similar

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).To(HaveOccurred())
+
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"vaerospikecluster.kb.io",
+							"cannot update replication-factor for multiple namespaces at the same time",
+						).
+						Validate(err)
+				})
```
abhishekdwivedi3060: We only need to test RF update across namespaces. We can simplify this test-case to only use multiple namespaces in global aerospikeConfig and update RF for all of them in 1 update. No need for rack level aerospikeConfig check

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					Expect(err).To(HaveOccurred())
+
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"vaerospikecluster.kb.io",
+							"different replication-factor values across racks",
+							"replication-factor must be the same in every rack",
+						).
+						Validate(err)
+				})
```
abhishekdwivedi3060: What different use-case are we testing in this case which is not already covered in `It("rejects RF update if RF is inconsistent across racks for same namespace)"` test-case apart from 2 vs 1 namespace update?

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					Expect(err).ToNot(HaveOccurred())
+
+					current, err := testCluster.GetCluster(envtests.K8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+
+					nsList := current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
+					nsConf := nsList[0].(map[string]interface{})
+					nsConf["replication-factor"] = rfAboveSchemaMax
+					nsList[0] = nsConf
+					current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
```
abhishekdwivedi3060: This code block is repeated multiple times at multiple places. Can move to a common func to reduce code duplication

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					nsList := current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
+					nsConf := nsList[0].(map[string]interface{})
+					nsConf["replication-factor"] = 4
+					nsList[0] = nsConf
+					current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
+
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
```
abhishekdwivedi3060: Can we not club these 2?

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,887 @@
...
+					current.Spec.RackConfig.Racks[0].InputAerospikeConfig = &asdbv1.AerospikeConfigSpec{
+						Value: map[string]interface{}{
+							asdbv1.ConfKeyNamespace: []interface{}{inMemoryNS(namespaceName2, 3)},
+						},
+					}
+
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).NotTo(HaveOccurred())
+				})
+
```
abhishekdwivedi3060: What are we exactly testing in this? We are just adding a new namespace without any RF manipulation

abhishekdwivedi3060 (changes requested):

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: CreateDummyAerospikeCluster uses an SC namespace. This test case must use an AP namespace.For that reason we override the default namespace config(i.e SC namespace config)  with an AP namespace here. The same logic applies anywhere else in this file where we need AP semantics for RF update tests.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: From a use-case perspective, I think we should check if updates can be applied even if we are not updating any values (include no change in RF value). That's why I have kept this TCs.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: 1. These TCs are included from a use-case point of view. RF = 256 checks the maximum allowed limit and RF =1 checks the minimum allowed limit.
2. I think RF= 256 should certainly be checked here as it is the upper limit.
3. RF=1 might seem unnecessary or an extra check but it is added just to ensure that minimum positive value is always passing. It is more related to boundary value test for the condition.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/cab83f78ec1d4450941110546fc4a0b4199abe6c

Removed "(envtests)" from Describe() string.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/cab83f78ec1d4450941110546fc4a0b4199abe6c

Created updateNamespaceRF with cases for AP/SC.

Usage example:
updateNamespaceRF("SC", 3, 2, 3)
updateNamespaceRF("AP", 3, 2, 3)

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

Removed this TC from this file.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

Moved it to deploy validation

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Address it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

Removed the const

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

Removed string "AP" or "AP namespace" everywhere in this file.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

used updateNamespaceRF to reduce redundant code.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/commit/cab83f78ec1d4450941110546fc4a0b4199abe6c

Used updateNamespaceRF here.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/920f6cebb20ffdf580a5da7440b1f2a0334a4d88

Moved testNs and defaultDevicePath to defaults.go
testNs == DefaultNamespace
defaultDevicePath == DefaultDevicePath

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Removed it.

Commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/027d34375dc6a2f1efcc1c6979fae04c7caacfa0

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: This is the TC:

- Preconditions: RackConfig.namespaces has "test". Rack has no InputAerospikeConfig (uses global).
- Global has test with RF=2.
- Update: Add InputAerospikeConfig to rack with test at RF=2.
- Must succeed: rack acquires RF override for namespace already in rackConfig.namespaces.

What are we testing?
Ans: if there is no RF in rackConfig during cluster creation, then one should be able to add RF parameter in rackConfig (provided it remains consistent or in-line with spec.aerospikeConfig's RF) using an update.

I have modified the code using commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/027d34375dc6a2f1efcc1c6979fae04c7caacfa0

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/027d34375dc6a2f1efcc1c6979fae04c7caacfa0

Clubbed the 2 TCs using `DescribeTable("allows rack-aware RF increase when consistent across racks")`

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/027d34375dc6a2f1efcc1c6979fae04c7caacfa0

Changes done:

- Moved TC to Context -->spec.aerospikeConfig (replication-factor)-->negative
- Removed rackConfig from TC. Now it has only global aerospikeConfig.

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,689 @@
...
+	}
+
+	AfterEach(func() {
+		aeroCluster := &asdbv1.AerospikeCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      clusterNamespacedName.Name,
+				Namespace: clusterNamespacedName.Namespace,
+			},
+		}
+		_ = envtests.K8sClient.Delete(ctx, aeroCluster)
```
tanmayja: Every negative test that calls the helper creates a cluster, attempts the update, fails, and then the AfterEach deletes. If object is still terminating, a second It() block using the same helper could get a name collision. the AfterEach uses _ = Delete (ignores error), which could mask cleanup failures.

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,689 @@
...
+				"type":    "device",
+				"devices": []interface{}{device},
+			},
+		}
+	}
+
+	// scNamespaceConfig returns an SC namespace config for envtest.
+	scNamespaceConfig := func(name string, rf int) map[string]interface{} {
+		cfg := apNamespaceConfig(name, rf, testutil.DefaultDevicePath)
+		cfg["strong-consistency"] = true
```
tanmayja: try to use the constant already defined in asdbv1 wherever possible. for example here asdbv1.ConfKeyStrongConsistency

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,689 @@
...
+					current.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
+
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
+				It("allows adding RF to existing namespace in RackConfig.namespaces", func() {
+					// Preconditions: RackConfig.namespaces has "test". Rack has no InputAerospikeConfig (uses global).
+					// Global has test with RF=2.
+					// Update: Add InputAerospikeConfig to rack with test at RF=3.
```
tanmayja: The comment here says "Update: Add InputAerospikeConfig to rack with test at RF=3" but the actual update on line 593 uses inMemoryNS(namespaceName, 2) — RF=2, same as the initial value.

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -0,0 +1,689 @@
...
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"vaerospikecluster.kb.io",
+							"different replication-factor values across racks",
+							"replication-factor must be the same in every rack",
+						).
+						Validate(err)
+				})
+
+				It("rejects RF update in rackConfig for value above schema max 256)", func() {
```
tanmayja: 1- nit: test description has a stray ) inside the string.
2- I don't see this test updates rf in rackConfig, it is updating spec.

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -79,39 +75,20 @@ var _ = Describe("AerospikeCluster dynamic replication-factor validation (envtes
...
-		aeroCluster := testCluster.CreateDummyAerospikeCluster(clusterNamespacedName, size)
-		aeroCluster.Spec.EnableDynamicConfigUpdate = ptr.To(true)
-		aeroCluster.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = []interface{}{
-			apNamespaceConfig(namespaceName, initialRF, defaultDevicePath),
+	// updateNamespaceRF creates a cluster with one namespace (AP or SC), updates its RF to newRF,
+	// and returns the update error. namespaceType is "AP" or "SC".
+	updateNamespaceRF := func(namespaceType string, size int32, initialRF, newRF int) error {
+		var nsConfig map[string]interface{}
+		switch namespaceType {
+		case "SC":
```
abhishekdwivedi3060: Two options here:
1. Define constants for SC and AP and use them in func OR
2. Use bool param instead like isSCNamespace and based on that decide the type of namespace. Pass true or false in the caller func

test/envtests/cluster/dynamic_rf_webhook_test.go
abhishekdwivedi3060: Can't reuse the func updateNamespaceRF for rackConfig.AerospikeConfig

test/envtests/cluster/dynamic_rf_webhook_test.go
abhishekdwivedi3060: We clubbed these in a DescribeTable but didn't remove the individual test-cases

test/envtests/cluster/dynamic_rf_webhook_test.go
abhishekdwivedi3060: > Ans: if there is no RF in rackConfig during cluster creation, then one should be able to add RF parameter in rackConfig (provided it remains consistent or in-line with spec.aerospikeConfig's RF) using an update.

If there is no RF in rackConfig, global spec level RF is mutated in the rackCofig. But in test-case we are adding the same RF in rackConfig during update

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: In TC#1, replication-factor and effective_replication_factor remains same whereas in TC#2 replication-factor and effective_replication_factor are different.
Also, in case of SC RF update, updating RF>clusterSize is not allowed but RF update within clusterSize is possible. So these 2 TCs are different in case of AP when it is compared to SC RF update behaviour.
May be we can add DB validation in future to complete the end-to-end validation. Hence the TCs were clubbed using DescribeTable but not removed.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Address it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/f4494ba0f33e15a7290b37518818c69a16fac2c9

Used below constants everywhere:

- asdbv1.ConfKeyName

- asdbv1.ConfKeyReplicationFactor

- asdbv1.ConfKeyStorageEngine

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/f4494ba0f33e15a7290b37518818c69a16fac2c9

Implemented option#2

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/f4494ba0f33e15a7290b37518818c69a16fac2c9

Made changes in all `test/envtests/*_test.go` to make code consistent across all files.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/f4494ba0f33e15a7290b37518818c69a16fac2c9

Changes done:
1. Removed stray ")" from TC description.
2. Corrected code to update RF in rackConfig
3. Updated updateNamespaceRF implementation to include rackConfig update.

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/f4494ba0f33e15a7290b37518818c69a16fac2c9

Corrected the TC comments to match TC implementation.

test/envtests/cluster/access_control_webhook_test.go
```diff
@@ -89,7 +89,11 @@ var _ = Describe("AerospikeCluster access control validation (envtests)", func()
 					Namespace: clusterNamespacedName.Namespace,
 				},
 			}
-			_ = envtests.K8sClient.Delete(ctx, aeroCluster)
+			Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+			Expect(testCluster.CleanupPVC(
+				envtests.K8sClient,
+				clusterNamespacedName.Namespace,
+				clusterNamespacedName.Name)).ToNot(HaveOccurred())
```
abhishekdwivedi3060: CleanupPVC is not required in envtest as these are just webhook test no e2e test. Remove it from everywhere

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Removed the individual TCs in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/c6e373efa16bb8de2d66e5be4db4d5bf7743f642

test/envtests/cluster/access_control_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/477/changes/d0bb103560bc127ba5214f6142af9de007dbc6e5

abhishekdwivedi3060 (approved):

PR #479, opened 2026-03-12
feat: Add support for ABS v3.5.0 [KO-488]

abhishekdwivedi3060:
- Added support for ABS v3.5.0
- Disabled dependabot config for gomod pkg ecosystem
- Added test-cases


pkg/backup-service/client.go
```diff
@@ -605,6 +627,36 @@ func (c *Client) GetFullBackupsForRoutine(routineName string) ([]interface{}, er
...
+
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to get incremental backups")
```
tanmayja, line 637: GetFullBackupsForRoutine has a vague error "failed to get backups" while the new GetIncrementalBackupsForRoutine says "failed to get incremental backups" — inconsistent.

pkg/backup-service/client.go
tanmayja, line 70: NewClient should simply call NewClientWithVersion(address, port, contextPath, "") to avoid drift if the struct grows more fields.

pkg/backup-service/client.go
```diff
@@ -631,6 +683,96 @@ func (c *Client) ScheduleBackup(routineName string, delay metav1.Duration) error
 	return nil
 }
 
+// TriggerFullBackup triggers a full on-demand backup for the given routine.
```
tanmayja, line 682: Both these functions are newly added, can we have just 1 generic function TriggerBackup and pass backuptype as argument?
This will remove the hardcoded "full" and "incremental" strings also.
Consider strings.ToLower(string(backupType)) to make the derivation explicit and tie it to the type.

pkg/backup-service/client.go
```diff
@@ -631,6 +683,96 @@ func (c *Client) ScheduleBackup(routineName string, delay metav1.Duration) error
...
+	}
+
+	if useNewAPI {
+		switch backupType {
+		case v1beta1.IncrementalBackup:
+			return c.TriggerIncrementalBackup(routineName, delay)
+		case v1beta1.FullBackup:
+			return c.TriggerFullBackup(routineName, delay)
+		default:
+			return fmt.Errorf("unknown backup type")
```
tanmayja: print backupType in error msg

internal/controller/backup/reconciler.go
```diff
@@ -496,9 +499,9 @@ func (r *SingleBackupReconciler) checkForDeletedRoutines(
 	return false
 }
 func (r *SingleBackupReconciler) reconcileOnDemandBackup() error {
-	// Schedule on-demand backup if given
+	// Trigger on-demand backup if given
 	if len(r.aeroBackup.Spec.OnDemandBackups) > 0 {
-		if err := r.scheduleOnDemandBackup(); err != nil {
+		if err := r.triggerOnDemandBackup(); err != nil {
 			r.Log.Error(err, "Failed to schedule backup")
```
tanmayja: should be "Failed to trigger backup"

internal/webhook/v1beta1/aerospikebackup_webhook.go
```diff
@@ -232,17 +244,16 @@ func getValidatedAerospikeClusters(backupConfig map[string]interface{},
 	return aeroClusters, nil
 }
 
-func validateOnDemandBackupsUpdate(oldObj, newObj *asdbv1beta1.AerospikeBackup) error {
-	// Check if backup config is updated along with onDemand backup add/update
-	if !reflect.DeepEqual(newObj.Spec.OnDemandBackups, newObj.Status.OnDemandBackups) &&
+func validateOnDemandBackupsUpdate(oldObj, newObj *asdbv1beta1.AerospikeBackup, version string) error {
+	// Normalise the Type field in the status to handles AKO upgrade scenarios where Type is not present in the CR
+	if !reflect.DeepEqual(newObj.Spec.OnDemandBackups, normaliseOnDemandTypes(newObj.Status.OnDemandBackups)) &&
```
tanmayja: check how deepEqual reacts if 1st argument is empty []OnDemandBackupSpec{} and second argument is nil.
normaliseOnDemandTypes returns nil if empty.

pkg/backup-service/client.go
```diff
@@ -631,6 +683,96 @@ func (c *Client) ScheduleBackup(routineName string, delay metav1.Duration) error
...
+		}
+	}
+
+	// Legacy path: ABS < 3.5.0 only supports full backups via /backups/schedule/{name}.
+	if backupType == v1beta1.IncrementalBackup {
+		return fmt.Errorf(
+			"%s on-demand backups require ABS >= %s; current version: %s",
+			v1beta1.IncrementalBackup, v1beta1.BackupSvcNewOnDemandAPIVersion, c.Version,
```
tanmayja, line 750: is it possible in any scenario that this errpr message prints "" c.Version. GetBackupServiceClient can populate version with "" in case of error.

internal/webhook/v1beta1/aerospikebackup_webhook.go
```diff
@@ -255,9 +266,52 @@ func validateOnDemandBackupsUpdate(oldObj, newObj *asdbv1beta1.AerospikeBackup)
...
+
+// validateOnDemandBackupType checks that IncrementalBackup is not requested against
+// an ABS version that does not support it (< 3.5.0), failing fast at admission time.
+func validateOnDemandBackupType(version string, backup *asdbv1beta1.AerospikeBackup) error {
+	for _, onDemand := range backup.Spec.OnDemandBackups {
+		if onDemand.Type != asdbv1beta1.IncrementalBackup {
+			continue
+		}
+
+		cmp, err := lib.CompareVersions(version, asdbv1beta1.BackupSvcNewOnDemandAPIVersion)
```
tanmayja, line 291: The version comparison result is the same for every iteration. Compute it once before the loop or after once you know its atleast one incremental.

pkg/backup-service/client.go
abhishekdwivedi3060: This won't happen in our reconcile flow as client is always created with version. This is only possible when old client is directly from outside

pkg/backup-service/client.go
abhishekdwivedi3060: As discussed, two new public funcs are intentionally exposed for 2 new ABS endpoints

internal/webhook/v1beta1/aerospikebackup_webhook.go
abhishekdwivedi3060: Handled this along with existing known issue of onDemandBackup removal and nil check

tanmayja (dismissed):
Reviewed and tested

tanmayja (approved):

PR #494, opened 2026-04-02
Test:KO-481 remove global storage validation

rpandey-stack:
Added TCs for enhancement/KO-481-remove-global-storage-validation and for enhancement/KO-480-K8-Objects-name-validation


test/envtests/cluster/cluster_webhook_utils.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+const (
+	aerospikeConfigVolName = "aerospike-config-secret"
+	workDir                = "/opt/aerospike"
+	otherStorageClass      = "other-storage-class"
+)
+
+func uniqueNamespacedName(suffix string) types.NamespacedName {
+	name := fmt.Sprintf("ko481-%s-%d", suffix, GinkgoParallelProcess())
```
tanmayja: 1- why ko481, these are generic changes
2- In this files we use GinkgoParallelProcess() in name but not in others. It only helps in when we run the tests in parallel. We should follow same conventions in all file. And I am ok in using this in name if in future we want to run thee tests in parallel.

test/envtests/cluster/cluster_webhook_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	return test.GetNamespacedName(name, testutil.DefaultNamespace)
+}
+
+func storageForDevice(devicePath string) asdbv1.AerospikeStorageSpec {
+	cd := false
+	initM := asdbv1.AerospikeVolumeMethodDeleteFiles
+
+	return asdbv1.AerospikeStorageSpec{
+		BlockVolumePolicy: asdbv1.AerospikePersistentVolumePolicySpec{
+			InputCascadeDelete: &cd,
```
tanmayja: suggestions: we can use ptr.To(false) instead of defining cd variable

test/envtests/cluster/cluster_webhook_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	otherStorageClass      = "other-storage-class"
+)
+
+func uniqueNamespacedName(suffix string) types.NamespacedName {
+	name := fmt.Sprintf("ko481-%s-%d", suffix, GinkgoParallelProcess())
+
+	return test.GetNamespacedName(name, testutil.DefaultNamespace)
+}
+
+func storageForDevice(devicePath string) asdbv1.AerospikeStorageSpec {
```
tanmayja: suggestion:
`getStorageSpecByDevicePath`: Precise and explicit about input and output

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,710 @@
...
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test"
+	testCluster "github.com/aerospike/aerospike-kubernetes-operator/v4/test/cluster"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test/envtests"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/test/testutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	newRackRevision = "v2"
+	clusterName     = "rack-revision-webhook"
```
tanmayja: rename it to `rackRevisionClusterName`.
Package-level const `clusterName` silently shadows local constants in sibling files

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,710 @@
...
+
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"\"vaerospikecluster.kb.io\"",
+							"computed at max rack ID",
+							"exceeds the maximum DNS label length",
+							"of 63 characters").
+						Validate(err)
+
+					Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
```
tanmayja: if we are expecting failure while deploying cluster, then what are we deleting here?

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,710 @@
...
+					cName := test.GetNamespacedName(shortName, clusterNamespacedName.Namespace)
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(cName, 2)
+					aeroCluster.Spec.RackConfig = asdbv1.RackConfig{
+						Namespaces: []string{"test"},
+						Racks:      []asdbv1.Rack{{ID: 1, Revision: "rev-alpha"}}, // 9 chars
+					}
+
+					err := envtests.K8sClient.Create(ctx, aeroCluster)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
```
tanmayja: try to have clean up login in afterEach block.
If any Expect between cluster creation and the final delete panics/fails, Ginkgo catches the panic but the cleanup line is never reached, leaving the cluster in the test environment. This can cause name collisions in subsequent test runs

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,710 @@
...
+
+	AfterEach(func() {
+		aeroCluster := &asdbv1.AerospikeCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      clusterNamespacedName.Name,
+				Namespace: clusterNamespacedName.Namespace,
+			},
+		}
+		// Delete the cluster after each test
+		Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
```
tanmayja: can we use deleteCluster defined in cluster_webhook_utils.go?

test/envtests/cluster/cluster_webhook_utils.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/af89808b4c6c3ec438ad47a70e8fe7d436b4f4e4

test/envtests/cluster/cluster_webhook_utils.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/af89808b4c6c3ec438ad47a70e8fe7d436b4f4e4

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/af89808b4c6c3ec438ad47a70e8fe7d436b4f4e4

Renamed to `rackRevisionClusterName`

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/af89808b4c6c3ec438ad47a70e8fe7d436b4f4e4

Changes done:  deleteCluster is used for both `clusterNamespacedName` and `cName`

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/af89808b4c6c3ec438ad47a70e8fe7d436b4f4e4

You are right, deletion is not needed here. Removed the code line for `DeleteCluster`

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/af89808b4c6c3ec438ad47a70e8fe7d436b4f4e4

Added deletion for both -`clusterNamespacedName` (used in most of the TCs) and `cName`(used in few TCs to have a specific cluster name as per the TCs)

test/envtests/cluster/cluster_webhook_utils.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/2a346608f4171ae47442dfcaa6e9f57f625e7472

Changes done:

- Removed GinkgoParallelProcess() for now. Will add it if it's required in future.

- Updated every test file in test/envtests/cluster to use uniqueNamespacedName() to create NamespacedName for cluster

Makefile
```diff
@@ -172,15 +172,15 @@ restore-test: manifests generate fmt vet setup-envtest ## Run tests.
...
+	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)"; cd $(shell pwd)/test/envtests/cluster && mkdir -p ../../test-results && go run github.com/onsi/ginkgo/v2/ginkgo -r --focus "$(FOCUS)" -coverprofile envcover.out -timeout=1h0m0s --junit-report=../../test-results/junit-envtests-cluster.xml -- . ${ARGS}
 
 .PHONY: env-test-eviction # Run test/envtests/eviction
 env-test-eviction:  fmt vet setup-envtest ## Run tests.
-	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)"; cd $(shell pwd)/test/envtests/eviction && mkdir -p ../../test-results && go run github.com/onsi/ginkgo/v2/ginkgo -r --focus "$(FOCUS)" -coverprofile envcover.out -v -show-node-events -timeout=1h0m0s --junit-report=../../test-results/junit-envtests-eviction.xml -- . ${ARGS}
```
abhishekdwivedi3060, line 183: These changes seem inconsistent like we didn't remove `-show-node-events` param here whereas removed in above 2

test/envtests/cluster/cluster_webhook_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	aerospikeConfigVolName = "aerospike-config-secret"
+	workDir                = "/opt/aerospike"
+	otherStorageClass      = "other-storage-class"
+)
+
+func uniqueNamespacedName(suffix string) types.NamespacedName {
+	name := fmt.Sprintf("envtests-%s", suffix)
+
+	return test.GetNamespacedName(name, testutil.DefaultNamespace)
+}
```
abhishekdwivedi3060, line 32: What's the use of this func? 
We are only adding a envtests- prefix for all, so uniqueness only depends on suffix passed.

test/envtests/cluster/rack_enabled_cluster_webhook_test.go
```diff
@@ -0,0 +1,220 @@
...
+			Context("negative", func() {
+				var nsName types.NamespacedName
+
+				BeforeEach(func() {
+					nsName = uniqueNamespacedName("rackcfg-neg")
+				})
+
+				AfterEach(func() {
+					deleteCluster(ctx, nsName)
+				})
```
abhishekdwivedi3060: Any reason for not moving these Before and After at the top level to keep it common for all?
We'll loose diff naming convention for each context but that fine IMO.

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -0,0 +1,159 @@
...
+				It("allows CREATE when spec.storage has no volumes"+
+					"and validation is satisfied via per-rack InputStorage", func() {
+					aero := testCluster.CreateDummyAerospikeCluster(nsName, 2)
+					fullRack := getStorageSpecForDevice("/test/dev/xvdf")
+					policies := aero.Spec.Storage
+					aero.Spec.Storage = asdbv1.AerospikeStorageSpec{
+						BlockVolumePolicy:      policies.BlockVolumePolicy,
+						FileSystemVolumePolicy: policies.FileSystemVolumePolicy,
+						Volumes:                nil,
+					}
```
abhishekdwivedi3060: We can test with complete removal of spec storage
```suggestion
					rackStorage := getStorageSpecForDevice("/test/dev/xvdf")
					aero.Spec.Storage = asdbv1.AerospikeStorageSpec{}
```

test/envtests/cluster/dynamic_rf_webhook_test.go
```diff
@@ -25,24 +25,23 @@ import (
...
 
 var _ = Describe("AerospikeCluster dynamic replication-factor validation", func() {
 	const (
-		clusterName      = "dynamic-rf-webhook-cluster"
-		maxSchemaRF      = 256
-		rfAboveSchemaMax = 257
-		namespaceName    = "test"
-		namespaceName2   = "test2"
-		namespaceName3   = "test3"
+		dynameicRFClusterName = "dynamic-rf-webhook-cluster"
```
tanmayja: nit: spelling

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+
+					current, err := testCluster.GetCluster(envtests.K8sClient, ctx, ns)
+					Expect(err).ToNot(HaveOccurred())
+
+					// Only touch the rack that actually has pods; rack 3 remains unused by the distribution.
+					current.Spec.RackConfig.Racks[0].Revision = newRackRevision
+					current.Spec.RackConfig.Racks[1].Revision = newRackRevision
+					Expect(envtests.K8sClient.Update(ctx, current)).To(Succeed())
+
+					Expect(testCluster.DeleteCluster(envtests.K8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
```
tanmayja: as per my previous comment, If Expect(envtests.K8sClient.Update(...)).To(Succeed()) panics, the DeleteCluster line is never reached

test/envtests/cluster/cluster_webhook_test.go
```diff
@@ -870,6 +871,21 @@ var _ = Describe("AerospikeCluster validation", func() {
...
+				It("accepts cluster name of exactly 47 chars (projected pod name = 63 chars)", func() {
+					exactName := strings.Repeat("a", 47)
+					cName := test.GetNamespacedName(exactName, clusterNamespacedName.Namespace)
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(cName, 2)
+
+					err := envtests.K8sClient.Create(ctx, aeroCluster)
+
+					defer func() { _ = envtests.K8sClient.Delete(ctx, aeroCluster) }()
+
+					Expect(err).ToNot(HaveOccurred())
```
tanmayja: we should place `defer` after this `expect` statement

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -0,0 +1,159 @@
...
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).To(HaveOccurred())
+
+					// Webhook response validation
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings("\"vaerospikecluster.kb.io\"",
+							"rack storage config cannot be updated",
+							"cannot change volumes old").
+						Validate(err)
+				})
```
abhishekdwivedi3060, line 109: We have similar context in `cluster_webhook_test.go` with similar test-case. Check once

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -0,0 +1,159 @@
...
+						},
+					}
+
+					Expect(envtests.K8sClient.Create(ctx, aero)).To(Succeed())
+				})
+			})
+		})
+	})
+
+	Context("Update validation", func() {
```
abhishekdwivedi3060, line 88: Can we a few more test-cases in update flow around:
1. Should allow removing the spec level storage when rack level storage is used. -> Check old customer flow
2. 2 rack setup - Should allow removing rack level from rack 1 when spec level is also defined
3. 2 rack setup - Should deny removing rack level from rack 1 when spec level storage is not defined.
4. Should allow moving from spec level storage to rack level storage by removing the spec level storage
5. Should allow moving from rack level storage to spec level storage

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+					err := testCluster.DeployCluster(envtests.K8sClient, ctx, aeroCluster)
+					Expect(err).To(HaveOccurred())
+
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"\"vaerospikecluster.kb.io\"",
+							"rack revision \"v_1\" for rack ID 1 is invalid",
+							"must consist of lower case alphanumeric characters or '-'").
+						Validate(err)
+				})
```
abhishekdwivedi3060: Can we move these test-case to a Table-driven test-cases?

test/envtests/cluster/cluster_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/cc23c60d7b893582564d09fe31991e80119a70f7

Changes done:
1. Updated AfterEach() to handle cluster deletion where cluster name is updated dynamically. It is done to keep teardown/cleanup in AfterEach() rather than keeping it at `It `level
2. Removed defer func() from below` It `-

     -  accepts cluster name starting with a lowercase letter
     -  accepts cluster name of exactly 47 chars (projected pod name = 63 chars)

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Missed handling this TC last time. Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/cc23c60d7b893582564d09fe31991e80119a70f7 and https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/4635e1642d15d261cc2503078f038356742982f6

Changes done:

- Removed DeleteCluster from `It`

- replaced `ns `with `cName `as deleteCluster() in AfterEach() is called only for `cName` or `clusterNamespacedName`

test/envtests/cluster/dynamic_rf_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/cc23c60d7b893582564d09fe31991e80119a70f7

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							"\"vaerospikecluster.kb.io\"",
+							"rack revision \"v-\" for rack ID 1 is invalid",
+							"must start and end with an alphanumeric character").
+						Validate(err)
+				})
+
+				// Dot in revision is a silent failure: k8s accepts the STS/pod as a
+				// DNS subdomain, but the dot breaks the pod's DNS label in the FQDN.
```
abhishekdwivedi3060: Comment is not clear.
Dot in the STS name or revision will throw error in the reconcile without the new webhook check. So what does it mean "silent DNS breakage without webhook check"?

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+					aeroCluster.Spec.RackConfig = asdbv1.RackConfig{
+						Namespaces: []string{"test"},
+						Racks:      []asdbv1.Rack{{ID: 1, Revision: "v"}},
+					}
+
+					err := envtests.K8sClient.Create(ctx, aeroCluster)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
+				It("accepts a 2-char revision with a hyphen (e.g. v1)", func() {
```
abhishekdwivedi3060:
```suggestion
				It("accepts a 2-char revision without a hyphen (e.g. v1)", func() {
```
Seems redundant if we are already checking 1 and 3-chart revision

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(cName, 2)
+					aeroCluster.Spec.RackConfig = asdbv1.RackConfig{
+						Namespaces: []string{"test"},
+						Racks:      []asdbv1.Rack{{ID: 1, Revision: strings.Repeat("a", 47)}},
+					}
+
+					err := envtests.K8sClient.Create(ctx, aeroCluster)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
```
abhishekdwivedi3060, line 285: Move to tabel-driven test-cases to avoid code duplication

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+					err := envtests.K8sClient.Create(ctx, aeroCluster)
+					Expect(err).ToNot(HaveOccurred())
+				})
+			})
+		})
+	})
+
+	// Test update validation for rack revision
+	Context("Update validation", func() {
+		Context("spec.rackConfig", func() {
```
abhishekdwivedi3060:
```suggestion
		Context("spec.rackConfig" (rack revision), func() {
```

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+			Context("negative", func() {
```
abhishekdwivedi3060, line 310: For these negative test-cases:
1. Similar test-case are present in `rack_revision_test.go`. Might need to remove them
2. Any reason for getting the diff storage and rackNSOverride in these cases? We can simply DeeCopy the storage config from spec and set in the rack storage and mark the spec storage as empty. Ref the test-cases defined in integration test

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+						WithMessageSubstrings(
+							"\"vaerospikecluster.kb.io\"",
+							"rack storage config cannot be updated",
+							"cannot change volumes",
+						).
+						Validate(err)
+				})
+
+				It("rejects update when rack Revision bumps to a revision"+
+					"already recorded in status with different Storage", func() {
```
abhishekdwivedi3060, line 340:
```suggestion
				It("rejects update when rack Revision bumps to a revision"+
					"already recorded in status with different Storage (revision rollback)", func() {
```

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+		})
+		// Conservative projected-name length bounds are checked on CREATE only.
+		// On UPDATE, the actual pod name (real rack ID, real revision, real
+		// per-rack max ordinal via DistributeItems) is validated as a DNS label
+		// to catch any silent overflow introduced by a revision change.
+		// Revision character validity IS re-checked on UPDATE because an
+		// invalid character (dot, uppercase, underscore) would silently corrupt
+		// the StatefulSet name at runtime regardless of when it was introduced.
+		Context("naming length bounds (CREATE-only projected check; UPDATE uses actual pod name)", func() {
+			Context("negative", func() {
```
abhishekdwivedi3060: Shouldn't these negative and positive contexts go under above "Update Validation" positive and negative context?

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+						WithMessageSubstrings(
+							"\"vaerospikecluster.kb.io\"",
+							"rack revision \"V2\" for rack ID 1 is invalid",
+							"must consist of lower case alphanumeric characters or '-'").
+						Validate(err)
+				})
+
+				// validateActualPodNames finds the longest pod name across all racks.
+				// Even if only one rack's pod name overflows, the whole UPDATE is rejected.
+				It("rejects UPDATE when only one rack's pod name exceeds the DNS label limit (multi-rack)", func() {
```
abhishekdwivedi3060:
```suggestion
				It("rejects UPDATE when only one rack's pod name exceeds the DNS label limit due to rack revision change(multi-rack)", func() {
```

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,706 @@
...
+					Expect(err).ToNot(HaveOccurred())
+
+					current, err := testCluster.GetCluster(envtests.K8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+
+					current.Spec.Size = 2
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).ToNot(HaveOccurred(),
+						"UPDATE with no user-defined racks must not fail on actual pod name check")
+				})
```
abhishekdwivedi3060: Didn't understand these test-cases. What are we exactly testing here? The size is still 2 in update

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,702 @@
...
+
+					current.Spec.RackConfig.Racks[0].Revision = "rev-alpha"
+					err = envtests.K8sClient.Update(ctx, current)
+					Expect(err).ToNot(HaveOccurred(),
+						"revision growth within the DNS label limit must be accepted")
+				})
+
+				It("allows UPDATE when one rack has zero pods under DistributeItems"+
+					"(validateActualPodNames skips rackSize 0)", func() {
+					cName := uniqueNamespacedName("actpod-zero-rack")
```
abhishekdwivedi3060:
```suggestion
					cName = uniqueNamespacedName("actpod-zero-rack")
```

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,702 @@
...
+					Expect(err).ToNot(HaveOccurred())
+
+					// Only touch the rack that actually has pods; rack 3 remains unused by the distribution.
+					current.Spec.RackConfig.Racks[0].Revision = newRackRevision
+					current.Spec.RackConfig.Racks[1].Revision = newRackRevision
+					Expect(envtests.K8sClient.Update(ctx, current)).To(Succeed())
+				})
+
+				It("allows UPDATE when max pod ordinal is greater than zero (validateActualPodNames uses rackSize-1)", func() {
+					cName := uniqueNamespacedName("actpod-ord-gt0")
```
abhishekdwivedi3060:
```suggestion
					cName = uniqueNamespacedName("actpod-ord-gt0")
```

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,702 @@
...
+					current, err := testCluster.GetCluster(envtests.K8sClient, ctx, cName)
+					Expect(err).ToNot(HaveOccurred())
+
+					// Only touch the rack that actually has pods; rack 3 remains unused by the distribution.
+					current.Spec.RackConfig.Racks[0].Revision = newRackRevision
+					current.Spec.RackConfig.Racks[1].Revision = newRackRevision
+					Expect(envtests.K8sClient.Update(ctx, current)).To(Succeed())
+				})
+
+				It("allows UPDATE when max pod ordinal is greater than zero (validateActualPodNames uses rackSize-1)", func() {
```
abhishekdwivedi3060: This seems redundant as this "allows UPDATE when max pod ordinal is greater than zero" is already getting checked in above positive test-cases.

abhishekdwivedi3060 (changes requested):

test/envtests/cluster/cluster_webhook_utils.go
rpandey-stack: The helper isn’t trying to generate random IDs—it centralizes the `envtests-` prefix, `default namespace`, and gives a single place to add parallel-safe or other naming rules later. Uniqueness is enforced by distinct suffixes per cluster; the prefix avoids collisions with non-envtest names in the same cluster.

Example of adding parallel-safe or unique name rules later:
`fmt.Sprintf("envtests-%s-%d", suffix, GinkgoParallelProcess())`

Let me know if this approach looks ok to you!?

test/envtests/cluster/rack_enabled_cluster_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/4abbf467e6b66ed89b060b367e7dbf7540d51c66

test/envtests/cluster/rack_enabled_cluster_webhook_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Context("negative", func() {
+				It("rejects when rack namespace device path is not covered by that rack's InputStorage", func() {
+					aero := testCluster.CreateDummyAerospikeCluster(nsName, 2)
+					s := getStorageSpecForDevice("/wrong/path/not-in-namespace")
+					policies := aero.Spec.Storage
+					aero.Spec.Storage = asdbv1.AerospikeStorageSpec{
+						BlockVolumePolicy:      policies.BlockVolumePolicy,
+						FileSystemVolumePolicy: policies.FileSystemVolumePolicy,
+						Volumes:                nil,
+					}
```
abhishekdwivedi3060: We can simply set it empty. Check wherver possible

test/envtests/cluster/rack_enabled_cluster_webhook_test.go
```diff
@@ -0,0 +1,201 @@
...
+						Racks: []asdbv1.Rack{
+							{ID: 1, Revision: "v1", InputStorage: &s1, InputAerospikeConfig: rackNSOverride("/rack1/xvda")},
+							{ID: 2, Revision: "v1", InputStorage: &s2, InputAerospikeConfig: rackNSOverride("/rack2/xvdb")},
+						},
+					}
+
+					Expect(envtests.K8sClient.Create(ctx, aero)).To(Succeed())
+				})
+
+				It("allows multiple racks each using distinct InputStorage paths consistent "+
```
abhishekdwivedi3060: How is this test-case different from above? We have diff storage paths for each rack in both the cases

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/4a1a40d6525f56cc656377cc1caf7ed8e6e4557e

Removed the TC from `cluster_webhook_test.go` and retained the TC in `storage_webhook_test.go`

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/4a1a40d6525f56cc656377cc1caf7ed8e6e4557e

Added all the TCs mentioned above.

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/4a1a40d6525f56cc656377cc1caf7ed8e6e4557e

test/envtests/cluster/rack_enabled_cluster_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/62d0a452eb6e2dad6e8d0e72a2d5651b03e3a6cd

Modified 2nd TC to _"allows one rack with InputStorage and another rack falling back to spec-level storage"_

test/envtests/cluster/rack_enabled_cluster_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/62d0a452eb6e2dad6e8d0e72a2d5651b03e3a6cd

Updated `rack_enabled_cluster_webhook_test.go` , `storage_webhook_test.go`, `rack_revision_webhook_test.go` wherever it is not needed.

Makefile
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/f3fbe7f9dd27d5e04024eb5b2a27a7d3f756ca61

Removed -show-node-events from `make env-test-eviction` command to reduce console log lines.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Changes done:

**DescribeTable("rejects invalid rack revision on CREATE")**

- Rack revision with uppercase letters

- Rack revision with hyphen

- Rack revision with dot

- Rack revision with underscore


**DescribeTable("accepts valid rack revision on CREATE")**

- Empty rack revision

- 1-char lowercase rack revision

- 2-char rack revision with only digits

- 3-char alphanumeric rack revision

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Modified this TC to "2-char rack revision with only digits" and moved it to DescribeTable("accepts valid rack revision on CREATE")

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Changes done:

- Created table: _DescribeTable("accepts CREATE when a rack revision keeps the projected pod name within limit")_

- Moved TC: _"revision longer than 3 chars when pod name stays within limit"_ and _"revision exactly at the projected pod name boundary (pod name = 63 chars)"_ to this table.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Changes done:
1. Removed TC _"Should reject in-place storage updates with same revision"_ from `rack_revision_test.go`
2. Used DeepCopy approach in all TCs where rackNSOverride was used and removed rackNSOverride usage.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Changes done:

1.  Kept this TC as it is.
2. Added TC: _"rejects rollback of rack revision after storage expansion (revision bump bypass)"_ and removed _"Should reject storage update validation bypass via revision bump and rollback"_ from `rack_revision_test.go`

**Q:** Why above two steps were done?
**Ans:** Both end up hitting the same webhook validation but they differ in the user story (i.e how one get there).
- **TC:** rejects update when rack Revision bumps to a revision already recorded in status with different Storage
  -  **User story:** User tries to apply another v2 with different storage while the recorded v2 in status is still the old one. The webhook should reject that spec update.
  - **Think:** “Don’t rewrite v2 with conflicting storage.”

- **TC:** rejects rollback of rack revision after storage expansion (revision bump bypass)
  - **User story:** After a revision bump that added storage, you try to roll back the revision label to v1 without removing the extra storage — a bypass pattern. The webhook should reject that update.
  - **Think:** “Don’t roll back revision while keeping storage from the newer revision.”

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Here initial intention was to have a separate Context() for K8 object name length related TCs in a separate Context().But this grouping is based on functionality and not based on spec levels. Since we are only following spec level differentiator ( or grouping), now I have moved all negative and positive TCs from `Context("naming length bounds (CREATE-only projected check; UPDATE uses actual pod name)" ` to `Context("spec.rackConfig (rack revision)"` under the Update Validation.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Changes done:

1. Removed this TC as it is duplicate of TC: _"allows UPDATE without re-checking CREATE-only length bounds"_
2. Renamed TC: _"allows UPDATE without re-checking CREATE-only length bounds"_  to _"allows UPDATE without re-applying CREATE-only projected pod name checks"_
3. Corrected cluster size update in above TC.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c
Removed this redundant TC.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: This code line is removed due to above TC removal. Commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

Changes done: 
- TC moved to DescribeTable with modified TC-level comments.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/7fa771b8b773b0708ca151989c3ca0f856d2a61c

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -67,34 +67,18 @@ var _ = Describe("Storage webhook validation", func() {
...
-					aero := testCluster.CreateDummyAerospikeCluster(nsName, 2)
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(nsName, 2)
 					fullRack := getStorageSpecForDevice("/test/dev/xvdf")
-					policies := aero.Spec.Storage
-					aero.Spec.Storage = asdbv1.AerospikeStorageSpec{
-						BlockVolumePolicy:      policies.BlockVolumePolicy,
-						FileSystemVolumePolicy: policies.FileSystemVolumePolicy,
-						Volumes:                nil,
-					}
-					aero.Spec.RackConfig = asdbv1.RackConfig{
```
abhishekdwivedi3060: Need to set the spec storage as empty

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -0,0 +1,653 @@
...
+
+					var st *asdbv1.AerospikeStorageSpec
+					if r0.InputStorage != nil {
+						st = r0.InputStorage.DeepCopy()
+					} else {
+						st = r0.Storage.DeepCopy()
+					}
+
+					patchFirstPVStorageClassSpec(st)
+					r0.InputStorage = st
```
abhishekdwivedi3060:
```suggestion
					patchFirstPVStorageClassSpec(r0.InputStorage))
```

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -440,50 +381,75 @@ var _ = Describe("Rack revision webhook validation", func() {
...
-			Context("positive", func() {
-				It("allows update that changes rack InputStorage when rack Revision changes"+
-					"and status does not imply a conflict", func() {
-					s1 := getStorageSpecForDevice("/r1/a")
+				// Mirrors test/cluster/rack_revision_test.go:
+				// "Should reject storage update validation bypass via revision bump and rollback".
+				// 1) Bump rack revision and expand InputStorage (append volume) — allowed.
+				// 2) Status records baseline v1 storage; spec rolls back to v1 but keeps expanded
+				//    storage — must be rejected (it needs status row so validateRackUpdate can compare).
+				It("rejects rollback of rack revision after storage expansion (revision bump bypass)", func() {
```
abhishekdwivedi3060, line 382: This test-case is same as above 
```
It("rejects update when rack Revision bumps to a revision"+
					"already recorded in status with different Storage"
```
In both the case we are updating the storage (StorageClass vs new PVC). 
I would keep this one and remove the above one

test/envtests/cluster/rack_revision_webhook_test.go
```diff
@@ -587,41 +554,47 @@ var _ = Describe("Rack revision webhook validation", func() {
...
+					// Change size — an otherwise valid update.
+					current.Spec.Size = 3
 					err = envtests.K8sClient.Update(ctx, current)
-					Expect(err).ToNot(HaveOccurred(),
-						"UPDATE with no user-defined racks must not fail on actual pod name check")
+					// The webhook must not surface any CREATE-only naming error on UPDATE.
+					if err != nil {
+						Expect(err).NotTo(MatchError(ContainSubstring("computed at max rack ID")),
+							"CREATE-only pod name length check must not be re-run on UPDATE")
+					}
```
abhishekdwivedi3060: Expect(err).ToNot(HaveOccurred()
Should not return any error at all

abhishekdwivedi3060 (commented):
Overall LGTM.
Left some minor comments.

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/a96ee295c647e03ff5067851fb445e5cc1157e76

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/a96ee295c647e03ff5067851fb445e5cc1157e76

test/envtests/cluster/rack_revision_webhook_test.go
rpandey-stack: Since TCs are being added as per use-case, I think it will be better to have both the test cases even if both TC are hitting the same webhook validation.
More details on use-case: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494#discussion_r3044704377
We can discuss further on this if required.

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed it in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/494/changes/a3604030f39ea032c06e56d8a08f151b132d5083

abhishekdwivedi3060 (approved):

PR #498, opened 2026-04-09
enhancement: Adding support for server 8.1.2. [KO-489]

tanmayja:


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -69,7 +69,19 @@ func (acv *AerospikeClusterCustomValidator) ValidateCreate(_ context.Context, ae
...
+	// For new clusters, promote cgroup-mem-tracking to a hard error and exclude
+	// it from warnings to avoid printing the same message twice.
+	var filteredWarns admission.Warnings
+
+	for _, w := range warns {
+		if strings.HasPrefix(w, "aerospikeConfig.service.cgroup-mem-tracking") {
+			return filteredWarns, fmt.Errorf("%s", w)
+		}
+
+		filteredWarns = append(filteredWarns, w)
```
abhishekdwivedi3060: We are already returning in case of cgroup-mem warning then there is no need to create a new filteredWarns slice

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1122,23 +1146,55 @@ func validateClusterSize(_ logr.Logger, sz int) error {
...
+	if err == nil && cmp >= 0 {
+		serviceConf, ok := configSpec.Value[asdbv1.ConfKeyService].(map[string]interface{})
+		if ok {
+			if msg := cgroupMemTrackingMessage(serviceConf); msg != "" {
+				warnings = append(warnings, msg)
+			}
+		}
+	}
+
+	return warnings, nil
```
abhishekdwivedi3060: This check is only limited to AKO validation. This won't work for cloud related aerospikeConfig validation which are triggered from the validation pkg

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1325,19 +1381,35 @@ func validateNamespaceConfig(
...
+	if err == nil && cmp >= 0 {
+		serviceConf, ok := newConf[asdbv1.ConfKeyService].(map[string]interface{})
+		if ok {
+			if msg := cgroupMemTrackingMessage(serviceConf); msg != "" {
+				warnings = append(warnings, msg)
+			}
+		}
+	}
+
+	return warnings, nil
```
abhishekdwivedi3060: Again this won't run for the validation pkg update aerospikeconfig check

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1122,23 +1146,55 @@ func validateClusterSize(_ logr.Logger, sz int) error {
...
+		return nil, err
 	}
 
-	return validateNamespaceConfig(configSpec.Value, storage)
+	var warnings admission.Warnings
+
+	cmp, err := lib.CompareVersions(version, minVersionForCgroupMemTracking)
+	if err == nil && cmp >= 0 {
+		serviceConf, ok := configSpec.Value[asdbv1.ConfKeyService].(map[string]interface{})
+		if ok {
```
abhishekdwivedi3060: Why not move this version check inside cgroupMemTrackingMessage and rename it?

internal/webhook/v1/utils.go
```diff
@@ -7,6 +7,7 @@ const (
 	baseInitVersion                        = "1.0.0"
 	minInitVersionForDynamicConf           = "2.2.0"
 	minVersionForEnterprisePKIOnlyAuthMode = "8.1.0.0"
+	minVersionForCgroupMemTracking         = "8.1.2.0"
```
abhishekdwivedi3060, line 10: We'll need to update this version everywhere like sample files, test-case, docs etc as new validation will start blocking everything
This is kind of big change wherever we have sample files. Can we revisit the decision about defaulting instead of denying?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -301,10 +316,14 @@ func validate(aslog logr.Logger, cluster *asdbv1.AerospikeCluster) (admission.Wa
 		}
 
 		// Validate common aerospike config schema and fields
-		err = validateAerospikeConfig(aslog, version,
+		var configWarns admission.Warnings
+
+		configWarns, err = validateAerospikeConfig(aslog, version,
```
abhishekdwivedi3060, line 429: Do you know the reason why we have spec level aerospikeConfig update check but not spec level create aerospikeConfig check?

abhishekdwivedi3060 (changes requested):

internal/webhook/v1/aerospikecluster_validating_webhook.go
tanmayja: I purposely did not add the validations in validation pkg
Need to confirm how cloud will react to those warnings or they need to do something in new cluster creation beforehand

internal/webhook/v1/utils.go
tanmayja: There is second commit pending to it with server tag update on tests and elsewhere along with these samples update with flag.
Sure we can discuss futher.

internal/webhook/v1/aerospikecluster_validating_webhook.go
tanmayja: This question I have asked in the past. No specific reason. Lets sync up around this as well.

test/cluster/cluster_test.go
```diff
@@ -1164,7 +1164,8 @@ func deployClusterForMaxIgnorablePods(ctx goctx.Context, clusterNamespacedName t
 // Test cluster deployment with all image post 5.7.0 except the latest version
 func DeployClusterForAllImagesPost570(ctx goctx.Context) {
 	versions := []string{
-		"8.1.0.0", "8.0.0.2", "7.2.0.6", "7.1.0.12", "7.0.0.20", "6.4.0.7", "6.3.0.13", "6.2.0.9", "6.1.0.14", "6.0.0.16",
+		"8.1.1.0", "8.1.0.0", "8.0.0.2", "7.2.0.6", "7.1.0.12", "7.0.0.20",
```
abhishekdwivedi3060: I think we only keep the old major version in the list. 8.1.1 and 8.1.2 are part of same major

test/testutil/doc.go
abhishekdwivedi3060, line 1: ???

test/envtests/cluster/cluster_webhook_test.go
```diff
@@ -1027,4 +1027,112 @@ var _ = Describe("AerospikeCluster validation", func() {
...
+		Context("Deploy validation", func() {
+			Context("with server version >= "+testutil.CgroupMemTrackingServerVersion, func() {
+				FIt("allows create and emits a warning when cgroup-mem-tracking is absent", func() {
+					aeroCluster := testCluster.CreateAerospikeClusterPost640(
+						clusterNamespacedName, 1, testutil.GetEnterpriseImage(testutil.CgroupMemTrackingServerVersion),
+					)
+
+					c, capture, err := envtests.NewClientWithWarningCapture()
+					Expect(err).ToNot(HaveOccurred())
+
```
abhishekdwivedi3060, line 1113: We are defining global clients for other client types, so why not define a similar global client for this along with the warning capture so that it can used throughout envtest

test/envtests/cluster/cluster_webhook_test.go
```diff
@@ -1027,4 +1027,112 @@ var _ = Describe("AerospikeCluster validation", func() {
...
+			if value == nil {
+				delete(serviceConf, "cgroup-mem-tracking")
+			} else {
+				serviceConf["cgroup-mem-tracking"] = value
+			}
+		}
+
+		Context("Deploy validation", func() {
+			Context("with server version >= "+testutil.CgroupMemTrackingServerVersion, func() {
+				FIt("allows create and emits a warning when cgroup-mem-tracking is absent", func() {
```
abhishekdwivedi3060: - Need to follow the context level categorisation like positive/negative
- remove FIT

test/envtests/cluster/cluster_webhook_test.go
```diff
@@ -1027,4 +1027,112 @@ var _ = Describe("AerospikeCluster validation", func() {
...
+
+					Expect(c.Create(ctx, aeroCluster)).ToNot(HaveOccurred())
+					Expect(capture.Warnings).NotTo(ContainElement(ContainSubstring(cgroupWarningSubstr)))
+				})
+			})
+
+			Context("with server version < "+testutil.CgroupMemTrackingServerVersion, func() {
+				It("allows create without warning when cgroup-mem-tracking is absent", func() {
+					aeroCluster := testCluster.CreateAerospikeClusterPost640(
+						clusterNamespacedName, 1, testutil.GetEnterpriseImage(testutil.LatestServerVersion),
```
abhishekdwivedi3060: This latest image is 8.1.2. Should be an older image

abhishekdwivedi3060 (approved):

PR #501, opened 2026-04-13
Fix: validation messages in webhook [KO-484]

tanmayja:


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1682,9 +1684,13 @@ func isEnterprise(image string) bool {
 }
 
 func validateImage(spec *asdbv1.AerospikeClusterSpec) error {
+	if spec.Image == "" {
+		return fmt.Errorf("image cannot be empty")
```
abhishekdwivedi3060: nit:
```suggestion
		return fmt.Errorf("spec.image cannot be empty")
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1682,9 +1684,13 @@ func isEnterprise(image string) bool {
...
+	}
+
 	// Validate image type. Only enterprise and federal image allowed for now.
 	if !isEnterprise(spec.Image) && !asdbv1.IsFederal(spec.Image) {
-		return fmt.Errorf("CommunityEdition Cluster not supported")
+		return fmt.Errorf("image %q is not supported: only enterprise and federal images are allowed", spec.Image)
```
abhishekdwivedi3060: nit 
```suggestion
		return fmt.Errorf("image %q is not supported: only Enterprise and Federal editions are allowed", spec.Image)
```

abhishekdwivedi3060 (approved):

PR #512, opened 2026-05-01
Test: KO-521 SC in-memory support

rpandey-stack:
TCs to validate feat/KO-521-SC-in-memory

### DIFF
--- a/test/cluster/cluster_helper.go
+++ b/test/cluster/cluster_helper.go
@@ -1655,6 +1655,26 @@ func getSCNamespaceConfigWithSet(name, path string) map[string]interface{} {
 	}
 }
 
+// getSCInMemoryNamespaceConfigWithSet returns a strong-consistency namespace with storage-engine memory
+// backed by a persistence device (type + devices only; Aerospike 8.x schema forbids data-size together
+// with devices under storage-engine). devicePath must match a block volume in spec.storage.
+func getSCInMemoryNamespaceConfigWithSet(name, devicePath string) map[string]interface{} {
+	return map[string]interface{}{
+		"name":               name,
+		"replication-factor": 2,
+		"strong-consistency": true,
+		asdbv1.ConfKeyStorageEngine: map[string]interface{}{
+			"type":    "memory",
+			"devices": []interface{}{devicePath},
+		},
+		"sets": []map[string]interface{}{
+			{
+				"name": "testset",
+			},
+		},
+	}
+}
+
 func getNonSCInMemoryNamespaceConfig(name string) map[string]interface{} {
 	return map[string]interface{}{
 		"name":               name,
--- a/test/cluster/cluster_helper.go
+++ b/test/cluster/cluster_helper.go
@@ -2,9 +2,11 @@ package cluster
 
 import (
 	goctx "context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -24,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	as "github.com/aerospike/aerospike-client-go/v8"
+	ast "github.com/aerospike/aerospike-client-go/v8/types"
 	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
 	"github.com/aerospike/aerospike-kubernetes-operator/v4/pkg/utils"
 	"github.com/aerospike/aerospike-kubernetes-operator/v4/test"
@@ -1729,6 +1732,29 @@ func getAeroClusterPVCList(
 	return pvcList.Items, nil
 }
 
+// isAerospikeTransientDataPlaneErr reports errors that often clear after topology or
+// operator "Completed" while the Aerospike client partition map is still warming.
+func isAerospikeTransientDataPlaneErr(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	var ae *as.AerospikeError
+	if errors.As(err, &ae) {
+		rc := ae.ResultCode
+
+		return rc == ast.INVALID_NAMESPACE ||
+			rc == ast.INVALID_CLUSTER_PARTITION_MAP ||
+			rc == ast.SERVER_NOT_AVAILABLE ||
+			rc == ast.NO_AVAILABLE_CONNECTIONS_TO_NODE ||
+			rc == ast.NETWORK_ERROR ||
+			rc == ast.TIMEOUT ||
+			rc == ast.MAX_RETRIES_EXCEEDED
+	}
+
+	return strings.Contains(strings.ToLower(err.Error()), "partition map empty")
+}
+
 func WriteDataToCluster(
 	aeroCluster *asdbv1.AerospikeCluster,
 	k8sClient client.Client,
@@ -1753,12 +1779,27 @@ func WriteDataToCluster(
 			return err
 		}
 
-		if err := asClient.Put(
-			wp, newKey, as.BinMap{
-				binName: binValue,
-			},
-		); err != nil {
-			return err
+		var putErr error
+
+		for attempt := 0; attempt < 90; attempt++ {
+			putErr = asClient.Put(
+				wp, newKey, as.BinMap{
+					binName: binValue,
+				},
+			)
+			if putErr == nil {
+				break
+			}
+
+			if !isAerospikeTransientDataPlaneErr(putErr) {
+				return putErr
+			}
+
+			time.Sleep(2 * time.Second)
+		}
+
+		if putErr != nil {
+			return putErr
 		}
 	}
 
@@ -1791,7 +1832,17 @@ func CheckDataInCluster(
 
 		record, err := asClient.Get(nil, newKey)
 		if err != nil {
-			return nil, nil
+			if errors.Is(err, as.ErrKeyNotFound) {
+				data[ns] = false
+
+				continue
+			}
+
+			return nil, err
+		}
+
+		if record == nil {
+			return nil, fmt.Errorf("nil record from Get for namespace %q", ns)
 		}
 
 		if bin, exists := record.Bins[binName]; exists {
--- a/test/cluster/cluster_helper.go
+++ b/test/cluster/cluster_helper.go
@@ -67,7 +67,6 @@ const (
 )
 
 var aerospikeVolumeInitMethodDeleteFiles = asdbv1.AerospikeVolumeMethodDeleteFiles
-
 var (
 	retryInterval      = time.Second * 30
 	shortRetryInterval = time.Second * 1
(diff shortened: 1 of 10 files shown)


test/cluster/storage_init_test.go
```diff
@@ -730,7 +730,7 @@ func hasData(pod *corev1.Pod, volume *asdbv1.VolumeSpec) bool {
 
 		// Log the error and wait before retrying
 		fmt.Printf("Attempt %d/%d failed: %v. Retrying in %v...\n", i+1, maxRetries, err, retryInterval)
```
tanmayja, line 732: change log message to use hasDataPollInterval instead of retryInterval

test/cluster/large_reconcile_test.go
```diff
@@ -187,8 +189,67 @@ var _ = Describe(
 					},
 				)
 				It(
-					"ColdStart", func() {
-						// Not needed for this, isClusterStable call should fail and this will requeue request.
+					"Cold restart to Completed takes longer after more records are loaded (no secondary indexes)", func() {
```
tanmayja: Is it a SC in-memory cluster?

test/cluster/large_reconcile_test.go
```diff
@@ -252,14 +320,65 @@ func loadDataInCluster(
...
 	}
 
 	fmt.Println("added records")
 
 	return nil
 }
 
+// measureColdRestartToClusterCompleted deletes all Aerospike pods (cold restart), waits for new pods to be
+// Ready, then waits for AerospikeCluster phase Completed. Returns elapsed wall time from first delete.
+func measureColdRestartToClusterCompleted(
```
tanmayja, line 415: This function is deleting pods one after another without waiting for any migration, If you want to check end to end time of cold restart of all pods, you can use AKO provided code restart operation.

test/cluster/strong_consistency_test.go
```diff
@@ -587,6 +587,202 @@ var _ = Describe("SCMode", func() {
...
+
+			Expect(waitForAerospikeCluster(
+				k8sClient, ctx, current, int(current.Spec.Size), retryInterval,
+				getTimeout(current.Spec.Size), []asdbv1.AerospikeClusterPhase{asdbv1.AerospikeClusterCompleted},
+			)).ToNot(HaveOccurred())
+
+			eventuallyAssertKO521DataAfterTopologyChange(clusterNamespacedName, ko521NSName)
+		})
+
+		It("Single pod delete/recreate preserves data", func() {
```
tanmayja, line 694: in above test, we are deleting all pods and check for data. Do we really need this test for single node?

test/cluster/strong_consistency_test.go
```diff
@@ -587,6 +587,202 @@ var _ = Describe("SCMode", func() {
...
+			eventuallyAssertKO521DataAfterTopologyChange(clusterNamespacedName, ko521NSName)
+
+			Expect(scaleUpClusterTest(k8sClient, ctx, clusterNamespacedName, 1)).ToNot(HaveOccurred())
+			eventuallyAssertKO521DataAfterTopologyChange(clusterNamespacedName, ko521NSName)
+
+			Expect(scaleDownClusterTest(k8sClient, ctx, clusterNamespacedName, 1)).ToNot(HaveOccurred())
+			eventuallyAssertKO521DataAfterTopologyChange(clusterNamespacedName, ko521NSName)
+		})
+
+		It("Rack-enabled deployment works for SC in-memory persistent namespace", func() {
```
tanmayja: We could have created a rack-aware cluster as part of one of the above tests, which would have allowed us to skip this separate test.

test/cluster/strong_consistency_test.go
```diff
@@ -587,6 +587,202 @@ var _ = Describe("SCMode", func() {
...
+			}
+
+			Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+			Expect(validateRackEnabledCluster(k8sClient, ctx, clusterNamespacedName)).ToNot(HaveOccurred())
+			eventuallyWaitForKO521RosterReady(clusterNamespacedName, ko521NSName)
+			Expect(WriteDataToCluster(aeroCluster, k8sClient, []string{ko521NSName})).ToNot(HaveOccurred())
+			eventuallyAssertKO521DataAfterTopologyChange(clusterNamespacedName, ko521NSName)
+		})
+
+		It("Invalid in-memory persistence deploy fails without partial long-running resources", func() {
```
tanmayja: "partial long-running resources" is confusing to me, if we are just talking about the device path not available

test/cluster/strong_consistency_test.go
```diff
@@ -587,6 +587,202 @@ var _ = Describe("SCMode", func() {
...
+			if existing == nil {
+				return
+			}
+
+			// Pods may still appear in Status (StatefulSet exists, ASD crash-loops on bad device path).
+			// Deploy failure is that we never reach a completed, healthy cluster.
+			Expect(existing.Status.Phase).ToNot(Equal(asdbv1.AerospikeClusterCompleted))
+		})
+
+		It("Invalid update removing persistence backing is rejected and cluster stays healthy", func() {
```
tanmayja, line 750: is it a webhook test?

test/cluster/large_reconcile_test.go
rpandey-stack: No — this test is not using an SC in-memory cluster.

`createDummyAerospikeCluster()` wires the default namespace with `getSCNamespaceConfig()`, which is strong-consistency on storage-engine.type: device (block device), not `type: memory`.

Should I also add similar TC with SC in-memory cluster?

test/cluster/storage_init_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/512/changes/175765bbb64d2aa9ccc8a92d683e7e240f8c346f

Updated comment as required.

test/cluster/large_reconcile_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/512/changes/175765bbb64d2aa9ccc8a92d683e7e240f8c346f

Extended the existing test with 2nd namespace of type SC in-memory.

test/cluster/large_reconcile_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/512/changes/175765bbb64d2aa9ccc8a92d683e7e240f8c346f

Used AKO provided cold restart operation.

test/cluster/strong_consistency_test.go
rpandey-stack: I think we should have both TCs cause they both stress different failure and recovery shapes i.e all pod deletion leads to cluster restart with cold restart and single pod deletion is more about partial pod loss. Although the backend logic of complete cluster restart and single pod addition to existing cluster might be same, but having both TCs will give us better confidence on what is working.
I am considering these 2 TCs as two different scenarios even if backend logic is same for both.

test/cluster/strong_consistency_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/512/changes/058b28b8e991f6812168663178bf5a3a89a2cfbb

Changes done:
In TC: `It("Rolling restart with batch size and cold restart preserves data")`
Right after DeployCluster succeeds, added:
```
By("Validate rack-aware StatefulSets and pod layout for rack config")
Expect(validateRackEnabledCluster(k8sClient, ctx, clusterNamespacedName)).ToNot(HaveOccurred())
```
Then the existing eventuallyWaitForKO521RosterReady → write → asserts → rolling restart → cold delete-all-pods flow is unchanged.

Removed TC `It("Rack-enabled deployment works for SC in-memory persistent namespace")`

test/cluster/strong_consistency_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/512/changes/058b28b8e991f6812168663178bf5a3a89a2cfbb

Changes done: 
Renamed TC to `Invalid SC in-memory persistence device path rejects deploy and cluster never completes`

test/cluster/strong_consistency_test.go
rpandey-stack: The failed Update is the validating webhook rejecting an invalid SC in-memory spec (persistence removed). But we also assert that the cluster remains Completed/healthy and data is still readable, which is beyond a pure webhook spec test and hence this TC is part of integration test suite.
There is an overlap with webhook test,  but this is added to ensure e2e functionality for "**_webhook reject + no harm_**" on real cluster. Rest all variants of invalid specs are part of envtests.

test/cluster/cluster_helper.go
```diff
@@ -1791,7 +1854,17 @@ func CheckDataInCluster(
 
 		record, err := asClient.Get(nil, newKey)
```
tanmayja, line 1855: can we explore passing the required bin name as third argument in Get function.
Need to understand the current logic and behaviour first.

test/cluster/large_reconcile_test.go
```diff
@@ -10,15 +10,22 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	as "github.com/aerospike/aerospike-client-go/v8"
 	asdbv1 "github.com/aerospike/aerospike-kubernetes-operator/v4/api/v1"
+	"github.com/aerospike/aerospike-kubernetes-operator/v4/pkg/utils"
 	"github.com/aerospike/aerospike-kubernetes-operator/v4/test"
 )
 
+// Second block device for SC in-memory persistence in the PodRestart duration test (alongside /test/dev/xvdf).
```
tanmayja, line 24: I don't really understand this comment.

Couple of suggestions:
1- avoid having unnecessary big variable names. we are in test/cluster/large_reconcile_test.go file, no need to add prefix largeReconcile.
2- if global variable names are clashing with other files and value is same then why not use common.

test/cluster/large_reconcile_test.go
```diff
@@ -187,8 +194,91 @@ var _ = Describe(
...
+						memVol := getStorageVolumeForAerospike("nsmem", largeReconcileSCInMemoryDevicePath)
+						aeroCluster.Spec.Storage.Volumes = append(append(append(
+							[]asdbv1.VolumeSpec{}, vols[0]), memVol), vols[1:]...)
+
+						nsList := aeroCluster.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace].([]interface{})
+						nsList = append(nsList, getSCInMemoryNamespaceConfigWithSet(
+							largeReconcileSCInMemoryNamespace, largeReconcileSCInMemoryDevicePath))
+						aeroCluster.Spec.AerospikeConfig.Value[asdbv1.ConfKeyNamespace] = nsList
+
+						aeroCluster.Spec.AerospikeNetworkPolicy = asdbv1.AerospikeNetworkPolicy{
```
tanmayja, line 214: why are we explicitly setting network policy if this testcase has nothing to do with networking?

PR #514, opened 2026-05-05
Chore: Move validations/defaulting to kubebuilder markers

tanmayja:
Moving only those:

1. which is supported by k8s version 1.23(minimum supported version by AKO) and above.
2. Skipping pattern validation in markers because error returned by pattern validation via markers is not human-readable

Wait for CI before merging.


config/webhook/manifests.yaml
```diff
@@ -24,26 +24,6 @@ webhooks:
     resources:
     - aerospikeclusters
   sideEffects: None
-- admissionReviewVersions:
```
abhishekdwivedi3060, line 27: Need to remove these webhook from helm-charts also

internal/webhook/v1beta1/aerospikebackup_webhook.go
```diff
@@ -39,38 +39,10 @@ import (
...
-
-	for i := range backup.Spec.OnDemandBackups {
-		if backup.Spec.OnDemandBackups[i].Type == "" {
-			backup.Spec.OnDemandBackups[i].Type = asdbv1beta1.FullBackup
-		}
-	}
-
-	return nil
-}
-
```
abhishekdwivedi3060, line 73: Lets only remove the webhook generate marker and keep the code for future ref and interface implementation

test/envtests/cluster/access_control_webhook_test.go
```diff
@@ -79,8 +79,7 @@ func expectDeployFailsACWebhook(ctx context.Context, cluster *asdbv1.AerospikeCl
 	err := envtests.K8sClient.Create(ctx, cluster)
 	Expect(err).To(HaveOccurred())
 
-	args := append([]string{"\"vaerospikecluster.kb.io\""}, subs...)
-	envtests.NewStatusErrorMatcher().WithMessageSubstrings(args...).Validate(err)
+	envtests.NewStatusErrorMatcher().WithMessageSubstrings(subs...).Validate(err)
```
abhishekdwivedi3060, line 171: I think we should at least test these CRD changes in 1.23 k8s version

internal/controller/cluster/client_policy.go
```diff
@@ -287,7 +287,7 @@ func (r *SingleClusterReconciler) appendCACertFromSecret(
 	found := &corev1.Secret{}
 
 	if secretSource.CaCertsSource != nil {
-		//nolint:staticcheck // SA1019: must read deprecated SecretNamespace to resolve secret until field is removed
```
abhishekdwivedi3060, line 290: IMO previous comments were better

abhishekdwivedi3060 (changes requested):

abhishekdwivedi3060 (commented):
1. Shouldn't we add test-cases for confirming the default marker working?
2. What about running these env-tests in 1.23 env-test as well to:
     - confirm these changes works for 1.23
     - catch any future changes to marker validation/defaulting run against our minimum supported verion

test/envtests/cluster/cluster_webhook_test.go
rpandey-stack, line 1: I think now we should also add positive tests verifying that the API server applies the 'default' value after create.
Few of the TCs which we can add: 

- `validationPolicy` default — Create without `validationPolicy`, Get and assert default {`skipWorkDirValidate: false`} 
- Rack ID upper bound — Add `rejects rack ID above MaxRackID (1000001)` with `CRDSchemaErrorPrefix `(integration still has this in rack_management_test.go; envtest only covers negative ID today).
- Rack ID = 0 —> CRD now allows `minimum: 0`; webhook may still reject `DefaultRackID` usage. Add envtest for `{ID: 0} `vs `{ID: DefaultRackID}` 
- Add TC for OutOfRangeRackID
- `dnsPolicy` positive paths — Add:
   - `allows dnsPolicy ClusterFirst`
   - `allows dnsPolicy None when dnsConfig is set`(integration has this incluster_test.go; envtest only has the negative case).

Below TCs can be added in a new file `network_policy_webhook_test.go` :

- Network policy defaults — Create with empty `AerospikeNetworkPolicy` fields, then Get and assert:
   - `access→hostInternal`
   - `alternateAccess→hostExternal`
   - `tlsAccess→hostInternal`
   - `tlsAlternateAccess→hostExternal`
- Invalid network enum at schema level — e.g. invalid `fabric / tlsFabric` type (integration: `network_policy_test.go ``InvalidFabricType`). Good envtest candidate now that enums are on the CRD.

test/envtests/cluster/storage_webhook_test.go
rpandey-stack, line 1: I think now we should also add positive test verifying that the API server applies the 'default' value after create. i.e `storage.cleanupThreads` default = 1 — Create without `cleanupThreads`, Get and assert `spec.storage.cleanupThreads == 1` after create.

helm-charts/aerospike-kubernetes-operator/crds/customresourcedefinition_aerospikebackups.asdb.aerospike.com.yaml
rpandey-stack, line 1: Currently we don't have any `envtests `for backup. Should we add `envtests `to validate it?

helm-charts/aerospike-kubernetes-operator/crds/customresourcedefinition_aerospikerestores.asdb.aerospike.com.yaml
rpandey-stack, line 1: Currently we don't have any `envtests` for restore. Should we add `envtests` to validate it?

rpandey-stack (approved):

abhishekdwivedi3060 (approved):

PR #521, opened 2026-05-20
fix: namespace storage devices/files validations. [KO-535]

tanmayja:
As per JIRA https://aerospike.atlassian.net/browse/KO-535.
This PR is handling following bugs:
Bug#1: rejects in-memory SC namespace when persistence device is not in storage config
Bug#3: rejects update when in-memory SC namespace persistence devices has invalid type.

### DIFF
--- a/pkg/validation/validate.go
+++ b/pkg/validation/validate.go
@@ -284,12 +284,7 @@ func validateNamespaceConfig(
 				)
 			}
 
-			if IsInMemoryNamespace(nsConf) {
-				// storage-engine memory
-				continue
-			}
-
-			if !IsDeviceOrPmemNamespace(nsConf) {
+			if !IsInMemoryNamespace(nsConf) && !IsDeviceOrPmemNamespace(nsConf) {
 				return fmt.Errorf(
 					"%s not supported for namespace %v", asdbv1.ConfKeyStorageEngine, nsConf,
 				)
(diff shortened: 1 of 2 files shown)


pkg/validation/validate.go
```diff
@@ -501,8 +496,20 @@ func ValidateStorageEngineDeviceList(nsConfList []interface{}) (deviceList, file
 		storage := nsConf[asdbv1.ConfKeyStorageEngine].(map[string]interface{})
 
 		if devices, ok := storage["devices"]; ok {
-			for _, d := range devices.([]interface{}) {
-				device := d.(string)
+			deviceSlice, ok := devices.([]interface{})
+			if !ok {
+				return nil, nil, fmt.Errorf(
+					"namespace %s storage devices must be a list, got %T", namespace, devices,
```
abhishekdwivedi3060, line 499: Lets use the existing error msg for consistency similar to validateNamespaceConfig func -> "namespace storage device format not valid" or use the new one everywhere

pkg/validation/validate.go
```diff
@@ -501,8 +496,20 @@ func ValidateStorageEngineDeviceList(nsConfList []interface{}) (deviceList, file
...
+				)
+			}
+
+			for _, d := range deviceSlice {
+				device, ok := d.(string)
+				if !ok {
+					return nil, nil, fmt.Errorf(
+						"namespace %s storage device must be a string, got %T", namespace, d,
```
abhishekdwivedi3060, line 507: same here and wherever applicable

pkg/validation/validate.go
tanmayja: changing existing messages.

PR #534, opened 2026-06-01
fix: prevent data loss during scale-down and failure recovery of failed pods [KO-536]

abhishekdwivedi3060:
- Add deleteLocalStorageOnPodRecovery storage flag (default false) as an explicit opt-in for local PVC deletion during unplanned pod failure recovery; preserves PVCs by default since disk data is intact after aprocess crash
- Thread isFailureRecovery bool through restartPods and deletePodAndEnsureImageUpdated so deleteLocalStorageOnRestart and eviction/node-blocklist triggers are not applied to crashed pods
- Block scale-down of non-ignorable failed pods in scaleDownRack; the old unconditional Insert into ignorablePodNames skipped all safety checks (migration, safe-stop, cascade-delete PVC), causing data loss on any storage type; maxIgnorablePods is now the escape valve
- Use getAllRevisionRackPodList in getIgnorablePods so old-revision pods are included in the maxIgnorablePods budget, preventing premature scale-down of the old STS during rack revision migration


internal/controller/cluster/rack.go
tanmayja, line 930: Do we need to calculate and pass runningPods now?

internal/controller/cluster/pod.go
```diff
@@ -509,10 +513,12 @@ func (r *SingleClusterReconciler) restartPods(
 
 			restartedASDPodNames = append(restartedASDPodNames, pod.Name)
 		case podRestart:
-			if r.isLocalPVCDeletionRequired(rackState, pod) {
+			if r.isLocalPVCDeletionRequired(rackState, pod, isFailureRecovery) {
 				if err := r.deleteLocalPVCs(rackState, pod); err != nil {
 					return common.ReconcileError(err)
 				}
+			} else if isFailureRecovery && len(rackState.Rack.Storage.LocalStorageClasses) != 0 {
```
tanmayja: if we move this print inside isLocalPVCDeletionRequired, it will be printed in both the calls and we won't need to have this code here

internal/controller/cluster/aero_info_calls.go
```diff
@@ -111,6 +127,18 @@ func (r *SingleClusterReconciler) quiescePods(
 		return err
 	}
 
+	// Belt-and-suspenders: the caller (waitForMultipleNodesSafeStopReady) already
```
tanmayja: not needed now

tanmayja (dismissed):

tanmayja (approved):

PR #535, opened 2026-06-03
fix: wipe ACL status from CR during cluster discovery. [KO-546]

tanmayja:


internal/controller/cluster/reconciler.go
```diff
@@ -791,9 +791,50 @@ func (r *SingleClusterReconciler) recoverFailedCreate() error {
 		}
 	}
 
+	// Clear ACL status so that the next reconcile applies credentials fresh
+	// against the default admin account. When STSes and PVCs are deleted above,
+	// all user/credential data on the Aerospike nodes is wiped. If the stale
+	// ACL status is left intact, the operator would attempt to authenticate with
+	// those old credentials on the newly recreated nodes, causing info commands
+	// to fail.
+	if err := r.clearAerospikeAccessControlStatus(); err != nil {
```
abhishekdwivedi3060, line 806: - What happens when failed pod PVCs are not deleted due to `CascadeDelete: false` and we clean the status ACL as prt of recovery? I guess we'll face similar issue of Invalid credentials
- There is no guarantee that the PVCs of failed pods are always deleted even when `CascadeDelete: true` because PVC deletion logic depends on  podNames. We are deleting STSs first and then checking pods list, so pod list can be partial/empty. Can you confirm this flow?

internal/controller/cluster/reconciler.go
```diff
@@ -786,14 +792,55 @@ func (r *SingleClusterReconciler) recoverFailedCreate() error {
...
+		return err
+	}
+
+	newAeroCluster.Status.AerospikeAccessControl = nil
+
+	if err := r.patchStatus(newAeroCluster); err != nil {
+		return fmt.Errorf("error clearing access control status: %w", err)
+	}
+
+	r.aeroCluster.Status.AerospikeAccessControl = nil
```
abhishekdwivedi3060: I think we are already handling this in patchStatus func

abhishekdwivedi3060 (approved):

PR #538, opened 2026-06-09
Test: KO-536 Failed pod handling

rpandey-stack:
PR created to added envtests and integration test for https://aerospike.atlassian.net/browse/KO-536

### DIFF
--- a/test/cluster/batch_scaledown_pods_test.go
+++ b/test/cluster/batch_scaledown_pods_test.go
@@ -130,24 +130,6 @@ var _ = Describe("BatchScaleDown", func() {
 			Expect(err).To(HaveOccurred())
 		})
 
-		It("Should allow scale-down when the failed pod is covered by maxIgnorablePods", func() {
-			aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
-			Expect(err).ToNot(HaveOccurred())
-
-			failedPodName := clusterName + "-0-3"
-
-			By("Marking the pod as failed")
-			Expect(markPodAsFailed(ctx, k8sClient, failedPodName, namespace)).ToNot(HaveOccurred())
-
-			By("Setting maxIgnorablePods=1 and triggering scale-down")
-
-			maxIgnorable := intstr.FromInt32(1)
-			aeroCluster.Spec.RackConfig.MaxIgnorablePods = &maxIgnorable
-			aeroCluster.Spec.Size--
-
-			Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
-		})
-
 		It("Should block scale-down when pod other than the pod to be scaled down is in failed and "+
 			"maxIgnorablePods is not set", func() {
 			aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
@@ -190,6 +172,67 @@ var _ = Describe("BatchScaleDown", func() {
 
 			Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
 		})
+	})
+
+	Context("PVC outcome on failed-pod scale-down follows cascadeDelete", func() {
+		DescribeTable("",
+			func(cascadeDelete bool, expectPVCsDeleted bool) {
+				aeroCluster := createNonSCDummyAerospikeCluster(clusterNamespacedName, 4)
+				if cascadeDelete {
+					aeroCluster.Spec.Storage.BlockVolumePolicy.InputCascadeDelete = &cascadeDeleteTrue
+					aeroCluster.Spec.Storage.FileSystemVolumePolicy.InputCascadeDelete = &cascadeDeleteTrue
+				}
+
+				Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+				failedPodName := clusterName + "-0-3"
+
+				By("Capturing scaled-down pod PVCs before failure")
+
+				pvcBefore, err := getPVCClaimUIDsForPod(ctx, k8sClient, failedPodName, namespace)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(pvcBefore).ToNot(BeEmpty())
+
+				By("Marking the pod as failed")
+				Expect(markPodAsFailed(ctx, k8sClient, failedPodName, namespace)).ToNot(HaveOccurred())
+
+				aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Setting maxIgnorablePods=1 and triggering scale-down")
+
+				maxIgnorable := intstr.FromInt32(1)
+				aeroCluster.Spec.RackConfig.MaxIgnorablePods = &maxIgnorable
+				aeroCluster.Spec.Size--
+
+				Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+				By("Waiting for scale-down to complete")
+
+				err = waitForAerospikeCluster(
+					k8sClient, ctx, aeroCluster, int(aeroCluster.Spec.Size),
+					retryInterval, getTimeout(aeroCluster.Spec.Size),
+					[]asdbv1.AerospikeClusterPhase{asdbv1.AerospikeClusterCompleted},
+				)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Asserting scaled-down failed pod PVC outcome")
+
+				if expectPVCsDeleted {
+					Expect(waitForPVCClaimsDeleted(ctx, k8sClient, pvcBefore, namespace)).To(Succeed())
+				} else {
+					Expect(validatePVCClaimUIDsUnchanged(ctx, k8sClient, pvcBefore, namespace)).To(Succeed())
+				}
+			},
+			Entry(
+				"Should allow scale-down and preserve scaled-down failed pod PVCs when cascadeDelete=false",
+				false, false,
+			),
+			Entry(
+				"Should allow scale-down and delete scaled-down failed pod PVCs when cascadeDelete=true",
+				true, true,
+			),
+		)
 	})
 
 	// TODO: Do we need to add all the invalid operation test-cases here?
--- a/test/cluster/batch_scaledown_pods_test.go
+++ b/test/cluster/batch_scaledown_pods_test.go
@@ -173,24 +173,22 @@ var _ = Describe("BatchScaleDown", func() {
 			Expect(err).To(HaveOccurred())
 		})
 
-		Context("maxIgnorablePods as percentage on scale-down with a failed pod", func() {
-			It("allows scale-down with a failed pod when maxIgnorablePods is a percentage covering one pod", func() {
-				aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
-				Expect(err).ToNot(HaveOccurred())
+		It("Should allow scale-down with a failed pod when maxIgnorablePods is a percentage covering one pod", func() {
+			aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
 
-				failedPodName := clusterName + "-0-3"
+			failedPodName := clusterName + "-0-3"
 
-				By("Marking the pod as failed")
-				Expect(markPodAsFailed(ctx, k8sClient, failedPodName, namespace)).ToNot(HaveOccurred())
+			By("Marking the pod as failed")
+			Expect(markPodAsFailed(ctx, k8sClient, failedPodName, namespace)).ToNot(HaveOccurred())
 
-				By("Setting maxIgnorablePods=34% and triggering scale-down")
+			By("Setting maxIgnorablePods=34% and triggering scale-down")
 
-				maxIgnorable := intstr.FromString("34%")
-				aeroCluster.Spec.RackConfig.MaxIgnorablePods = &maxIgnorable
-				aeroCluster.Spec.Size--
+			maxIgnorable := intstr.FromString("34%")
+			aeroCluster.Spec.RackConfig.MaxIgnorablePods = &maxIgnorable
+			aeroCluster.Spec.Size--
 
-				Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
-			})
+			Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
 		})
 	})
 
--- a/test/cluster/batch_scaledown_pods_test.go
+++ b/test/cluster/batch_scaledown_pods_test.go
@@ -172,6 +172,26 @@ var _ = Describe("BatchScaleDown", func() {
 			)
 			Expect(err).To(HaveOccurred())
 		})
+
+		Context("maxIgnorablePods as percentage on scale-down with a failed pod", func() {
+			It("allows scale-down with a failed pod when maxIgnorablePods is a percentage covering one pod", func() {
+				aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+				Expect(err).ToNot(HaveOccurred())
+
+				failedPodName := clusterName + "-0-3"
+
+				By("Marking the pod as failed")
+				Expect(markPodAsFailed(ctx, k8sClient, failedPodName, namespace)).ToNot(HaveOccurred())
+
+				By("Setting maxIgnorablePods=34% and triggering scale-down")
+
+				maxIgnorable := intstr.FromString("34%")
+				aeroCluster.Spec.RackConfig.MaxIgnorablePods = &maxIgnorable
+				aeroCluster.Spec.Size--
+
+				Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+			})
+		})
 	})
 
 	// TODO: Do we need to add all the invalid operation test-cases here?
--- a/test/cluster/cluster_helper.go
+++ b/test/cluster/cluster_helper.go
@@ -1888,25 +1888,29 @@ func extractPodPVC(ctx goctx.Context, k8sClient client.Client, pod *corev1.Pod) 
 	pvcUIDMap := make(map[string]types.UID)
 
 	for idx := range pod.Spec.Volumes {
-		if pod.Spec.Volumes[idx].PersistentVolumeClaim != nil {
-			pvcUIDMap[pod.Spec.Volumes[idx].PersistentVolumeClaim.ClaimName] = ""
+		claim := pod.Spec.Volumes[idx].PersistentVolumeClaim
+		if claim == nil {
+			continue
 		}
-	}
 
-	for p := range pvcUIDMap {
+		claimName := claim.ClaimName
+		if _, seen := pvcUIDMap[claimName]; seen {
+			continue
+		}
+
 		pvc := &corev1.PersistentVolumeClaim{}
-		if err := k8sClient.Get(ctx, test.GetNamespacedName(p, pod.Namespace), pvc); err != nil {
+		if err := k8sClient.Get(ctx, test.GetNamespacedName(claimName, pod.Namespace), pvc); err != nil {
 			return nil, err
 		}
 
-		pvcUIDMap[p] = pvc.UID
+		pvcUIDMap[claimName] = pvc.UID
 	}
 
 	return pvcUIDMap, nil
 }
 
-// pvcClaimUIDsForPod returns PVC claim names and UIDs for local volumes on the named pod.
-func pvcClaimUIDsForPod(
+// getPVCClaimUIDsForPod returns PVC claim names and UIDs for local volumes on the named pod.
+func getPVCClaimUIDsForPod(
 	ctx goctx.Context, k8sClient client.Client, podName, ns string,
 ) (map[string]types.UID, error) {
 	pod := &corev1.Pod{}
@@ -1917,8 +1921,8 @@ func pvcClaimUIDsForPod(
 	return extractPodPVC(ctx, k8sClient, pod)
 }
 
-// pvcClaimsEventuallyNotFound waits until each PVC in claims is absent from the API (deleted).
-func pvcClaimsEventuallyNotFound(
+// waitForPVCClaimsDeleted waits until each PVC in claims is absent from the API (deleted).
+func waitForPVCClaimsDeleted(
 	ctx goctx.Context, k8sClient client.Client, claims map[string]types.UID, ns string,
 ) error {
 	for claimName := range claims {
@@ -1948,8 +1952,8 @@ func pvcClaimsEventuallyNotFound(
 	return nil
 }
 
-// pvcClaimsKeepStableUIDs asserts each PVC still exists with the same UID as before.
-func pvcClaimsKeepStableUIDs(
+// validatePVCClaimUIDsUnchanged asserts each PVC still exists with the same UID as before.
+func validatePVCClaimUIDsUnchanged(
 	ctx goctx.Context, k8sClient client.Client, claims map[string]types.UID, ns string,
 ) error {
 	for claimName, wantUID := range claims {
--- a/test/cluster/cluster_helper.go
+++ b/test/cluster/cluster_helper.go
@@ -63,7 +63,10 @@ const (
 	binValue = "binValue"
 )
 
-var aerospikeVolumeInitMethodDeleteFiles = asdbv1.AerospikeVolumeMethodDeleteFiles
+var (
+	aerospikeVolumeInitMethodDeleteFiles = asdbv1.AerospikeVolumeMethodDeleteFiles
+	aerospikeVolumeInitMethodDD          = asdbv1.AerospikeVolumeMethodDD
+)
 
 var (
 	retryInterval      = time.Second * 30
@@ -1558,6 +1561,7 @@ func aerospikeClusterCreateUpdate(
 func getBasicStorageSpecObject() asdbv1.AerospikeStorageSpec {
 	storage := asdbv1.AerospikeStorageSpec{
 		BlockVolumePolicy: asdbv1.AerospikePersistentVolumePolicySpec{
+			InputInitMethod:    &aerospikeVolumeInitMethodDD,
 			InputCascadeDelete: &cascadeDeleteFalse,
 		},
 		FileSystemVolumePolicy: asdbv1.AerospikePersistentVolumePolicySpec{
@@ -1877,6 +1881,89 @@ func markPodAsFailed(ctx goctx.Context, k8sClient client.Client, name, namespace
 	pod.Spec.Containers[0].Image = wrongImage
 
 	return k8sClient.Update(ctx, pod)
+}
+
+// extractPodPVC returns a map of PVC claim name to UID for volumes attached to the pod.
+func extractPodPVC(ctx goctx.Context, k8sClient client.Client, pod *corev1.Pod) (map[string]types.UID, error) {
+	pvcUIDMap := make(map[string]types.UID)
+
+	for idx := range pod.Spec.Volumes {
+		if pod.Spec.Volumes[idx].PersistentVolumeClaim != nil {
+			pvcUIDMap[pod.Spec.Volumes[idx].PersistentVolumeClaim.ClaimName] = ""
+		}
+	}
+
+	for p := range pvcUIDMap {
+		pvc := &corev1.PersistentVolumeClaim{}
+		if err := k8sClient.Get(ctx, test.GetNamespacedName(p, pod.Namespace), pvc); err != nil {
+			return nil, err
+		}
+
+		pvcUIDMap[p] = pvc.UID
+	}
+
+	return pvcUIDMap, nil
+}
+
+// pvcClaimUIDsForPod returns PVC claim names and UIDs for local volumes on the named pod.
+func pvcClaimUIDsForPod(
+	ctx goctx.Context, k8sClient client.Client, podName, ns string,
+) (map[string]types.UID, error) {
+	pod := &corev1.Pod{}
+	if err := k8sClient.Get(ctx, types.NamespacedName{Name: podName, Namespace: ns}, pod); err != nil {
+		return nil, err
+	}
+
+	return extractPodPVC(ctx, k8sClient, pod)
+}
+
+// pvcClaimsEventuallyNotFound waits until each PVC in claims is absent from the API (deleted).
+func pvcClaimsEventuallyNotFound(
+	ctx goctx.Context, k8sClient client.Client, claims map[string]types.UID, ns string,
+) error {
+	for claimName := range claims {
+		n := claimName
+
+		err := wait.PollUntilContextTimeout(ctx, retryInterval, getTimeout(4), true,
+			func(ctx goctx.Context) (done bool, err error) {
+				pvc := &corev1.PersistentVolumeClaim{}
+
+				getErr := k8sClient.Get(ctx, test.GetNamespacedName(n, ns), pvc)
+				if getErr != nil {
+					if k8serrors.IsNotFound(getErr) {
+						return true, nil
+					}
+
+					return false, getErr
+				}
+
+				return false, nil
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("PVC %s should be deleted after scale-down: %w", n, err)
+		}
+	}
+
+	return nil
+}
+
+// pvcClaimsKeepStableUIDs asserts each PVC still exists with the same UID as before.
+func pvcClaimsKeepStableUIDs(
+	ctx goctx.Context, k8sClient client.Client, claims map[string]types.UID, ns string,
+) error {
+	for claimName, wantUID := range claims {
+		pvc := &corev1.PersistentVolumeClaim{}
+		if err := k8sClient.Get(ctx, test.GetNamespacedName(claimName, ns), pvc); err != nil {
+			return fmt.Errorf("get PVC %s: %w", claimName, err)
+		}
+
+		if pvc.UID != wantUID {
+			return fmt.Errorf("PVC %s UID changed (%s -> %s)", claimName, wantUID, pvc.UID)
+		}
+	}
+
+	return nil
 }
 
 func checkClientConnection(
--- a/test/cluster/cluster_test.go
+++ b/test/cluster/cluster_test.go
@@ -899,6 +899,7 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
 		"UpdateClusterWithMaxIgnorablePodAndFailedPod", func() {
 			BeforeEach(
 				func() {
+					validateMinNodeCountOrSkip(ctx, 6, "maxIgnorablePods rack operations with failed pod in another rack")
 					deployClusterForMaxIgnorablePods(ctx, clusterNamespacedName, 4)
 				},
 			)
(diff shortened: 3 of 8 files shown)


test/envtests/cluster/storage_webhook_test.go
```diff
@@ -367,6 +531,22 @@ var _ = Describe("Storage webhook validation", func() {
...
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(nsName, 2)
+					aeroCluster.Spec.Storage.DeleteLocalStorageOnPodRecovery = ptr.To(false)
+					aeroCluster.Spec.Storage.LocalStorageClasses = []string{testutil.StorageClass}
+					Expect(envtests.K8sClient.Create(ctx, aeroCluster)).To(Succeed())
+
+					current, err := testCluster.GetCluster(envtests.K8sClient, ctx, nsName)
+					Expect(err).ToNot(HaveOccurred())
+
+					current.Spec.Storage.DeleteLocalStorageOnPodRecovery = ptr.To(true)
+					current.Spec.Storage.LocalStorageClasses = []string{testutil.StorageClass}
```
abhishekdwivedi3060: nit: already set above

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -326,9 +463,35 @@ var _ = Describe("Storage webhook validation", func() {
...
+					envtests.NewStatusErrorMatcher().
+						WithMessageSubstrings(
+							testutil.WebhookErrorPrefix,
+							"forceBlockFromRoster cannot be used together with maxIgnorablePods",
+						).
+						Validate(err)
+				})
+			})
+		})
+
```
abhishekdwivedi3060: Seems unrelated to this PR, please pull changes in a diff PR

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/e23463d7c10bf06f8a3fbb65bca353595f77d85c

It was caused due to code conflict.

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/25a5fdef397da367b50dc5b4755472c13a54aa63

Added this TC in a separate PR: https://github.com/aerospike/aerospike-kubernetes-operator/pull/541

test/cluster/batch_scaledown_pods_test.go
```diff
@@ -172,6 +172,26 @@ var _ = Describe("BatchScaleDown", func() {
 			)
 			Expect(err).To(HaveOccurred())
 		})
+
+		Context("maxIgnorablePods as percentage on scale-down with a failed pod", func() {
```
abhishekdwivedi3060: nit: No need to a separate context here

test/cluster/k8snode_block_list_test.go
```diff
@@ -217,6 +197,10 @@ func validatePVCDeletion(ctx context.Context, pvcUIDMap map[string]types.UID, sh
 		)
 
 		if err := k8sClient.Get(ctx, pvcNamespacesName, pvc); err != nil {
+			if shouldDeletePVC && apierrors.IsNotFound(err) {
+				continue
+			}
```
abhishekdwivedi3060, line 203: What's the reason for this check? PVC should always be present because a STS replica creates all required PVCs if not found.

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -142,7 +143,16 @@ var _ = Describe("Storage webhook validation", func() {
...
+				// Allow create: delete-local recovery false with localStorageClasses set (no conflict).
+				It("allows deploy when deleteLocalStorageOnPodRecovery is false with localStorageClasses set", func() {
+					aeroCluster := testCluster.CreateDummyAerospikeCluster(nsName, 2)
+					aeroCluster.Spec.Storage.DeleteLocalStorageOnPodRecovery = ptr.To(false)
+					aeroCluster.Spec.Storage.LocalStorageClasses = []string{testutil.StorageClass}
+
+					Expect(envtests.K8sClient.Create(ctx, aeroCluster)).To(Succeed())
+				})
+
+				It("allows deploy when deleteLocalStorageOnPodFailureRecovery is false "+
```
abhishekdwivedi3060:
```suggestion
				It("allows deploy when deleteLocalStorageOnPodRecovery is false "+
```

test/envtests/cluster/storage_webhook_test.go
```diff
@@ -326,7 +336,7 @@ var _ = Describe("Storage webhook validation", func() {
 			})
 		})
 
-		Context("spec.storage deleteLocalStorageOnPodRecovery validation", func() {
+		Context("spec.storage deleteLocalStorageOnPodFailureRecovery validation", func() {
```
abhishekdwivedi3060: Revert this and fix the new field name at other places wherever applicable

test/cluster/rack_revision_test.go
```diff
@@ -244,6 +246,76 @@ var _ = Describe(
...
+							),
+							Entry(
+								"Should delete failed pod local PVCs during rack revision when rack InputStorage sets "+
+									"deleteLocalStorageOnPodRecovery true and per-rack cascade delete is enabled "+
+									"(PVC removal on old-revision teardown follows cascadeDelete policy)",
+								true,
+								"Creating cluster with per-rack InputStorage (delete on failure recovery + cascade delete)",
+								true,
+							),
+							Serial,
```
abhishekdwivedi3060, line 318: Rack revision doesn't depend on deleteLocalStorageOnPodRecovery at all as old revision failed are not recovered at all. These test-cases seem redundant

test/cluster/rack_revision_test.go
```diff
@@ -244,6 +246,76 @@ var _ = Describe(
...
+								)
+								Expect(err).ToNot(HaveOccurred())
+
+								err = validateRackEnabledCluster(k8sClient, ctx, clusterNamespacedName)
+								Expect(err).ToNot(HaveOccurred())
+
+								err = validateRackRevisionCleanup(k8sClient, ctx, aeroCluster, []int{1}, versionV1)
+								Expect(err).ToNot(HaveOccurred())
+
+								Expect(validatePVCDeletion(ctx, oldPVC, expectPVCDeleted)).ToNot(HaveOccurred())
```
abhishekdwivedi3060, line 301: This check is going to fail because old STS pods are gone so corresponding PVCs will also get deleted

test/cluster/cluster_test.go
```diff
@@ -990,6 +990,115 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+			It("blocks scale-down when a pending pod consumes maxIgnorablePods=1 and another pod is failed", func() {
+				nodeList, lerr := test.GetNodeList(ctx, k8sClient)
+				Expect(lerr).ToNot(HaveOccurred())
+
+				n := utils.Len32(nodeList.Items)
+				if n < 2 {
+					Skip("requires at least 2 nodes for pending+failed budget test")
+				}
+
+				aeroCluster := createNonSCDummyAerospikeCluster(clusterNamespacedName, n)
```
abhishekdwivedi3060, line 1003: - We need to set MultiPodPerHost to false because we are limiting the number of pods to the number of k8s nodes
- Need to use deployClusterForMaxIgnorablePods func because that func handles everything related to MultiPodPerHost: false

test/cluster/cluster_test.go
```diff
@@ -990,6 +990,115 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+				for i := range podList.Items {
+					p := &podList.Items[i]
+					if p.Status.Phase == v1.PodRunning && utils.IsPodRunningAndReady(p) {
+						victim = p.Name
+
+						break
+					}
+				}
+
+				Expect(victim).ToNot(BeEmpty(), "expected a running pod to mark failed")
```
abhishekdwivedi3060: We can consider the ordinal "-0" pod for simplification

test/cluster/cluster_test.go
```diff
@@ -990,6 +990,115 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+
+				Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+				Expect(markPodAsFailed(ctx, k8sClient, clusterName+"-1-0", clusterNamespacedName.Namespace)).ToNot(HaveOccurred())
+				Expect(markPodAsFailed(ctx, k8sClient, clusterName+"-1-1", clusterNamespacedName.Namespace)).ToNot(HaveOccurred())
+
+				aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+				Expect(err).ToNot(HaveOccurred())
+
+				aeroCluster.Spec.Size--
```
abhishekdwivedi3060: Decrease by 2 so that rack 1 scale down is attempted by AKO

test/cluster/local_pvc_delete_test.go
```diff
@@ -166,6 +200,58 @@ var _ = Describe(
...
+				Entry("deletes removed pod PVCs on scale-down when cascadeDelete=true even if "+
+					"deleteLocalStorageOnPodRecovery=true",
+					true, true),
+				Entry(
+					"preserves removed pod PVCs on scale-down when cascadeDelete=false even if "+
+						"deleteLocalStorageOnPodRecovery=true",
+					false, false),
+			)
+		})
+
```
abhishekdwivedi3060: IMO these test-cases are kind of redundant. deleteLocalStorageOnPodRecovery is applicable for failed pod recovery and in scale down operation there is no pod recovery. Pod is going to go away. Only thing to confirm is PVC is deleted or retained based on `cascadeDelete` and I think we might already be having cascadeDelete related test-cases. Confirm this

test/cluster/local_pvc_delete_test.go
```diff
@@ -349,6 +435,173 @@ var _ = Describe(
...
+			// rack overrides deleteLocalStorageOnPodRecovery while global is false;
+			// only PVCs whose storage class is listed on that rack's localStorageClasses are deleted on recovery.
+			It("Should delete only rack-local-class PVCs on failed-pod recovery when rack overrides "+
+				"deleteLocalStorageOnPodRecovery and global deleteLocalStorageOnPodRecovery is false", func() {
+				aeroCluster := createDummyAerospikeCluster(clusterNamespacedName, 4)
+				aeroCluster.Spec.Storage.DeleteLocalStorageOnRestart = ptr.To(true)
+				aeroCluster.Spec.Storage.DeleteLocalStorageOnPodRecovery = ptr.To(false)
+				aeroCluster.Spec.Storage.LocalStorageClasses = []string{storageClass}
+
+				altSC := os.Getenv("AEROSPIKE_CLUSTER_TEST_ALT_STORAGE_CLASS")
```
abhishekdwivedi3060: Where are we setting this env?

test/cluster/local_pvc_delete_test.go
```diff
@@ -349,6 +435,173 @@ var _ = Describe(
...
+				Expect(err).ToNot(HaveOccurred())
+
+				failedPodName := clusterName + "-1-0"
+				Expect(markPodAsFailed(ctx, k8sClient, failedPodName, namespace)).ToNot(HaveOccurred())
+
+				aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+				Expect(err).ToNot(HaveOccurred())
+
+				aeroCluster.Spec.PodSpec.AerospikeObjectMeta = asdbv1.AerospikeObjectMeta{
+					Labels: map[string]string{"test-label": "int-fph-039-040-rack-local-select"},
```
abhishekdwivedi3060: nit: use readable test-label values wherever applicable

test/cluster/local_pvc_delete_test.go
```diff
@@ -349,6 +435,173 @@ var _ = Describe(
...
+
+					aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+
+					aeroCluster.Spec.PodSpec.AerospikeObjectMeta = asdbv1.AerospikeObjectMeta{
+						Labels: map[string]string{"test-label": "int-fph-041-mixed-rack"},
+					}
+					Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+					validateClusterPVCDeletion(ctx, oldPvcInfoPerPod, clusterName+"-2-0")
```
abhishekdwivedi3060, line 477: I think "-1-1" pod's PVC also won't be deleted as rack 1 has no DeleteLocalStorageOnRestart flag

test/cluster/local_pvc_delete_test.go
```diff
@@ -349,6 +435,173 @@ var _ = Describe(
...
+				Expect(err).ToNot(HaveOccurred())
+
+				aeroCluster.Spec.PodSpec.AerospikeObjectMeta = asdbv1.AerospikeObjectMeta{
+					Labels: map[string]string{"test-label": "int-fph-039-040-rack-local-select"},
+				}
+				Expect(updateCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+				localSCs := sets.New(storageClass)
+				Expect(validateClusterPVCMetaAfterRollingRestart(ctx, oldPvcMetaPerPod, localSCs)).To(Succeed())
+			})
```
abhishekdwivedi3060: This test-case seems overly complicated. Can we simplify this?
- We can use the "local-ssd" name for rack 1 storage volumes
- We can use the validateClusterPVCDeletion func for PVC deletion validation
Overall, we can merge this test-case with the below test-case at line 489.

test/cluster/local_pvc_delete_test.go
```diff
@@ -349,6 +435,173 @@ var _ = Describe(
...
+					}
+
+					oldPVC, err := pvcClaimUIDsForPod(ctx, k8sClient, failedPodName, namespace)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(oldPVC).ToNot(BeEmpty())
+
+					Expect(updateClusterWithTO(k8sClient, ctx, aeroCluster, 20*time.Minute)).ToNot(HaveOccurred())
+
+					Expect(validatePVCDeletion(ctx, oldPVC, true)).To(Succeed())
+				})
```
abhishekdwivedi3060: Seems duplicate of It("Should delete local PVCs for a failed pod during rolling restart when "+
				"deleteLocalStorageOnPodRecovery is set",

test/cluster/local_pvc_delete_test.go
```diff
@@ -349,6 +435,173 @@ var _ = Describe(
...
+					data, derr := CheckDataInCluster(ac, k8sClient, []string{"test"})
+					if derr != nil || data == nil {
+						return false
+					}
+
+					return data["test"]
+				}).WithTimeout(5*time.Minute).WithPolling(retryInterval).Should(BeTrue(),
+					"written bin should remain readable after failed-pod recovery with deleteLocalStorageOnPodRecovery=false")
+			})
+
```
abhishekdwivedi3060: Can we not merge the data integrity part with any existing test-case like It("Should NOT delete local PVCs for a failed pod during rolling restart even if deleteLocalStorageOnRestart is set"

test/cluster/local_pvc_delete_test.go
```diff
@@ -362,6 +615,132 @@ var _ = Describe(
...
+			if pvc.UID != meta.UID {
+				return fmt.Errorf("pod %s PVC %s (storageClass=%q) should be preserved; UID changed %s -> %s",
+					podName, claimName, meta.StorageClass, meta.UID, pvc.UID)
+			}
+		}
+	}
+
+	return nil
+}
```
abhishekdwivedi3060: I think these funcs will be redundant after review comments

abhishekdwivedi3060 (changes requested):

test/cluster/batch_scaledown_pods_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/0b02aa700fcbb215a418f854c1e93a60a3a57d4d

test/cluster/rack_revision_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/0a467a6dc1a3b9a0564f6763ee8bc79505ec57e3

Changes done: 
- **DescribeTable title**
Now: "failed pod PVCs after rack revision: per-rack InputStorage InputCascadeDelete policy" so it matches what actually varies.

- **Table function**
Parameter is now `inputCascadeDelete` bool instead of `deleteOnRecovery` bool.

- **Per-rack InputStorage**
     - DeleteLocalStorageOnPodRecovery is ptr.To(false) for every row (with a short comment that rack-revision old-revision teardown does not use this flag).
     - LocalStorageClasses unchanged.
     - InputCascadeDelete on block + filesystem policies is set only when inputCascadeDelete is true (same as before for the “PVCs deleted” row).

- **Entry descriptions / By text**
Updated to talk about InputCascadeDelete enabled vs default (off), not deleteLocalStorageOnPodRecovery.

- **Comment above the table**
States that teardown follows cascade-delete policy, not DeleteLocalStorageOnPodRecovery.

Not removing this TC as redundant because this TC deals with `createDummyClusterWithRackRevision → failed pod → rack revision bump → validateRackRevisionCleanup → validatePVCDeletion`, whereas other TCs deals only  with `per-rack InputStorage`, `MaxIgnorablePods`, and `InputCascadeDelete ` and validation for rack revision cleanup and PVC deletion are not present. Hence this TC is different from others.

test/cluster/k8snode_block_list_test.go
rpandey-stack: - We’re not asserting that no PVCs exist for the StatefulSet — we’re asserting that the specific PVCs recorded in `oldPvcInfo` are gone when local deletion is expected.
- A StatefulSet may recreate volumes later, but that’s orthogonal: `Get` on the old object returning `NotFound` is the signal that deletion happened; without this `if` branch, we’d fail the test by treating a correct `NotFound` as a generic `Get` error.

It is true that STS always creates PVC but that's for steady state with a running replica and volume templates, but 
1. after cluster logic deletes local PVCs, the old PersistentVolumeClaim resource really can disappear, and,
2. there can be a short window where the old name is NotFound before anything new exists — the test should accept NotFound only in the `shouldDeletePVC` path, which this does.

To clarify this, added a comment above the `if `condition using commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/a198b61549f91fc78265ad38e9deae6225470453

test/cluster/rack_revision_test.go
rpandey-stack: For this, a check is added for `NotFound` in `validatePVCDeletion()` in `test/cluster/k8snode_block_list_test.go` 

Line#199-203

> if err := k8sClient.Get(ctx, pvcNamespacesName, pvc); err != nil {
> 			// when expecting local PVC deletion, NotFound means the old PVC was removed; treat as success
> 			if shouldDeletePVC && apierrors.IsNotFound(err) {
> 				continue
> 			}

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/664c6e6de4ccd7e4fcbe89c9f4a7d95e5d1a513c

Changes done at all places wherever applicable.

test/cluster/cluster_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/5cd3a7b7b69b5528a2db2bf8d90db274a2f5d1f2

Changes done: 
- MultiPodPerHost is set to false.
- Added a short comment that `deployClusterForMaxIgnorablePods` is intentionally not used here because that topology breaks this assertion.

Changes not done:
- Did not use `deployClusterForMaxIgnorablePods()` because it builds a _2-rack + bar volume_ cluster. For this scenario, reconciliation could still reach a state where waitForAerospikeCluster treats as valid (size/spec alignment + phase InProgress or Error) within the 2m window, so the poll returned nil. The test expects err != nil (timeout / no such “valid” stuck state), so it failed with “Expected an error… Got: nil”.

test/cluster/cluster_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/5cd3a7b7b69b5528a2db2bf8d90db274a2f5d1f2

test/cluster/cluster_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/5cd3a7b7b69b5528a2db2bf8d90db274a2f5d1f2

test/envtests/cluster/storage_webhook_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/664c6e6de4ccd7e4fcbe89c9f4a7d95e5d1a513c

test/cluster/local_pvc_delete_test.go
rpandey-stack: - Rack-1 InputStorage is a full `DeepCopy` of cluster storage, so rack 1 still has `deleteLocalStorageOnRestart=true`; only recovery-related fields differ from global i.e `deleteLocalStorageOnPodRecovery `is set to true from false.
- …-1-1 is a healthy pod: PVC deletion on rollout follows the planned restart branch and `deleteLocalStorageOnRestart`, not “missing” rack-level restart.
- …-2-0 is failed: the failure recovery branch uses `deleteLocalStorageOnPodRecovery` only (global false on that rack), so we skip it in `validateClusterPVCDeletion`.

Changes done in code to make it more explicit:
  - Add a short comment above `validateClusterPVCDeletion` stating that rack-1 InputStorage is a full storage copy so `deleteLocalStorageOnRestart` remains true for non-failed rack-1 pods.
- Assert …-1-1 pod explicitly (`validatePVCDeletion` for that pod expecting deletion)

Commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/e9517c1cd51263e9401cc99e161b13c9503671ce

test/cluster/local_pvc_delete_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/e9517c1cd51263e9401cc99e161b13c9503671ce

combined TC: `Should delete only rack-local-class PVCs on failed-pod recovery when rack overrides deleteLocalStorageOnPodRecovery and global deleteLocalStorageOnPodRecovery is false` with TC: `Should apply rack InputStorage deleteLocalStorageOnPodRecovery per rack when global is false`

test/cluster/local_pvc_delete_test.go
rpandey-stack: Addressed in commit-id: e9517c1

- combined this TC with  TC: `Should apply rack InputStorage deleteLocalStorageOnPodRecovery per rack when global is false`
- Updated test-label value to `rack-override-delete-local-pvc` in TC: "Should apply rack InputStorage deleteLocalStorageOnPodRecovery per rack when global is false"

test/cluster/local_pvc_delete_test.go
rpandey-stack: Addressed in commit-id: e9517c1

combined this TC with TC: "Should apply rack InputStorage deleteLocalStorageOnPodRecovery per rack when global is false". So this comment is not applicable anymore.

test/cluster/local_pvc_delete_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/3a47315e6e3de6aaa673d2374371fcfa039ec30f

Merged both the data integrity TCs into existing TCs so that there is no need to have a separate Context() for data integrity test:

- TC "_Should keep previously written Aerospike records when deleteLocalStorageOnPodRecovery is false"_ merged into TC _It("Should NOT delete local PVCs for a failed pod during rolling restart even if deleteLocalStorageOnRestart is set"_ . This TC is further renamed to - It("Should NOT delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is false and keep Aerospike data (RF=1)"

- TC _"Should delete failed pod local PVCs when deleteLocalStorageOnPodRecovery is true during rolling restart"_ is merged into TC: _It("Should delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is set"_

test/cluster/local_pvc_delete_test.go
rpandey-stack: Addressed in commit-id: 3a47315

Merged it with TC: It("Should delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is set"

test/cluster/local_pvc_delete_test.go
rpandey-stack: Yes these functions are removed after addressing other review comments.

test/cluster/local_pvc_delete_test.go
rpandey-stack: Agree that `deleteLocalStorageOnPodRecovery` is not part of scale-down semantics. The intent here was to guard the `failed-pod` + `maxIgnorablePods` scale-down path and ensure PVC outcome still follows `InputCascadeDelete`. I can remain the TCs to make it more relevant to the TC intent. May be something like - 
- "deletes removed pod PVCs on scale-down when cascadeDelete=true"
- "preserves removed pod PVCs on scale-down when cascadeDelete=false"

In `cluster_storage_cleanup_test.go`, there are TCs related PVC deletion/retained based on `cascadeDelete` but those doesn't use `maxIgnorablePods`

test/cluster/cluster_helper.go
```diff
@@ -1879,6 +1883,89 @@ func markPodAsFailed(ctx goctx.Context, k8sClient client.Client, name, namespace
 	return k8sClient.Update(ctx, pod)
 }
 
+// extractPodPVC returns a map of PVC claim name to UID for volumes attached to the pod.
+func extractPodPVC(ctx goctx.Context, k8sClient client.Client, pod *corev1.Pod) (map[string]types.UID, error) {
```
tanmayja, line 1887: Do we need two separate loops here? It looks like we can fetch the PVC and populate the UID in a single pass while still deduplicating by PVC name. That would make the flow a bit simpler and avoid iterating over the map a second time.

test/cluster/cluster_helper.go
```diff
@@ -1879,6 +1883,89 @@ func markPodAsFailed(ctx goctx.Context, k8sClient client.Client, name, namespace
...
+		}
+
+		pvcUIDMap[p] = pvc.UID
+	}
+
+	return pvcUIDMap, nil
+}
+
+// pvcClaimUIDsForPod returns PVC claim names and UIDs for local volumes on the named pod.
+func pvcClaimUIDsForPod(
```
tanmayja: IMO function names should generally start with a verb, especially when they perform an action, validation, or wait operation. Names like pvcClaimsEventuallyNotFound() read more like a condition than a function call.

Please go through all function names once.

test/cluster/cluster_test.go
```diff
@@ -990,6 +990,108 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+
+	Context(
+		"Pending pod consumes maxIgnorablePods=1 so scale-down blocks with a separate failed pod",
+		func() {
+			It("blocks scale-down when a pending pod consumes maxIgnorablePods=1 and another pod is failed", func() {
+				nodeList, lerr := test.GetNodeList(ctx, k8sClient)
+				Expect(lerr).ToNot(HaveOccurred())
+
+				n := utils.Len32(nodeList.Items)
+				if n < 2 {
```
tanmayja: For the test **"Should allow rack deletion with failed pods in a different rack"**, we require at least a 6-node cluster; otherwise, the test will fail.

However, currently we're skipping the test when the cluster has fewer than 2 nodes. Let's keep the behavior consistent—either skip the test when the required node count is not met in both places, or allow it to fail in both cases.

test/cluster/cluster_test.go
```diff
@@ -990,6 +990,108 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+				aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+				Expect(err).ToNot(HaveOccurred())
+
+				maxIgnorable := intstr.FromInt32(1)
+				aeroCluster.Spec.RackConfig.MaxIgnorablePods = &maxIgnorable
+				aeroCluster.Spec.Size--
+
+				Expect(k8sClient.Update(ctx, aeroCluster)).ToNot(HaveOccurred())
+
+				err = waitForAerospikeCluster(
```
tanmayja, line 1071: can we validate that scale-down did not go through?
waitForAerospikeCluster can get failed due to other reasons as well.

tanmayja (commented):
Could you point me to a testcase where 
1- we are failing rf number of nodes 
2-doing some rolling restart let's say vertical scaling. 
3- saving the data by not deleting local pvc
4- bringing failed pods back to running state
5- check data recovered

test/cluster/cluster_helper.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/fa1b7bcb33e3440fb3b56d10aa814f5e86f97e01

Changes done:
`pvcClaimUIDsForPod` → `getPVCClaimUIDsForPod`,
`pvcClaimsEventuallyNotFound` → `waitForPVCClaimsDeleted`,
`pvcClaimsKeepStableUIDs` → `validatePVCClaimUIDsUnchanged`.
Other `extract`* / `validate`* helpers already follow this pattern, so I kept scope to the new PVC helpers and their call sites.

test/cluster/cluster_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/fa1b7bcb33e3440fb3b56d10aa814f5e86f97e01

I added a shared `validateMinNodeCountOrSkip()` helper and use explicit Skip when the Kubernetes cluster doesn’t meet the topology minimum for each scenario:
- 6 nodes for `UpdateClusterWithMaxIgnorablePodAndFailedPod` (2-rack deploy + rack delete / failed-pod recovery),
- 2 nodes for the slimmer pending+failed scale-down block case. `It("blocks scale-down when a pending pod consumes maxIgnorablePods=1 and another pod is failed")`

**Why skip instead of fail on undersized CI clusters:**
- These tests depend on environment capacity (schedulable nodes, pending pods), not operator logic under test.
- A hard fail on a small K8 cluster will look like a regression bug instead of environment capacity issue and can waste triage time.
- Skip documents the precondition in the report (requires at least N nodes: …) so CI stays green on small envs while still running on adequately sized ones.
- Same pattern as other integration tests in this suite that Skip when prerequisites aren’t met.
- Minimums differ by design (6 vs 2) because the topologies differ; the consistent part is enforce via skip, not comment-only vs implicit failure.

test/cluster/cluster_helper.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/fa1b7bcb33e3440fb3b56d10aa814f5e86f97e01

Simplified it as suggested.

test/cluster/cluster_test.go
rpandey-stack: Addressed in commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/fa1b7bcb33e3440fb3b56d10aa814f5e86f97e01

Changes done:
`assertScaleDownBlocked() `helper created to validate scale-down did not complete: status size and pod count stay at their pre-request baselines while spec is lowered.

_It takes three pre-scale-down baselines:_
- specSizeBefore
- statusSizeBefore
- podCountBefore

_Assertions:_
- Consistently → Status.Size stays at statusSizeBefore
- Consistently → pod count stays at podCountBefore
- Eventually → Spec.Size < specSizeBefore, pod count unchanged, pending/failed pods still present

_Pending + failed pod test:_

- After Spec.Size++, waits for pod count == spec before pending-pod check
- Captures specSizeBeforeScaleDown, statusSizeBeforeScaleDown, podCountBeforeScaleDown from observed state before decrement
- Passes all three into assertScaleDownBlocked

`waitForAerospikeCluster` timeout is still kept as supplemental.

rpandey-stack:
> Could you point me to a testcase where 1- we are failing rf number of nodes 2-doing some rolling restart let's say vertical scaling. 3- saving the data by not deleting local pvc 4- bringing failed pods back to running state 5- check data recovered

1- we are failing rf number of nodes
Closest match: 
- _It(blocks scale-down when two pods are failed on the same rack with maxIgnorablePods=1)_ in `cluster_test.go`
- _It("Should NOT delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is false and keep Aerospike data (RF=1)")_ in `local_pvc_delete_test.go`

2-doing some rolling restart let's say vertical scaling.
Rolling restart cases:
- _It("Should NOT delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is false and keep Aerospike data (RF=1)")_ in `local_pvc_delete_test.go`
- _It("Should NOT delete local PVCs for a failed pod during image upgrade even if deleteLocalStorageOnRestart is set")_ in `local_pvc_delete_test.go`
- Vertical scaling scenario is not included to test failed pod recovery during rolling restart 

3- saving the data by not deleting local pvc
- _It("Should NOT delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is false and keep Aerospike data (RF=1")_ in `local_pvc_delete_test.go`
- It("Should NOT delete local PVCs for a failed pod during rolling restart even if deleteLocalStorageOnRestart is set") in `local_pvc_delete_test.go`

4- bringing failed pods back to running state

- _It("Should allow rack deletion with failed pods in different rack")_ in `cluster_test.go`
- _It("Should NOT delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is false and keep Aerospike data (RF=1)")_ in `local_pvc_delete_test.go` 
   - recovery is implicit via successful updateClusterWithTO + readable data; no explicit pod Ready check 

5- check data recovered
- _It("Should NOT delete local PVCs for a failed pod during rolling restart when deleteLocalStorageOnPodRecovery is false and keep Aerospike data (RF=1)")_ in `local_pvc_delete_test.go` 
@tanmayja

test/cluster/local_pvc_delete_test.go
rpandey-stack: Commit-id: https://github.com/aerospike/aerospike-kubernetes-operator/pull/538/changes/b162866d21deb7a4ec817b7ac5d3be2b5a498037

Changes in `local_pvc_delete_test.go:`

- Removed this complete `DescribeTable` from `local_pvc_delete_test.go`

Changes in `batch_scaledown_pods_test.go`:

- Removed the duplicate I_t("Should allow scale-down when the failed pod is covered by maxIgnorablePods")_ with inline PVC checks.
- Added Context("PVC outcome on failed-pod scale-down follows cascadeDelete") with a DescribeTable that:
   - Deploys per entry (with optional cascadeDelete=true)
   - Runs failed pod + maxIgnorablePods=1 + scale-down
   - Waits for completion
   - Asserts PVC preserved (cascadeDelete=false) or deleted (cascadeDelete=true)

abhishekdwivedi3060 (approved):

PR #552, opened 2026-07-01
fix: sidecars failure handling. [KO-539]

tanmayja:


internal/controller/cluster/aero_info_calls.go
```diff
@@ -245,18 +251,19 @@ func (r *SingleClusterReconciler) newPodsHostConnWithOption(pods []corev1.Pod, i
...
 				r.Log.Info(
-					"Ignoring info call on non-running Pod", "pod", utils.GetNamespacedName(pod),
+					"Ignoring info call on pod with non-running server container", "pod", utils.GetNamespacedName(pod),
 				)
 
 				continue
 			}
 
-			return nil, fmt.Errorf("status for Pod %s: not ready", utils.GetNamespacedNameString(pod))
+			return nil, fmt.Errorf("pod %s server container is not running", utils.GetNamespacedNameString(pod))
```
abhishekdwivedi3060, line 268: Let's use the new standard here

internal/controller/cluster/aero_info_calls.go
```diff
@@ -35,22 +35,28 @@ import (
...
 // ------------------------------------------------------------------------------------
 
-// waitForMultipleNodesSafeStopReady waits until the input pods are safe to stop,
-// skipping pods that are not running and present in ignorablePodNames for stability check.
-// The ignorablePodNames is the list of failed or pending pods that are either::
-// 1. going to be deleted eventually and are safe to ignore in stability checks
-// 2. given in ignorePodList by the user and are safe to ignore in stability checks
+// waitForMultipleNodesSafeStopReady waits until the input pods are safe to stop.
+// ignorablePodNames are pods whose Aerospike server is unreachable and are
+// skipped for cluster-operation queries (host connections, roster, quiesce).
```
abhishekdwivedi3060:
```suggestion
// skipped from cluster-operation queries (host connections, roster, quiesce).
```

internal/controller/cluster/statefulset.go
```diff
@@ -246,6 +246,81 @@ func (r *SingleClusterReconciler) deleteSTS(ctx context.Context, st *appsv1.Stat
 	return r.Delete(ctx, st)
 }
 
+func (r *SingleClusterReconciler) waitForAerospikeServerReady(
```
abhishekdwivedi3060: 1. Can we rename it to convey something like that this func checks server container readiness across all replicas of a STS? 
2. Update the logging and error as per new format

internal/controller/cluster/aero_info_calls.go
```diff
@@ -35,22 +35,28 @@ import (
...
+	// Wait for all non-ignorable pods to have their Aerospike server containers
+	// ready before making any cluster-level info calls. This replaces the old
+	// waitForAllSTSToBeReady pre-check (which required full pod readiness
+	// including sidecars). Server-only readiness is sufficient here — sidecar
+	// failures do not prevent the server from accepting info calls. The wait
+	// uses the same blocking-retry semantics (up to 18×10s) so that a pod which
+	// was just restarted in a previous batch has time to bring its server up
+	// before we attempt the migration/quiesce checks.
+	if err := r.waitForAllAerospikeServersReady(ctx, ignorablePodNames); err != nil {
+		return common.ReconcileError(fmt.Errorf("wait for all Aerospike server containers to be ready: %w", err))
```
abhishekdwivedi3060: nit:
```suggestion
		return common.ReconcileError(fmt.Errorf("wait for Aerospike server containers across all StatefulSets to be ready: %w", err))
```

internal/controller/cluster/aero_info_calls.go
```diff
@@ -245,18 +251,19 @@ func (r *SingleClusterReconciler) newPodsHostConnWithOption(pods []corev1.Pod, i
...
-		if !utils.IsPodRunningAndReady(pod) {
+		// Only the Aerospike server container needs to be running to accept info calls.
+		// Sidecar failures do not prevent the server from being reachable.
+		if !utils.IsAerospikeServerReady(pod) {
 			if ignorablePodNames.Has(pod.Name) {
-				// This pod is not running and ignorable.
+				// This pod's aerospike server is not running and it is marked ignorable.
 				r.Log.Info(
-					"Ignoring info call on non-running Pod", "pod", utils.GetNamespacedName(pod),
+					"Ignoring info call on pod with non-running server container", "pod", utils.GetNamespacedName(pod),
```
abhishekdwivedi3060: Lets use the new convention here and all other places.
```suggestion
					"Ignoring info call on Pod with non-running server container", "pod", utils.GetNamespacedName(pod),
```

internal/controller/cluster/reconciler.go
```diff
@@ -960,43 +890,62 @@ func (r *SingleClusterReconciler) handleClusterDeletion(ctx context.Context, fin
...
 
-	return false, common.ReconcileSuccess()
+	if inGracePeriod {
+		r.Log.Info("Pods are failed but within grace period, requeueing...")
+		return false, common.ReconcileRequeueAfter(asdbv1.RequeueIntervalSeconds10)
+	}
+
+	// All pods have hard-failed and status is empty — the cluster failed during
+	// its initial create and needs to be recovered.
+	return true, common.ReconcileError(r.recoverFailedCreate(ctx))
```
abhishekdwivedi3060: Why does it have to be inside common.ReconcileError always?
It should be decided based on the error returned from r.recoverFailedCreate(ctx) func

internal/controller/cluster/pod.go
```diff
@@ -1222,29 +1296,6 @@ func (r *SingleClusterReconciler) getClusterPodList(ctx context.Context) (
 	return podList, nil
 }
 
-func (r *SingleClusterReconciler) isAnyPodInImageFailedState(podList []*corev1.Pod, ignorablePodNames sets.Set[string],
```
abhishekdwivedi3060, line 1225: Need to confirm this

pkg/utils/pod.go
```diff
@@ -102,34 +116,60 @@ func CheckPodFailedWithGrace(pod *corev1.Pod, allowGrace bool) PodState {
...
+		}
+	}
+
+	// Inspect regular containers: always check the aerospike server container;
+	// check sidecars only when includeSidecarFailures is true.
+	for idx := range pod.Status.ContainerStatuses {
+		container := &pod.Status.ContainerStatuses[idx]
+		if container.Name != asdbv1.AerospikeServerContainerName && !includeSidecarFailures {
+			continue
+		}
```
abhishekdwivedi3060: We can follow the previous approach of collecting all the required container status based on flags and then do required checks instead of duplicating the checks in 2 loops

pkg/utils/pod.go
```diff
@@ -64,33 +64,47 @@ func CheckPodFailedWithGrace(pod *corev1.Pod, allowGrace bool) PodState {
...
 	}
 
-	// Check for container-level failures if no pod-level failure found
-	if failureReason == "" {
-		failureReason = checkContainerFailures(pod)
+	return PodState{
+		State:  PodFailed,
+		Reason: failureReason,
 	}
+}
```
abhishekdwivedi3060: This code block can be extracted in a common func -> getPodState(failureReason, allowGrace)

pkg/utils/pod.go
```diff
@@ -64,33 +64,47 @@ func CheckPodFailedWithGrace(pod *corev1.Pod, allowGrace bool) PodState {
...
 
-	// If no failure found, pod is healthy
+// CheckServerFailedWithGrace checks if the pod's Aerospike server container (or
+// a blocking init container) has a terminal failure, optionally within a grace
+// period. Sidecar container failures are intentionally ignored — use
+// CheckSidecarFailedWithGrace for those.
```
abhishekdwivedi3060: Remove this line

pkg/utils/pod.go
```diff
@@ -64,33 +64,47 @@ func CheckPodFailedWithGrace(pod *corev1.Pod, allowGrace bool) PodState {
...
+// Returns PodState containing:
+// - State: PodHealthy, PodFailedInGrace, or PodFailed
+// - Reason: Human-readable description of failure (empty if healthy)
+func CheckServerFailedWithGrace(pod *corev1.Pod, allowGrace bool) PodState {
```
abhishekdwivedi3060, line 73: This is exactly same as CheckPodFailedWithGrace func except the includeSidecarFailures func param.
Can we not write a common private func for all the internal logic and then have 2 top level wrapper funcs for diff use-case.

internal/controller/cluster/statefulset.go
```diff
@@ -246,6 +246,81 @@ func (r *SingleClusterReconciler) deleteSTS(ctx context.Context, st *appsv1.Stat
...
+			time.Sleep(podStatusRetryInterval)
+		}
+
+		if !isReady {
+			return fmt.Errorf(
+				"server container in pod %s did not become ready. Status: %v",
+				utils.NamespacedName(st.Namespace, podName), pod.Status.ContainerStatuses,
+			)
+		}
+	}
```
abhishekdwivedi3060, line 320: Should we not check STS replicas here similar to waitForSTSToBeReady func?

internal/controller/cluster/statefulset.go
```diff
@@ -1082,21 +1157,25 @@ func updateSTSContainers(
...
-func (r *SingleClusterReconciler) waitForAllSTSToBeReady(
+// waitForAllAerospikeServersReady waits for the Aerospike server container's
+// readiy state to be true on every non-ignorable pod across all cluster
+// StatefulSets. It does NOT require sidecars or other containers to be ready.
+// Pods whose names are in ignorablePodNames are skipped entirely; sidecar-failed
+// pods are NOT skipped because their server containers are still reachable.
+func (r *SingleClusterReconciler) waitForAllAerospikeServersReady(
 	ctx context.Context, ignorablePodNames sets.Set[string]) error {
-	r.Log.Info("Waiting for all cluster STSs to be ready")
+	r.Log.Info("Waiting for all Aerospike server containers to be ready")
```
abhishekdwivedi3060:
```suggestion
	r.Log.Info("Waiting for Aerospike server containers across all STSs to be ready")
```

internal/controller/cluster/statefulset.go
```diff
@@ -1082,21 +1157,25 @@ func updateSTSContainers(
...
+		allRackIdentifiers.Insert(utils.GetRackIdentifier(rack.ID, rack.Revision))
 	}
 
-	// Check for newly added racks also because we do not check for these racks just after they are added
-	specRacks := r.aeroCluster.Spec.RackConfig.Racks
-	for idx := range specRacks {
-		allRackIdentifiers.Insert(utils.GetRackIdentifier(specRacks[idx].ID, specRacks[idx].Revision))
+	for idx := range r.aeroCluster.Spec.RackConfig.Racks {
+		rack := r.aeroCluster.Spec.RackConfig.Racks[idx]
+		allRackIdentifiers.Insert(utils.GetRackIdentifier(rack.ID, rack.Revision))
```
abhishekdwivedi3060, line 1226:
```suggestion
	for idx := range r.aeroCluster.Status.RackConfig.Racks {
		rack := &r.aeroCluster.Status.RackConfig.Racks[idx]
		allRackIdentifiers.Insert(utils.GetRackIdentifier(rack.ID, rack.Revision))
}

	for idx := range r.aeroCluster.Spec.RackConfig.Racks {
		rack := &r.aeroCluster.Spec.RackConfig.Racks[idx]
		allRackIdentifiers.Insert(utils.GetRackIdentifier(rack.ID, rack.Revision))
```

internal/controller/cluster/statefulset.go
```diff
@@ -1082,21 +1157,25 @@ func updateSTSContainers(
 	return finalContainers
 }
 
-func (r *SingleClusterReconciler) waitForAllSTSToBeReady(
+// waitForAllAerospikeServersReady waits for the Aerospike server container's
+// readiy state to be true on every non-ignorable pod across all cluster
```
abhishekdwivedi3060:
```suggestion
// ready state to be true on every non-ignorable pod across all cluster
```

internal/controller/cluster/rack.go
```diff
@@ -2235,62 +2237,123 @@ func (r *SingleClusterReconciler) handleFailedPodsInRack(
...
+		r.Log.Info("Pods still not ready after sidecar reconcile, requeuing",
 			"rackID", rackState.Rack.ID, "rackRevision", rackState.Rack.Revision,
-			"failedPods", getPodNames(failedPods))
-		// Requeue after 1 second to fetch latest CR object with updated pod status
-		return common.ReconcileRequeueAfter(1)
+			"notReadyPods", getPodNames(notReadyPods))
 	}
 
-	return common.ReconcileSuccess()
+	return common.ReconcileRequeueAfter(1)
```
abhishekdwivedi3060: For sidecar failure scenarios where there is no change in sidecar config but some change in the aerospikeConfig, won't this ReconcileRequeueAfter(1) be too aggressive? This will be infinite unnecessary if sidecar config is not updated. We should think of exponential backupoff here

internal/controller/cluster/rack.go
abhishekdwivedi3060: This check was initially added for overall failed pod considering server failed.
Do we need to run setMigrateFillDelay for only sidecar failure pods?

internal/controller/cluster/rack.go
```diff
@@ -794,28 +816,17 @@ func (r *SingleClusterReconciler) upgradeRack(
 			continue
 		}
 
-		if r.isPodUpgraded(pod) {
+		if r.isPodOnDesiredImage(pod, false) {
```
abhishekdwivedi3060, line 920: Reason for this change?
IIRC, previous check was saving us from going to the next batch if the previous batch pods didn't come up

internal/controller/cluster/rack.go
```diff
@@ -1348,6 +1345,11 @@ func (r *SingleClusterReconciler) getRollingRestartInfo(
 		}
 	}
 
+	if serverFailedPodsNames.Len() > 0 {
```
abhishekdwivedi3060, line 1475: Better to add comment for this

internal/controller/cluster/pod.go
```diff
@@ -175,6 +175,12 @@ func (r *SingleClusterReconciler) getRollingRestartTypeMap(
 			return nil, nil, err
 		}
 
+		// A quick restart only rolls config dynamically; a server-failed pod
+		// needs a full pod restart to recover the crashed server process.
+		if restartType == quickRestart && serverFailedPodNames.Has(pods[idx].Name) {
```
abhishekdwivedi3060: Q: Wouldn't Dynamic update restartType for a server failed Pod would also require PodRestart?

internal/controller/cluster/pod.go
```diff
@@ -188,17 +194,18 @@ func (r *SingleClusterReconciler) getRollingRestartTypePod(
 ) (RestartType, error) {
 	restartType := noRestart
 
-	// AerospikeConfig nil means status not updated yet
-	if r.IsStatusEmpty() {
+	// If this pod has no status entry yet it was just created with the current
+	// config — no restart is needed and the hash fields would all be empty
+	// strings, which would incorrectly trigger every comparison below.
+	podStatus, exists := r.aeroCluster.Status.Pods[pod.Name]
```
abhishekdwivedi3060: Is this intentional?
For a new pending pod due to invalid resource config - what do we expect when the resources are corrected. If I'm understanding the new flow correctly it'll skip the failed pod rolling restart flow. It'll restarted in the second phase of failed pod

internal/controller/cluster/rack.go
```diff
@@ -675,11 +702,6 @@ func (r *SingleClusterReconciler) scaleUpRack(
 		))
 	}
 
-	if r.isAnyPodInImageFailedState(podList, ignorablePodNames) {
```
abhishekdwivedi3060, line 678: Discuss this:
IsPodUpgraded check removed and this check removal
Scenario - Bad image updated with size increase in the update.

Same with rolling restart + sidecar failure + scale up

internal/controller/cluster/rack.go
```diff
@@ -939,8 +945,14 @@ func (r *SingleClusterReconciler) scaleDownRack(
 		isAnyPodRunningAndReady bool
 	)
 
+	// Clone so we can extend it with server-failed pods from this batch without
+	// mutating the caller's set.
+	ignorablePodNames = ignorablePodNames.Clone()
```
abhishekdwivedi3060: Do we really need this? This list is created new in every reconcile

internal/controller/cluster/rack.go
```diff
@@ -2235,62 +2237,123 @@ func (r *SingleClusterReconciler) handleFailedPodsInRack(
...
+	if asdbv1.GetBool(r.aeroCluster.Spec.IgnoreSidecarFailure) {
+		return common.ReconcileSuccess()
+	}
+
+	sidecarFailedPods := getSidecarFailedPods(podList)
+	if len(sidecarFailedPods) == 0 {
+		return common.ReconcileSuccess()
+	}
+
+	r.Log.Info("Checking sidecar-failed pods for config-change-driven restart",
```
abhishekdwivedi3060: nit:
For consistency, we should use the same wording here like "Reconcile sidecar-failed pods...."

abhishekdwivedi3060 (changes requested):

internal/controller/cluster/rack.go
tanmayja: It will be handled in failed pod handling

internal/controller/cluster/reconciler.go
tanmayja: r.recoverFailedCreate(ctx) never returned non-nil error.
But changed it to return ReconcileError only if there is some err in recoverFailedCreate.

internal/controller/cluster/pod.go
tanmayja: Added a general pod failed check during scale-up so that new nodes does not come up with wrong config/image.

pkg/utils/pod.go
tanmayja: wrote checkFailedWithGrace a common function and 2 wrappers CheckServerFailedWithGrace and CheckPodFailedWithGrace

internal/controller/cluster/rack.go
tanmayja: started returning reconciler error which will make it exponential back off

internal/controller/cluster/rack.go
tanmayja: It was a missed code change from earlier proposed solution. Ideally sidecar failure pods means server is running. so should not be treated as failed pods from server's operations perspective.

internal/controller/cluster/pod.go
tanmayja: yes, now returning pod restart in all cases except noRestart (where there is actually no action needed on that pod)

internal/controller/cluster/pod.go
tanmayja: We fixed this issue during cluster recovery, hence can rollback this change.

test/cluster/cluster_test.go
```diff
@@ -78,6 +78,11 @@ var _ = Describe(
 				UpdateClusterTest(ctx)
 			},
 		)
+		FContext(
```
abhishekdwivedi3060: Remove focus

internal/controller/cluster/rack.go
```diff
@@ -702,6 +726,33 @@ func (r *SingleClusterReconciler) scaleUpRack(
...
+			podState = utils.CheckPodFailedWithGrace(pod, false)
+		}
+
+		if podState.State == utils.PodFailed {
+			return found, common.ReconcileError(fmt.Errorf(
+				"cannot scale up rack %d: pod %s is in failed state: %s",
+				rackState.Rack.ID, utils.GetNamespacedNameString(pod), podState.Reason))
+		}
+	}
+
```
abhishekdwivedi3060: With new changes we are blocking scenarios where the pod is in failed state due to any external reason and user is trying to scale up. Previous flow allowed this behaviour expect for image error (which was genuine). I also understand this check is helping us block bad config change along with scale up. Need to think how we can handle this

internal/controller/cluster/rack.go
```diff
@@ -735,11 +786,22 @@ func (r *SingleClusterReconciler) scaleUpRack(
...
 	}
 
-	// update replicas here to avoid new replicas count comparison while cleaning up dangling pods of rack
-	found.Spec.Replicas = &desiredSize
+	// Scale up the statefulset. Re-fetch the STS inside a RetryOnConflict
+	// loop so that a stale resourceVersion (e.g. from a status update made by
+	// the Kubernetes StatefulSet controller between our last Get and this
+	// Update) does not cause a permanent conflict error. The replica count is
+	// set inside the closure so each retry operates on the latest object.
+	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
abhishekdwivedi3060, line 833: Did we see any issue without RetryOnConflict or it's just to be on safe side?
Also, we might need to add at other places where we are directly updating the STS like scaleDown

internal/controller/cluster/rack.go
```diff
@@ -564,9 +572,23 @@ func (r *SingleClusterReconciler) handleNSOrDeviceRemovalForIgnorablePods(
...
+//     not suppressed for sidecar-failed pods.
+type failedPodsInfo struct {
+	pods           []*corev1.Pod
+	isServerFailed bool
+}
+
 func (r *SingleClusterReconciler) reconcileRack(
 	ctx context.Context, found *appsv1.StatefulSet, rackState *RackState,
-	ignorablePodNames sets.Set[string], failedPods []*corev1.Pod,
+	ignorablePodNames sets.Set[string], failedPods *failedPodsInfo,
```
abhishekdwivedi3060: See if we need to change var name `failedPods`, not that we have change the type

internal/controller/cluster/rack.go
```diff
@@ -418,8 +421,13 @@ func (r *SingleClusterReconciler) upgradeOrRollingRestartRack(
 		}
 
 		if rollingRestartInfo.needRestart {
+			var failedPodsList []*corev1.Pod
+			if failedPods != nil {
+				failedPodsList = failedPods.pods
+			}
+
```
abhishekdwivedi3060: Why are we doing like this?
Why not simply pass the complete *failedPodsInfo and check inside similar to other places

internal/controller/cluster/rack.go
tanmayja: I encountered the issue couple of times. we can add in other places as well.

internal/controller/cluster/reconciler_test.go
```diff
@@ -0,0 +1,196 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:              clusterName + "-0-0",
+				Namespace:         namespace,
+				Labels:            clusterLabels(clusterName),
+				CreationTimestamp: metav1.NewTime(time.Now().Add(-10 * time.Minute)),
+			},
+			Status: corev1.PodStatus{
+				Phase: corev1.PodRunning,
+				Conditions: []corev1.PodCondition{
+					{Type: corev1.PodReady, Status: corev1.ConditionTrue},
```
abhishekdwivedi3060: I think we should set containerStatus here instead of conditions because checkPreviouslyFailedCluster relies on that

internal/controller/cluster/aero_info_calls_test.go
```diff
@@ -0,0 +1,150 @@
...
+	// serverFailedPod has a non-running server container.
+	serverFailedPod := corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "pod-server-fail"},
+		Status: corev1.PodStatus{
+			Phase: corev1.PodRunning,
+			ContainerStatuses: []corev1.ContainerStatus{
+				serverContainer(false),
+			},
+		},
+	}
```
abhishekdwivedi3060, line 95: See if we can leverage/reuse/merge the funcs `crashLoopServerPod` and `sidecarCrashPod` funcs already defined in rack_test.go

internal/controller/cluster/aero_info_calls_test.go
abhishekdwivedi3060, line 1: Can we not move these test-cases to table-driven approach similar to other files?

internal/controller/cluster/rack.go
```diff
@@ -1081,17 +1076,27 @@ func (r *SingleClusterReconciler) scaleDownRack(
...
+		current.Spec.Replicas = &newSize
+		found = current
+
+		return r.Update(ctx, current, common.UpdateOption)
+	}); err != nil {
 		return found, common.ReconcileError(
 			fmt.Errorf(
 				"scale StatefulSet %s to %d replicas: %w",
-				utils.GetNamespacedNameString(found), newSize, err,
+				utils.GetNamespacedNameString(found), *found.Spec.Replicas, err,
```
abhishekdwivedi3060: nit: Cosmetic change
getSTS failure in RetryOnConflict failure won't update the replicas size so this log with print the old size

internal/controller/cluster/rack.go
```diff
@@ -1114,22 +1119,29 @@ func (r *SingleClusterReconciler) scaleDownRack(
...
+				current.Spec.Replicas = &rollbackSize
+				found = current
+
+				return r.Update(ctx, current, common.UpdateOption)
+			}); rollbackErr != nil {
 				return found, common.ReconcileError(
 					fmt.Errorf(
 						"scale StatefulSet %s to %d replicas: %w",
-						utils.GetNamespacedNameString(found), newSize, err,
+						utils.GetNamespacedNameString(found), *found.Spec.Replicas, rollbackErr,
```
abhishekdwivedi3060: same here

internal/controller/cluster/pod_test.go
```diff
@@ -0,0 +1,325 @@
...
+// (order-independent).
+func assertPodNames(t *testing.T, label string, got, want []string) {
+	t.Helper()
+
+	if len(got) != len(want) {
+		t.Errorf("%s: got %v (len %d), want %v (len %d)", label, got, len(got), want, len(want))
+		return
+	}
+
+	wantSet := make(map[string]bool, len(want))
```
abhishekdwivedi3060: nit:
Can use the sets directly instead of map

internal/controller/cluster/pod_test.go
```diff
@@ -0,0 +1,325 @@
...
+			ContainerStatuses: statuses,
+		},
+	}
+}
+
+// TestGetServerFailedAndActivePods_SidecarFailedGoesToActive is the central
+// invariant: a pod with a crashing sidecar but a healthy server must be
+// classified as active, never as failed.  This ensures sidecar-failed pods
+// are not incorrectly skipped during safety checks or batching.
+func TestGetServerFailedAndActivePods_SidecarFailedGoesToActive(t *testing.T) {
```
abhishekdwivedi3060: Isn't this TC already covered in TestGetServerFailedAndActivePods?

internal/controller/cluster/service_test.go
```diff
@@ -0,0 +1,114 @@
...
+// TestCreateOrUpdatePodService verifies invariants of the NodePort service that
+// is created (or updated) per pod:
+//   - PublishNotReadyAddresses is always true so that sidecar-failing pods
+//     (whose Aerospike server is still reachable) continue to receive traffic.
+//   - The call is idempotent — a second invocation must not return an error.
+//   - The behaviour is the same regardless of the IgnoreSidecarFailure flag.
+func TestCreateOrUpdatePodService(t *testing.T) {
+	const (
+		namespace   = "test-ns"
+		clusterName = "test-cluster"
```
abhishekdwivedi3060: We are defining same constants at multiple places

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+					"expected Aerospike server to be running on pod %s", pod.Name)
+				Expect(podHasCrashingSidecar(pod)).To(BeTrue(),
+					"expected crashing sidecar on pod %s (all pods carry the new spec)", pod.Name)
+			}
+		})
+	})
+})
+
+// intOrStr is a convenience wrapper so test code can build an IntOrString inline.
+func intOrStr(v int) intstr.IntOrString {
```
abhishekdwivedi3060: Seems redundant, we can simply use intstr.FromInt(v)

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			// so the update reliably reaches Completed.
+			By("Deploying a healthy 2-node cluster")
+
+			aeroCluster := createDummyAerospikeCluster(clusterNamespacedName, 2)
+			Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+			By("Adding a crashing sidecar with IgnoreSidecarFailure=true via a rolling restart")
+
+			aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
```
abhishekdwivedi3060: We can avoid these API calls before update everywhere because updateCluster already fetches the latest object before update

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+
+			By("Verifying Aerospike server is running and reachable in every pod (nodeIDs populated)")
+
+			cluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
+
+			for podName, podStatus := range cluster.Status.Pods {
+				Expect(podStatus.Aerospike.NodeID).ToNot(BeEmpty(),
+					"expected nodeID to be set for pod %s", podName)
+			}
```
abhishekdwivedi3060: These would be already populated after first deployment so I'm not sure if check is useful here.
Should we try asinfo calls to check server is reachable?

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+				podList, podListErr := getClusterPodList(k8sClient, ctx, cluster)
+				g.Expect(podListErr).ToNot(HaveOccurred())
+
+				crashingCount := 0
+				readyCount := 0
+
+				for idx := range podList.Items {
+					pod := &podList.Items[idx]
+					if podHasCrashingSidecar(pod) {
+						crashingCount++
```
abhishekdwivedi3060: We can continue from here

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			Expect(finalPodList.Items).To(HaveLen(int(updatedCluster.Spec.Size)))
+
+			for idx := range finalPodList.Items {
+				pod := &finalPodList.Items[idx]
+				Expect(podHasCrashingSidecar(pod)).To(BeTrue(),
+					"expected crashing sidecar on pod %s after full rolling restart", pod.Name)
+				Expect(utils.IsAerospikeServerReady(pod)).To(BeTrue(),
+					"expected Aerospike server to be running on pod %s", pod.Name)
+			}
+		})
```
abhishekdwivedi3060: We are testing this multiple times, lets move it to a common validation func

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			Consistently(checkBlockedState, 30*time.Second, 5*time.Second).Should(Succeed())
+
+			By("Enabling IgnoreSidecarFailure=true and verifying cluster reaches Completed")
+
+			cluster.Spec.IgnoreSidecarFailure = ptr.To(true)
+			Expect(updateCluster(k8sClient, ctx, cluster)).ToNot(HaveOccurred())
+
+			updatedCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(updatedCluster.Status.Phase).To(Equal(asdbv1.AerospikeClusterCompleted))
```
abhishekdwivedi3060: Not required. This is already covered in updateCluster above

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+				}
+
+				g.Expect(crashingCount).To(Equal(1),
+					"expected exactly 1 pod with crashing sidecar; rolling restart blocked after the first pod")
+				g.Expect(readyCount).To(Equal(int(cluster.Spec.Size)-1),
+					"expected the remaining pods to still be fully ready (not yet restarted)")
+			}
+
+			Eventually(checkBlockedState, 5*time.Minute, 5*time.Second).Should(Succeed())
+			Consistently(checkBlockedState, 30*time.Second, 5*time.Second).Should(Succeed())
```
abhishekdwivedi3060: This also is duplicate, can be moved to a func

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      clusterName,
+					Namespace: namespace,
+				},
+			}
+			Expect(DeleteCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+			Expect(CleanupPVC(k8sClient, aeroCluster.Namespace, aeroCluster.Name)).ToNot(HaveOccurred())
+		})
+
+		It("Should complete rolling restart when IgnoreSidecarFailure is true and sidecar is crashing", func() {
```
abhishekdwivedi3060: This TC seems redundant.
Sidecar addition + config change would result in PodRestart only which is already covered in above context 
```
It("Should complete reconciliation when IgnoreSidecarFailure is true and sidecar is crashing"
```

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			// again (sidecar crash event) so we do not re-check it.
+			cluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
+
+			for podName, podStatus := range cluster.Status.Pods {
+				Expect(utils.IsImageEqual(podStatus.Image, nextImage)).To(BeTrue(),
+					"expected pod %s to be on image %s after upgrade, got %s",
+					podName, nextImage, podStatus.Image)
+			}
+		})
```
abhishekdwivedi3060: This image check is already covered in updateCluster func internally

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			// restarts within the pod so the fail signal survives a restart.
+			signalVolume := asdbv1.VolumeSpec{
+				Name: "sidecar-signal",
+				Source: asdbv1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
+				Sidecars: []asdbv1.VolumeAttachment{
+					{ContainerName: "signal-sidecar", Path: "/signal"},
+				},
+			}
```
abhishekdwivedi3060: Use signalControlledSidecar and signalVolumeForSidecar instead

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			By("Verifying all pods are fully ready — sidecar is healthy before the fail signal is sent")
+
+			podList, err := getClusterPodList(k8sClient, ctx, aeroCluster)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(podList.Items).To(HaveLen(2))
+
+			for idx := range podList.Items {
+				Expect(utils.IsPodRunningAndReady(&podList.Items[idx])).To(BeTrue(),
+					"expected pod %s to be fully ready before fail signal", podList.Items[idx].Name)
+			}
```
abhishekdwivedi3060: I'm not sure if these checks are required for fresh deployed clusters because Completed phase of the CR denotes that

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			By("Verifying rolling restart is stuck in InProgress — sidecar failure blocks readiness wait")
+
+			Expect(waitForClusterPhase(k8sClient, ctx, clusterNamespacedName,
+				2*time.Minute, asdbv1.AerospikeClusterInProgress),
+			).ToNot(HaveOccurred())
+
+			stuckCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(stuckCluster.Status.Phase).ToNot(Equal(asdbv1.AerospikeClusterCompleted),
+				"cluster should not have completed: sidecar failure must block the rolling restart")
```
abhishekdwivedi3060: Isn't this already covered in waitForClusterPhase above?

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+
+			By("Setting IgnoreSidecarFailure=true and verifying rolling restart completes")
+
+			stuckCluster.Spec.IgnoreSidecarFailure = ptr.To(true)
+			Expect(updateCluster(k8sClient, ctx, stuckCluster)).ToNot(HaveOccurred())
+
+			completedCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(completedCluster.Status.Phase).To(Equal(asdbv1.AerospikeClusterCompleted))
```
abhishekdwivedi3060: redundant

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+			By("Verifying the SC roster includes the sidecar-failed pod — its Aerospike server was never affected")
+
+			// The sidecar-failed pod's Aerospike server ran throughout; the pod
+			// must remain in the roster. If the sidecar failure were incorrectly
+			// counted against MaxIgnorablePods the pod would be skipped and
+			// could drop out of the roster.
+			validateRoster(k8sClient, ctx, clusterNamespacedName, scNamespace)
+		})
+
+		It("Should not block rolling restart when multiple sidecar-failed pods exceed the MaxIgnorablePods budget", func() {
```
abhishekdwivedi3060: This seems redundant. What extra are we testing here?

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+
+			By("Failing the sidecar on pod-0 AND pod-1 — 2 pods sidecar-failed, exceeding budget of 1")
+
+			for _, suffix := range []string{"-0-0", "-0-1"} {
+				Expect(failPodSidecar(namespace, clusterName+suffix)).ToNot(HaveOccurred())
+			}
+
+			Eventually(func(g Gomega) {
+				cluster, clusterErr := getCluster(k8sClient, ctx, clusterNamespacedName)
+				g.Expect(clusterErr).ToNot(HaveOccurred())
```
abhishekdwivedi3060: redundant. Check at all places

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+					g.Expect(utils.IsImageEqual(podStatus.Image, nextImage)).To(BeTrue(),
+						"expected pod %s to be on image %s after upgrade, got %s",
+						pod.Name, nextImage, podStatus.Image)
+					g.Expect(utils.IsAerospikeServerReady(pod)).To(BeTrue(),
+						"expected Aerospike server to be running on pod %s after upgrade", pod.Name)
+					g.Expect(podHasCrashingSidecar(pod)).To(BeTrue(),
+						"expected sidecar to still be crashing on pod %s after upgrade", pod.Name)
+				}
+			}, 3*time.Minute, 10*time.Second).Should(Succeed())
+		})
```
abhishekdwivedi3060: We can use the verifyClusterState func here and wherever required

test/cluster/sidecar_failure_test.go
```diff
@@ -0,0 +1,1124 @@
...
+	})
+
+	Context("Multi-rack cluster with crashing sidecar", func() {
```
abhishekdwivedi3060: Again this can be clubbed with the existing test-cases, seems redundant

internal/controller/cluster/rack.go
```diff
@@ -476,9 +485,62 @@ func (r *SingleClusterReconciler) upgradeOrRollingRestartRack(
...
+
+		var podState utils.PodState
+		if ignoreSidecar {
+			podState = utils.CheckServerFailedWithGrace(pod, false)
+		} else {
+			podState = utils.CheckPodFailedWithGrace(pod, false)
+		}
+
+		if podState.State == utils.PodFailed {
+			return common.ReconcileResult{IsSuccess: false}
```
abhishekdwivedi3060, line 546: Lets add a comment here

internal/controller/cluster/rack.go
```diff
@@ -2326,7 +2361,7 @@ func (r *SingleClusterReconciler) handleFailedPodsInRack(
 		if res := r.reconcileRack(
 			ctx, found, rackState, ignorablePodNames,
 			&failedPodsInfo{pods: serverFailedPods, isServerFailed: true},
-		); !res.IsSuccess {
+		); res.Err != nil {
```
abhishekdwivedi3060, line 2408: Lets add comment here and below for future ref why we checking error instead of Success

internal/controller/cluster/rack_test.go
```diff
@@ -86,6 +89,191 @@ func newTestReconciler(
...
+			Build()
+
+		return &SingleClusterReconciler{
+			Client:      fakeClient,
+			Log:         logr.Discard(),
+			Scheme:      scheme,
+			aeroCluster: aeroCluster,
+			Recorder:    record.NewFakeRecorder(10),
+		}
+	}
```
abhishekdwivedi3060: Use newTestReconciler instead

test/cluster/cluster_test.go
```diff
@@ -2283,3 +2288,154 @@ func negativeUpdateClusterValidationTest(
...
+
+				for idx := range podList.Items {
+					g.Expect(podList.Items[idx].Name).NotTo(Equal(scaleUpPodName),
+						"scale-up pod %s must not be created while existing pods are failing with bad image",
+						scaleUpPodName)
+				}
+			}
+
+			By("Waiting for the cluster to enter Error state (bad-image rolling restart blocks scale-up)")
+			Eventually(checkScaleUpBlocked, 5*time.Minute, 10*time.Second).Should(Succeed())
```
abhishekdwivedi3060: This check doesn't ensure that cluster has entered error state which the By() says

test/cluster/cluster_test.go
```diff
@@ -2283,3 +2288,154 @@ func negativeUpdateClusterValidationTest(
...
+			Eventually(func(g Gomega) {
+				cluster, clusterErr := getCluster(k8sClient, ctx, clusterNamespacedName)
+				g.Expect(clusterErr).ToNot(HaveOccurred())
+				g.Expect(cluster.Status.Phase).To(Equal(asdbv1.AerospikeClusterCompleted))
+
+				podList, podListErr := getClusterPodList(k8sClient, ctx, cluster)
+				g.Expect(podListErr).ToNot(HaveOccurred())
+				g.Expect(podList.Items).To(HaveLen(3),
+					"expected 3 pods after cluster recovery and scale-up")
+			}, getTimeout(3), 15*time.Second).Should(Succeed())
```
abhishekdwivedi3060: updateCluster func already covers similar checks internally

abhishekdwivedi3060 (changes requested):
1. There are multiple duplicate funcs for different pod types, container statuses in different files. Try to define generic func and reuse them
2. Remove redundant API calls to reduce the CI run time
3. Remove/merge redundant test-case

internal/controller/cluster/rack.go
tanmayja: Moved these checks to upgradeOrRollingRestart function so that we can skip these validation in pure scale-up.

test/cluster/cluster_test.go
```diff
@@ -2357,10 +2357,7 @@ func ScaleUpPreflightChecksTest(ctx goctx.Context) {
 				}
 			}
 
-			By("Waiting for the cluster to enter Error state (bad-image rolling restart blocks scale-up)")
-			Eventually(checkScaleUpBlocked, 5*time.Minute, 10*time.Second).Should(Succeed())
-
-			By("Confirming the scale-up pod stays absent for a sustained period")
+			By("Confirming scale-up pod stays absent while the bad-image rolling restart is active")
 			Consistently(checkScaleUpBlocked, 30*time.Second, 5*time.Second).Should(Succeed())
```
abhishekdwivedi3060: Might need to increase the timeout for this check. 30s is not sufficient for this check

internal/controller/cluster/rack.go
abhishekdwivedi3060: nit: 
We can use the scaleDownTargetSize var instead

abhishekdwivedi3060: same here

internal/controller/cluster/rack_test.go
abhishekdwivedi3060: serverCrashLoopContainer

test/cluster/sidecar_failure_test.go
```diff
@@ -1013,40 +678,18 @@ done`,
...
-			// running, all sidecars are crashing, and the pod count matches
-			// expectedSize. Phase is not checked here because the perpetual
-			// CrashLoopBackOff events can briefly push the cluster back to
-			// InProgress between reconcile cycles; the important invariant is
-			// that every Aerospike server is up and every sidecar is crashing.
+			// verifyClusterState asserts all servers are up, all sidecars are
+			// crashing, and the pod count matches expectedSize. Phase is not
+			// checked because perpetual CrashLoopBackOff events can briefly push
+			// the cluster back to InProgress between reconcile cycles.
 			verifyClusterState := func(expectedSize int) {
```
abhishekdwivedi3060: Unnecessary wrapper

test/cluster/sidecar_failure_test.go
```diff
@@ -569,58 +357,54 @@ var _ = Describe("SidecarFailure", func() {
 
 			validateRoster(k8sClient, ctx, clusterNamespacedName, scNamespace)
 		})
-	})
 
-	Context("Image upgrade with crashing sidecar", func() {
-		clusterName := fmt.Sprintf("sidecar-upgrade-%d", GinkgoParallelProcess())
-		clusterNamespacedName := test.GetNamespacedName(clusterName, namespace)
+		It("Should block rolling restart across all racks when IgnoreSidecarFailure is false, then"+
```
abhishekdwivedi3060, line 373: Can we not combine this TC and "Should complete reconciliation when IgnoreSidecarFailure is true and sidecar is crashing" together with a multi-rack and rolling restart.
Also, lets have a TC for upgrade flow with sidecar failure

abhishekdwivedi3060 (dismissed):

test/cluster/sidecar_failure_test.go
tanmayja: did not merge the tests but added upgrade flow in existing "Should complete reconciliation when IgnoreSidecarFailure is true"

internal/controller/cluster/rack.go
```diff
@@ -2473,23 +2498,39 @@ func (r *SingleClusterReconciler) handleFailedPodsInRack(
...
+		if utils.CheckPodFailed(pod) != nil {
+			crashedPods = append(crashedPods, pod)
+		} else {
+			stillNotReadyPods = append(stillNotReadyPods, pod)
 		}
 	}
 
-	notReadyPods = getNonIgnorablePods(notReadyPods, ignorablePodNames)
+	crashedPods = getNonIgnorablePods(crashedPods, ignorablePodNames)
+	stillNotReadyPods = getNonIgnorablePods(stillNotReadyPods, ignorablePodNames)
```
abhishekdwivedi3060, line 2517: Why not do it once for the podList and then start the loop?

abhishekdwivedi3060 (dismissed):

abhishekdwivedi3060 (approved):

PR #560, opened 2026-07-17
chore: add operator helm chart schema [KO-587]

sidntrivedi:
Addresses https://aerospike.atlassian.net/browse/KO-587.

## Summary
- Adds helm values schema for the aerospike-kubernetes-operator helm chart.
- Adds test for the chart values according to the schema using [helm-unittest](https://github.com/helm-unittest/helm-unittest)
- Modifies the helm lint github action to run the added tests as well

PR to strip `@schema` and `# --` from `values.yaml` while releasing: https://github.com/aerospike/aerospike-kubernetes-operator/pull/574.

### Automated test suite result

```
❯ make helm-test
==> helm unittest helm-charts/aerospike-kubernetes-operator

### Chart [ aerospike-kubernetes-operator ] helm-charts/aerospike-kubernetes-operator

 PASS  values.schema.json edge cases — certs	helm-charts/aerospike-kubernetes-operator/tests/certs_test.yaml
 PASS  values.schema.json edge cases — image and registry	helm-charts/aerospike-kubernetes-operator/tests/image_test.yaml
2026/08/05 11:29:07 warning: cannot overwrite table with non table for aerospike-kubernetes-operator.metricsService.labels (map[])
2026/08/05 11:29:07 warning: cannot overwrite table with non table for aerospike-kubernetes-operator.metricsService.annotations (map[])
2026/08/05 11:29:07 warning: cannot overwrite table with non table for aerospike-kubernetes-operator.webhookService.labels (map[])
2026/08/05 11:29:07 warning: cannot overwrite table with non table for aerospike-kubernetes-operator.webhookService.annotations (map[])
 PASS  values.schema.json edge cases — networking (ports and services)	helm-charts/aerospike-kubernetes-operator/tests/networking_test.yaml
 PASS  values.schema.json edge cases — operator runtime behavior	helm-charts/aerospike-kubernetes-operator/tests/operator_test.yaml
 PASS  values.schema.json edge cases — resources	helm-charts/aerospike-kubernetes-operator/tests/resources_test.yaml
2026/08/05 11:29:07 warning: skipped value for aerospike-kubernetes-operator.imagePullSecrets: Not a table.
 PASS  values.schema.json edge cases — scheduling and k8s-native passthrough fields	helm-charts/aerospike-kubernetes-operator/tests/scheduling_and_passthrough_test.yaml

Charts:      1 passed, 1 total
Test Suites: 6 passed, 6 total
Tests:       128 passed, 128 total
Snapshot:    0 passed, 0 total
Time:        163.282792ms
```


helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                },
+                "required": ["serviceAccountName"]
+            },
+            "description": "RBAC resource creation for the operator"
+        },
+        "readinessProbe": {
+            "type": "object",
+            "description": "Readiness probe configuration for the operator container, passed through as-is (validated by the Kubernetes API server)"
+        },
+        "replicas": {
```
shubhamyelikar99, line 395: Value-name typos are silently ignored.

JSON Schema permits unlisted properties unless told otherwise, so a misspelled key is accepted and its intended target keeps its default:
~~~
replica: 5     # meant "replicas"
~~~
--set replica=5 installs fine and deploys 2 replicas, silently.

install succeeds and the Deployment gets 2 replicas. The user believes they configured 5

Adding "additionalProperties": false turns it into:
~~~
at '': additional properties 'replica' not allowed
~~~
Four keys must be declared first, or this breaks working setups:

- nameOverride and fullnameOverride — read by _helpers.tpl (L6, L15) but present in neither values.yaml nor this schema. Worth declaring regardless, since they're currently supported-but-undocumented.
- webhookServicePort — the deprecated field trapped by validateDeprecateFields. Without declaring it, strict mode downgrades that helpful error:
~~~
# today
Error: .Values.webhookServicePort field is deprecated, use .Values.webhookService.targetPort instead
# with naive additionalProperties: false
Error: additional properties 'webhookServicePort' not allowed
~~~
- global — injected by Helm whenever this chart is used as a subchart. 

Please check if there are any other such fields.

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+        },
+        "labels": {
+            "type": "object",
+            "description": "Deployment labels, inherited by pods automatically, passed through as-is (validated by the Kubernetes API server)"
+        },
+        "livenessProbe": {
+            "type": "object",
+            "description": "Liveness probe configuration for the operator container, passed through as-is (validated by the Kubernetes API server)"
+        },
+        "logging": {
```
shubhamyelikar99, line 201: No root-level required, so a blank line in a values file causes a Go panic.

In YAML a key with nothing after the colon is null, and Helm's coalescing deletes null keys rather than merging them. Both of these produce logging: null, so logging disappears from the merged values entirely:
~~~
logging:                    # intending to fill in later
logging:
  # development: true       # children commented out
~~~
The schema only validates keys that are present, so validation passes and rendering then dereferences a nil map:
~~~
Error: ...deployment.yaml:56:34 executing "...deployment.yaml" at <.Values.logging.development>:
  nil pointer evaluating interface {}.development
~~~
That's a Go internals error citing a template line the user never wrote.

Similar for other fields operatorImage:, and livenessProbe: — every object the templates dereference without a with/if guard.

Declaring these as required at the root turns it into at '': missing property 'logging'. Suggested set, based on which objects the templates dereference unguarded:
~~~
"required": [
    "replicas", "operatorImage", "rbac", "healthPort", "metricsPort",
    "certs", "logging", "metrics", "watchNamespaces", "metricsService",
    "webhookService", "livenessProbe", "readinessProbe", "safePodEviction",
    "aerospikeKubernetesInitRegistry", "aerospikeKubernetesInitRegistryNamespace",
    "aerospikeKubernetesInitNameTag"
]
~~~
Check if there are any other such fields

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                            "description": "Create a self-signed certificate for the webhook endpoint"
+                        },
+                        "webhookServerCertSecretName": {
+                            "type": "string",
+                            "minLength": 1,
+                            "default": "webhook-server-cert",
+                            "description": "Name of the secret holding the webhook server certificate, always mounted regardless of webhook.create; webhook certs are mandatory for AKO, so if webhook.create is false this secret must already exist and be provided by the user"
+                        }
+                    },
+                    "required": ["webhookServerCertSecretName"]
```
shubhamyelikar99: This required can never fire in the case that needs it.

certs.webhook requires webhookServerCertSecretName, but nothing requires certs.webhook to exist, and nothing requires certs to exist. So a user who removes the parent bypasses the rule entirely:
~~~
certs:      # -> null -> key deleted -> nothing left to validate
Error: ...webhook-serving-cert.yaml:1:14 at <.Values.certs.webhook.create>:
  nil pointer evaluating interface {}.webhook
~~~
The requirement only holds when it wasn't needed. Adding "required": ["webhook", "metrics"] to certs closes the gap.
Check for similar pattern in other fields

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                    "type": "string"
+                },
+                {
+                    "type": "integer",
+                    "minimum": 1
+                }
+            ],
+            "default": "60",
+            "description": "Grace period in seconds to delete/recover failed Aerospike pods; must be a positive integer (silently falls back to 60 if invalid or unset)"
+        },
```
shubhamyelikar99, line 150: This anyOf accepts any string at all, so the field is effectively unvalidated.

anyOf means "satisfy at least one branch," and branch 1 is {"type": "string"} with no constraints. Any string satisfies it and branch 2 never gets to object:
~~~
failedPodGracePeriodSeconds: "abc"    # ACCEPTED
failedPodGracePeriodSeconds: "-5"     # ACCEPTED
~~~
Why the test suite doesn't catch this. The test uses set: failedPodGracePeriodSeconds: -5, and set:/--set parse -5 as a number — so branch 1 fails (not a string) and branch 2 fails (< 1), the value is rejected, and the test goes green. But the same value quoted in a values file takes branch 1 and passes.

And the quoted form is the one users will write, because that's what the chart ships — values.yaml L60 is failedPodGracePeriodSeconds: "60". So the real-world path is the unvalidated one while the tests exercise the validated one.

There's a small irony in the description on L106: "must be a positive integer (silently falls back to 60 if invalid or unset)" — it documents precisely the failure it then allows. Constraining the string branch keeps "60" working while rejecting the rest:
~~~
"anyOf": [
    {
        "type": "string",
        "pattern": "^[1-9][0-9]*$"
    },
    {
        "type": "integer",
        "minimum": 1
    }
],
~~~
Check for similar patterns

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                "timeoutSeconds": {
+                    "anyOf": [
+                        {
+                            "type": "string"
+                        },
+                        {
+                            "type": "integer",
+                            "minimum": 1,
+                            "maximum": 30
+                        }
```
shubhamyelikar99, line 467: Same open-string-branch issue, and here it defeats a hard Kubernetes limit.

The integer branch correctly caps at 30, matching the ValidatingWebhookConfiguration API. The string branch ignores it:
~~~
safePodEviction:
  timeoutSeconds: "999"    # ACCEPTED
~~~
Verified. values.yaml L57 ships timeoutSeconds: "20" — quoted — so users copying the shipped style land on the unvalidated branch, and the API server rejects the webhook config later.
~~~
"anyOf": [
    {
        "type": "string",
        "pattern": "^([1-9]|[12][0-9]|30)$"
    },
    {
        "type": "integer",
        "minimum": 1,
        "maximum": 30
    }
],
~~~
Check if this works

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                                }
+                            ],
+                            "default": "64Mi",
+                            "description": "Memory request as a Kubernetes quantity, must be greater than 0 (e.g. 64Mi, 1Gi)"
+                        }
+                    }
+                }
+            },
+            "description": "Resource requests and limits for the operator container"
+        },
```
shubhamyelikar99, line 427: All four resource fields accept arbitrary strings, and bare numbers are a foot-gun.

Same open string branch as the two comments above.:
~~~
resources:
  limits:
    cpu: "abc"                # ACCEPTED -> API server rejects the Deployment later
    memory: "notaquantity"    # ACCEPTED -> same
~~~
Separately, the number branch means bare values are legal, and in Kubernetes a bare number for memory is bytes — so memory: 512 requests 512 bytes and the operator is OOM-killed immediately. Hard to construct a case where anyone means that.

Since the same anyOf is repeated four times here, a $defs entry keeps it in one place:
~~~
"$defs": {
    "quantity": {
        "anyOf": [
            {
                "type": "string",
                "pattern": "^([0-9]+(\\.[0-9]*)?|\\.[0-9]+)(([KMGTPE]i)|[numkMGTPE]|[eE](\\+|-)?[0-9]+)?$"
            },
            { "type": "number", "exclusiveMinimum": 0 }
        ]
    }
}
~~~
Each field then becomes:

~~~
"cpu": {
    "$ref": "#/$defs/quantity",
    "default": "400m",
    "description": "CPU limit as a Kubernetes quantity, must be greater than 0 (e.g. 400m, 1)"
}
~~~
Check if this works

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+            "description": "Node labels for pod assignment, passed through as-is (validated by the Kubernetes API server)"
+        },
+        "operatorImage": {
+            "type": "object",
+            "properties": {
+                "pullPolicy": {
+                    "type": "string",
+                    "default": "IfNotPresent",
+                    "description": "Operator image pull policy"
+                },
```
shubhamyelikar99, line 306: pullPolicy has exactly three legal values but accepts anything.
~~~
operatorImage:
  pullPolicy: Sometimes     # ACCEPTED
~~~

Dropping validation for open-ended structures — label shapes, toleration structure, probe config — is right, since duplicating the API server's rules would go stale. 
But Always | IfNotPresent | Never is a closed set that hasn't changed in a long time.

~~~
"pullPolicy": {
    "type": "string",
    "enum": ["Always", "IfNotPresent", "Never"],
    "default": "IfNotPresent",
    "description": "Operator image pull policy"
},
~~~
Gives: at '/operatorImage/pullPolicy': value must be one of 'Always', 'IfNotPresent', 'Never'

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                    "type": "string",
+                    "default": "aerospike/aerospike-kubernetes-operator",
+                    "description": "Operator image repository"
+                },
+                "tag": {
+                    "type": "string",
+                    "default": "4.5.0",
+                    "description": "Operator image tag"
+                }
+            },
```
shubhamyelikar99, line 315: An empty repository or tag produces a YAML parse error instead of a schema error.

Neither field has minLength, so "" passes validation. The template then renders image: {{ .repository }}:{{ .tag }} as image: aerospike/aerospike-kubernetes-operator:, which is malformed YAML:

~~~
Error: YAML parse error on .../deployment.yaml: error converting YAML to JSON:
  yaml: line 26: mapping values are not allowed in this context
~~~
Check if this works
~~~
"repository": {
    "type": "string",
    "minLength": 1,
    "default": "aerospike/aerospike-kubernetes-operator",
    "description": "Operator image repository"
},
"tag": {
    "type": "string",
    "minLength": 1,
    "default": "4.5.0",
    "description": "Operator image tag"
}
~~~
Gives: at '/operatorImage/tag': minLength: got 0, want 1

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,441 @@
...
+                    "minimum": 1,
+                    "maximum": 65535,
+                    "default": 8443,
+                    "description": "Service port"
+                },
+                "type": {
+                    "type": "string",
+                    "default": "ClusterIP",
+                    "description": "Kubernetes Service type"
+                },
```
shubhamyelikar99: Service type is a closed Kubernetes enum but accepts anything.
~~~
metricsService:
  type: Anything     # ACCEPTED
~~~
Same reasoning as the pullPolicy comment — four stable values, worth constraining.
~~~
"type": {
    "type": "string",
    "enum": ["ClusterIP", "NodePort", "LoadBalancer", "ExternalName"],
    "default": "ClusterIP",
    "description": "Kubernetes Service type"
},
~~~
Same change applies to webhookService.type at L417–421.

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -11,7 +11,7 @@ operatorImage:
 ## authentication, a secret containing credentials can be added
 ## imagePullSecrets:
 ##   - secret_with_credentials_to_custom_registry
-imagePullSecrets: {}
+imagePullSecrets: []
```
shubhamyelikar99: imagePullSecrets: {} → [] (values.yaml L14) is a correct fix, but it could be a breaking change for existing users. Please verify it once.

The old default was a map while the template and the documented example both treat it as a list — the mismatch was invisible because with skips empty maps and empty lists alike. The new type: array forced the fix, which is the schema doing its job.

But anyone carrying imagePullSecrets: {} in their own values file (likely, since that's what the old values.yaml shipped) now fails validation on upgrade:
~~~
at '/imagePullSecrets': got object, want array
~~~
Their rendered manifest is identical either way, so it's pure upgrade friction.

shubhamyelikar99 (changes requested):

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Post discussion with @abhishekdwivedi3060 , we had concluded not to validate kubernetes specific fields/features.

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Post discussion with @abhishekdwivedi3060 , we had concluded not to validate kubernetes specific fields/features. Here the resource values we depend on user to provide the right ones that comply with kubernetes standard.

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Post discussion, we concluded not to validate kubernetes specific fields/features.

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.schema.json
```diff
@@ -0,0 +1,551 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "required": [
+        "replicas",
+        "operatorImage",
+        "rbac",
+        "healthPort",
+        "healthPortTest",
+        "metricsPort",
```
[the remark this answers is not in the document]
sidntrivedi: Ack

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Have made this flexible now. Both object and array types are now supported.

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.schema.json
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
+operatorImage: # @schema required; additionalProperties: false
+  repository: aerospike/aerospike-kubernetes-operator # @schema required; minLength: 1
+  tag: 4.5.0 # @schema required; minLength: 1
+  pullPolicy: IfNotPresent # @schema required
+
+# -- In case the above image is pulled from a registry that requires
+# authentication, a secret containing credentials can be added
+# imagePullSecrets:
+#   - secret_with_credentials_to_custom_registry
+imagePullSecrets: {} # @schema anyOf: [{type: array, items: {type: string}}, {type: object}]
```
tanmayja: i think we have decided to change it to [] as default value, but allow both types as anyof

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
-   failureThreshold: 3
-readinessProbe:
-   initialDelaySeconds: 5
-   periodSeconds: 10
-   timeoutSeconds: 1
-   successThreshold: 1
-   failureThreshold: 3
\ No newline at end of file
+# -- Liveness probe configuration
+livenessProbe: # @schema required; additionalProperties: false
```
tanmayja: with additionalProperties: false, are we going to block other fields which user might add like `terminationGracePeriodSeconds`?

sidntrivedi: We are directly referencing all the fields of `livenessProbe` and `readinessProbe` in the template:

```
          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
          successThreshold: {{ .Values.livenessProbe.successThreshold }}
          failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
```

https://github.com/aerospike/aerospike-kubernetes-operator/blob/7d97c07cb7cbec0b860e682da271967be4147dba/helm-charts/aerospike-kubernetes-operator/templates/aerospike-operator-controller-manager-deployment.yaml#L108-L125

If the user needs to add `terminationGracePeriodSeconds`, I don't think its currently supported. We might need to change the template to allow users to add any field that is supported in the probes.

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: If I recall it was decided to currently allow both object and array and in major release it was planned to switch the type to [].

@abhishekdwivedi3060 can you please confirm.

abhishekdwivedi3060: It was decide to change to [] along with support for both in the schema. Will drop anyof in major version if required.

sidntrivedi: Made the change in [bc98364](https://github.com/aerospike/aerospike-kubernetes-operator/pull/560/commits/bc9836495f3bb46319ba6131682ba60aaaa335bc).

Makefile
```diff
@@ -197,6 +197,43 @@ env-test-backup-service: fmt vet setup-envtest ## Run backup service envtests.
...
 env-test-restore: fmt vet setup-envtest ## Run restore envtests.
 	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)"; cd $(shell pwd)/test/envtests/restore && mkdir -p ../../test-results && go run github.com/onsi/ginkgo/v2/ginkgo -r --focus "$(FOCUS)" -coverprofile envcover.out -timeout=1h0m0s --junit-report=../../test-results/junit-envtests-restore.xml -- . ${ARGS}
+
+##@ Helm
+
+.PHONY: helm-unittest-plugin
+helm-unittest-plugin: ## Install the pinned helm-unittest plugin locally, replacing any other version.
+	@helm plugin list 2>/dev/null | awk '$$1 == "unittest" { print $$2 }' | grep -qx '$(HELM_UNITTEST_VERSION)' || { \
+		helm plugin uninstall unittest >/dev/null 2>&1 || true; \
+		helm plugin install https://github.com/helm-unittest/helm-unittest --verify=false --version v$(HELM_UNITTEST_VERSION); \
```
abhishekdwivedi3060: This `--verify` flag is not supported by older versions of helm like v3.17.0. Make this compatible for v3+ helm verions

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
-  # encoder: console
-
-## Operator manager /metrics (--metrics-secure)
-## Always passed in the Deployment args for clarity; defaults to TLS on (`true`).
-metrics:
-  secure: true
-
-##  Operator configurations
+# -- Operator manager logging (controller-runtime / Zap)
+# development must stay set; level/encoder are optional (uncomment or use --set).
```
abhishekdwivedi3060:
```suggestion
# development must stay set; level/encoder are optional (update value or use --set).
```

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
+# @schema additionalProperties: false
+# -- Enable the eviction webhook to safely block Aerospike pod evictions during node maintenance
 # Also enables Prometheus metrics: aerospike_ako_eviction_webhook_requests_total (labels: eviction_namespace, decision)
-safePodEviction:
-  enable: false
-  # Eviction webhook timeout in seconds
-  timeoutSeconds: "20"
+safePodEviction: # @schema required
+  enable: false # @schema required
+  timeoutSeconds: "20" # @schema anyOf: [{type: string, "pattern": "^([1-9]|[12][0-9]|30)$"}, {type: integer, minimum: 1, maximum: 30}]
```
abhishekdwivedi3060: Why are we defining the maximum limit as 30 here?

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
   limits:
-    cpu: 400m
-    memory: 512Mi
+    cpu: 400m # @schema anyOf: [{type: string}, {type: number, exclusiveMinimum: 0}]
+    memory: 512Mi # @schema anyOf: [{type: string}, {type: number, exclusiveMinimum: 0}]
   requests:
-    cpu: 10m
-    memory: 64Mi
+    cpu: 10m # @schema anyOf: [{type: string}, {type: number, exclusiveMinimum: 0}]
+    memory: 64Mi # @schema anyOf: [{type: string}, {type: number, exclusiveMinimum: 0}]
```
abhishekdwivedi3060: Lets remove these k8s specific validations

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
 affinity: {}
-  # nodeAffinity:
-  #   requiredDuringSchedulingIgnoredDuringExecution:
-  #     nodeSelectorTerms:
-  #     - matchExpressions:
-  #       - key: kubernetes.io/hostname
-  #         operator: In
-  #         values:
-  #         - <Kubernetes Host>
-
```
abhishekdwivedi3060, line 90: Why did we remove the eg?

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
 extraEnv: {}
 
-## Node selector
+# -- Node selector
 nodeSelector: {}
 
-## Tolerations
-tolerations: []
+# -- Tolerations
+tolerations: [] # @schema item: object
```
abhishekdwivedi3060: remove schema marker from here

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
+  successThreshold: 1 # @schema required
+  failureThreshold: 3 # @schema required
+
+# -- Readiness probe configuration
+readinessProbe: # @schema required; additionalProperties: false
+  initialDelaySeconds: 5 # @schema required
+  periodSeconds: 10 # @schema required
+  timeoutSeconds: 1 # @schema required
+  successThreshold: 1 # @schema required
+  failureThreshold: 3 # @schema required
```
abhishekdwivedi3060: 1. All these nested fields are optional at the k8s level
2. We should skip all validation related to k8s fields from everywhere

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
+
+# -- (Optional) Override the chart name used in generated resource names
+nameOverride: ""
+
+# -- (Optional) Override the fully qualified name used for generated resource names
+fullnameOverride: ""
+
+# -- Deprecated, use webhookService.targetPort instead. Setting this fails chart rendering with a message pointing to the replacement
+webhookServicePort: # @schema type: [integer, null]; deprecated; minimum: 1; maximum: 65535
```
abhishekdwivedi3060: Remove this. We've a validation around this to not allow user to set this

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -1,141 +1,176 @@
...
+  periodSeconds: 10 # @schema required
+  timeoutSeconds: 1 # @schema required
+  successThreshold: 1 # @schema required
+  failureThreshold: 3 # @schema required
+
+# -- (Optional) Override the chart name used in generated resource names
+nameOverride: ""
+
+# -- (Optional) Override the fully qualified name used for generated resource names
+fullnameOverride: ""
```
abhishekdwivedi3060: Move these items at the very beginning

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -31,6 +31,40 @@ Operator uses admission webhooks, which needs TLS certificates. These are issued
 helm install aerospike-kubernetes-operator ./aerospike-kubernetes-operator --set replicas=3
 ```
 
+## Values validation
```
abhishekdwivedi3060: Helmchart readme is meant for end-users and this section seems more of a contribution/dev worfklow.
We can add this in a separate README at the root of helm-charts dir

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -80,6 +116,9 @@ helm install aerospike-kubernetes-operator ./aerospike-kubernetes-operator --set
...
+| `nameOverride`                      | Override the chart name used in generated resource names                                                                                                                                                                  | `""` (nil)                                                                                                         |
+| `fullnameOverride`                  | Override the fully qualified name used for generated resource names                                                                                                                                                       | `""` (nil)                                                                                                         |
+| `webhookServicePort`                | **Deprecated**, use `webhookService.targetPort` instead. Setting this fails chart rendering with a message pointing to the replacement                                                                                    | `null`                                                                                                             |
```
abhishekdwivedi3060: remove this

abhishekdwivedi3060 (changes requested):
Lets start with lenient root level additionalProperties: true to not break anything which have been silently ignored so far. It's fine to keep the object level additionalProperties: false. We can make it stricter in a major version.

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Kubernetes has 30 seconds as the max limit for the `timeoutSeconds` configuration in the validatingwebhookconfiguration. Default value is 10. https://kubernetes.io/docs/reference/kubernetes-api/admissionregistration/validating-webhook-configuration-v1/#ValidatingWebhook

sidntrivedi: Removing this since we are depending on K8s api server for all the K8s specific validations.

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Added back

Makefile
sidntrivedi: Done. Added check to add this flag only when helm version >= 4.

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Removed

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Removed

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Removed

helm-charts/aerospike-kubernetes-operator/values.yaml
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/README.md
sidntrivedi: Moved this section to a new README in the helm-charts/ folder.

helm-charts/aerospike-kubernetes-operator/README.md
sidntrivedi: Removed

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -33,53 +34,56 @@ helm install aerospike-kubernetes-operator ./aerospike-kubernetes-operator --set
...
-| `livenessProbe`                     | Liveliness probe for operator container                                                                                                                                                                                   | `initialDelaySeconds: 15`, `periodSeconds: 20`, `timeoutSeconds: 1`, `successThreshold: 1`, `failureThreshold: 3`  |
-| `readinessProbe`                    | Readiness probe for the operator container                                                                                                                                                                                | `initialDelaySeconds: 5`, `periodSeconds: 10`, `timeoutSeconds: 1`, `successThreshold: 1`, `failureThreshold: 3`   |
+| Name                                        | Description                                                                                                                                                                                                               | Default                                                                                                           |
```
abhishekdwivedi3060, line 37: Missed nameOverride and fullnameOverride

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -33,53 +34,56 @@ helm install aerospike-kubernetes-operator ./aerospike-kubernetes-operator --set
...
+| `operatorImage.tag`                         | Operator image tag                                                                                                                                                                                                        | `4.5.0`                                                                                                           |
+| `operatorImage.pullPolicy`                  | Image pull policy                                                                                                                                                                                                         | `IfNotPresent`                                                                                                    |
+| `imagePullSecrets`                          | Secrets containing credentials to pull Operator image from a private registry                                                                                                                                             | `{}` (nil)                                                                                                        |
```
abhishekdwivedi3060: make imagePullSecrets as []

abhishekdwivedi3060 (dismissed):

helm-charts/aerospike-kubernetes-operator/README.md
sidntrivedi: Done

helm-charts/aerospike-kubernetes-operator/README.md
sidntrivedi: Done

abhishekdwivedi3060 (approved):

tanmayja (approved):

shubhamyelikar99 (approved):

PR #563, opened 2026-07-21
fix(helm): static webhook CA bundle alternative

neeraj-1605:
Keeps cert-manager as the default path while documenting how to supply caBundle via certs.webhook values for environments that manage webhook TLS externally.

# Summary
Keeps cert-manager.io/inject-ca-from as default when certs.webhook.caBundle is empty (no behavior change).
Adds certs.webhook.caBundle under the existing certs.webhook values (removes webhookCerts).
Skips cert-manager CA injection when caBundle is set.
Injects clientConfig.caBundle on all webhooks (9 total: 4 mutating + 5 validating).
For cert-manager-free installs, the client must pre-create the serving cert secret (certs.webhook.webhookServerCertSecretName, default webhook-server-cert) with tls.crt and tls.key — the operator deployment already mounts this secret; no deployment changes needed.
Adds README documentation for base64-encoded caBundle (how to encode the CA PEM and pass it via values).

## Base64-encoded caBundle
certs.webhook.caBundle expects the base64 encoding of the CA certificate PEM that signed the webhook serving certificate — not the serving certificate itself.

### Linux
CA_BUNDLE=$(base64 -w0 ca.crt)
### macOS
CA_BUNDLE=$(base64 -i ca.crt)
Pass via Helm:

helm install aerospike-kubernetes-operator ./aerospike-kubernetes-operator \
  --set certs.webhook.caBundle="${CA_BUNDLE}"
Or in a values file:

certs:
  webhook:
    caBundle: "LS0tLS1CRUdJTi..."   # base64-encoded CA PEM
When caBundle is set, the chart injects clientConfig.caBundle on all webhooks and skips cert-manager.io/inject-ca-from.

# Test plan
helm template with defaults — cert-manager annotation active, no caBundle
helm template --set certs.webhook.caBundle=dGVzdA== — caBundle set on all webhooks, no cert-manager annotation
helm template --set certs.webhook.create=false --set certs.webhook.caBundle=dGVzdA== — no cert-manager Certificate CR, caBundle set on all webhooks, deployment still mounts webhook-server-cert

cc: @deepakpunjabi


helm-charts/aerospike-kubernetes-operator/templates/aerospike-operator-mutating-webhook-configuration.yaml
```diff
@@ -12,6 +12,7 @@ webhooks:
 - admissionReviewVersions:
   - v1
   clientConfig:
+    {{/* caBundle: {{ .Values.webhookCerts.tlsCrt }} */}}
```
deepakpunjabi: Why keep it commented? Values file default behaviour can stay on cert-manager instead of self-signed cert. We shouldn't expect users to uncomment templates for using self-signed certs.

deepakpunjabi: Better to call it `caBundle`, `tlsCrt` is very misleading. Naming it tlsCrt implies the serving cert not the the signing CA.

deepakpunjabi (commented):
This PR does not address the main concern - **making webhook certs configurable**.

- [ ] Ensure cert-manager injection is only enabled when static certs are not present.
- [ ] The chart already has a documented `certs.webhook`. Utilize it instead of adding a top level `webhookCerts` key.
- [ ] Going cert-manager-free also requires the operator's serving cert secret (webhook-server-cert, mounted at `deployment.yaml`.
- [ ] Check with maintainers if this change can be acceptable for other helm components(eg. aerospike-backup-service) in a single PR.
- [ ] Add documentation around base64 encoded caBundle in README.
```
# values.yaml
certs:
  webhook:
    create: true          # cert-manager path
    caBundle: ""           # if set, use static bundle instead
# webhook config metadata
{{- if not .Values.certs.webhook.caBundle }}
  annotations:
    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/aerospike-operator-serving-cert
{{- end }}
# clientConfig
    {{- with .Values.certs.webhook.caBundle }}
    caBundle: {{ . }}
    {{- end }}
```

helm-charts/aerospike-kubernetes-operator/templates/aerospike-operator-mutating-webhook-configuration.yaml
```diff
@@ -1,8 +1,10 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
+  {{- if not .Values.certs.webhook.caBundle }}
```
deepakpunjabi: There are some existing bugs here. 
Let's move above check into _helpers.tpl.
```
{{- if and .Values.certs.webhook.create .Values.certs.webhook.caBundle }}
  {{- fail "certs.webhook.caBundle and certs.webhook.create are mutually exclusive: set certs.webhook.create=false when supplying a static caBundle" }}
  {{- end }}
```

deepakpunjabi: Fix the README example to include --set certs.webhook.create=false.
Explain webhook.create delegates cert creation to cert-manager.

helm-charts/aerospike-kubernetes-operator/templates/aerospike-operator-mutating-webhook-configuration.yaml
```diff
@@ -12,6 +14,9 @@ webhooks:
 - admissionReviewVersions:
   - v1
   clientConfig:
+    {{- with .Values.certs.webhook.caBundle }}
+    caBundle: {{ . }}
```
deepakpunjabi: This `caBundle: {{ . | replace "\n" "" | quote }}` would be better to handle any copy paste issues.

deepakpunjabi: Move duplicate logic to _helpers.tpl.
```
  {{- define "aerospike-kubernetes-operator.webhookCABundle" -}}
  {{- with .Values.certs.webhook.caBundle }}
  caBundle: {{ . | replace "\n" "" | quote }}
  {{- end }}
  {{- end }}
```

deepakpunjabi: template will contain `{{- include "aerospike-kubernetes-operator.webhookCABundle" . | nindent 4 }}`

deepakpunjabi (commented):
Document that creation on serving certs has to be done separately from helm chart as it contains private keys which should not be logged in CI systems.

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -31,6 +31,61 @@ Operator uses admission webhooks, which needs TLS certificates. These are issued
 helm install aerospike-kubernetes-operator ./aerospike-kubernetes-operator --set replicas=3
 ```
 
+## Webhook TLS Configuration
+
+By default, webhook TLS is managed by cert-manager. When `certs.webhook.create` is `true` (default), the chart creates cert-manager `Certificate` resources and relies on `cert-manager.io/inject-ca-from` to populate webhook `caBundle` values.
```
deepakpunjabi, line 36:
```suggestion
By default, webhook TLS is managed by cert-manager. When `certs.webhook.create` is `true` (default), the chart creates cert-manager `Certificate` resources and relies on `cert-manager.io/inject-ca-from` annotation to populate webhook `caBundle` values.
```

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,32 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+
+{{/*
+Inject cert-manager CA when a static caBundle is not supplied.
+*/}}
+{{- define "aerospike-kubernetes-operator.webhookCAInjectionAnnotations" -}}
+{{- if not .Values.certs.webhook.caBundle }}
+annotations:
+  cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/aerospike-operator-serving-cert
+{{- end }}
+{{- end -}}
```
deepakpunjabi, line 83: Don't think this is adding more value. We can revert this.

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -31,6 +31,61 @@ Operator uses admission webhooks, which needs TLS certificates. These are issued
...
+  --namespace <namespace> \
+  --set certs.webhook.create=false \
+  --set certs.webhook.caBundle="${CA_BUNDLE}"
+```
+
+### Base64-encoded caBundle
+
+`certs.webhook.caBundle` expects the **base64 encoding of the CA certificate PEM** that signed the webhook serving certificate. It is not the serving certificate itself.
+
+When set, the chart injects `clientConfig.caBundle` on all webhooks and skips the `cert-manager.io/inject-ca-from` annotation.
```
deepakpunjabi, line 58: No need to explain this. Create a one liner in previous paragraph.

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -31,6 +31,61 @@ Operator uses admission webhooks, which needs TLS certificates. These are issued
...
+
+```sh
+# Linux
+CA_BUNDLE=$(base64 -w0 ca.crt)
+
+# macOS
+CA_BUNDLE=$(base64 -i ca.crt)
+```
+
+The result must be a single-line string with no line breaks.
```
deepakpunjabi, line 70: No need to explain this. Just declare it is base64 encoded.

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -26,6 +26,9 @@ metricsPort: 8443
 certs:
   webhook:
     create: true
+    # Base64-encoded CA certificate for webhook clientConfig.caBundle.
+    # When set, cert-manager inject-ca-from is skipped on webhook configurations.
```
deepakpunjabi, line 30: Make it concise and a single line comment.

abhishekdwivedi3060:
Hey @neeraj-1605  @deepakpunjabi, thank you for working on this.
Considering your use case and some other Helm chart changes we had planned, we’ve created a generic flow to support different types of certificate inputs from users. This new flow should cover your use case as well.

Please take a look at the PR and let us know your thoughts: https://github.com/aerospike/aerospike-kubernetes-operator/pull/596

Will be closing this PR once that feature is released

PR #570, opened 2026-07-28
feat: KO-569 Uplift ABS version from v3.5.0 -> v3.6.1

shubhamyelikar99:
## Summary
Uplifts the Aerospike Backup Service dependency from v3.5.0 to v3.6.1 

## Changes
- **Dependency**: `aerospike-backup-service/v3` v3.5.0 → v3.6.1 in `go.mod`/`go.sum`,
  pulling transitive bumps (`backup-go`, `aws-sdk-go-v2`, `grpc`).
- **Restore status mapping**: ABS 3.6.x renamed restore job statuses
  (`Running`/`Done`/`Failed` → `running`/`success`/`failure`/`canceled`). `statusToPhase`
  now delegates to `dto.ParseJobStatus`, which accepts both the new canonical values and
  the legacy ones, so status reporting works against old and new ABS. Also maps the
  previously unhandled `canceled` state to the `Failed` phase instead of an empty phase.
- **Restore-by-timestamp storage override (BKRS-212)**: removed the hard block that
  rejected `source` on `Timestamp` restores, and added a `source-name` constant. This
  enables cross-region restore, where backup files are copied to local storage but the
  original routine is not configured in the restore-side ABS.
- **Version references**: ABS image pinned to `3.6.1` in samples, Helm values/README, and
  test constants.
- **Tests**: unit coverage for the status mapping (including legacy values, which a live
  3.6.1 ABS can never emit), envtests for the storage override and for `compact` in
  backup policies, and an E2E timestamp restore using `source-name`.

`compact` and `source`/`source-name` are passthrough fields in free-form ABS config and
are intentionally not version-gated, consistent with the ABS v3.4.0 uplift (#417), which
added `use-scan-compression`, `max-concurrent-nodes`, and `rack-list` the same way.

## Testing
- [x] make backup-service-test — 22/22 passed
- [x] make backup-test — passed
- [x] make restore-test — passed (Full/Incremental/Timestamp + data validation)
- [x] `make unit-test` — passes, including the new `TestStatusToPhase`.
- [x] `make env-test` — all 6 suites pass (280 cluster, 13 restore, 10 backup, 4 eviction,
  3 backup service).
- [x] `go build ./...` and `go vet ./...` clean.
- [x] Verified new ABS field "compact" fails if tried with older ABS version v3.5.0

~~~
 aerospike-kubernetes-operator git:(feature/KO-569-add-abs-v3.6.0-support) ✗ k logs -n aerospike -f aerospikebackupservice-b6c8848cc-7dwcc 
2026/07/28 17:16:19 INFO Aerospike Backup Service version=v3.5.0 commit=03f4b5781b368be9867aa68c02b6637902f16795 buildTime=2026-03-10T13:56:43Z
2026/07/28 17:16:19 INFO Read service configuration from file=/etc/aerospike-backup-service/aerospike-backup-service.yml remote=false
2026/07/28 17:16:19 INFO Service configuration:
aerospike-clusters:
  aerospike-aerocluster-backup-test-cluster:
    credentials:
      password: admin123
      user: admin
    seed-nodes:
    - host-name: aerocluster.aerospike.svc.cluster.local
      port: 3000
backup-policies:
  test-policy:
    compact: true
    parallel: 3
backup-routines:
  aerospike-aerocluster-backup-test-routine:
    backup-policy: test-policy
    incr-interval-cron: 0 */8 * * * *
    interval-cron: 0 */5 * * * *
    namespaces:
    - test
    source-cluster: aerospike-aerocluster-backup-test-cluster
    storage: local
service:
  http:
    port: 8081
storage:
  local:
    local-storage:
      path: /tmp/localStorage
Error: failed to load configuration: failed to read configuration: failed to unmarshal configuration: line 11: field compact not found in type dto.BackupPolicy
Usage:
  aerospike-backup-service [flags]

Flags:
  -c, --config string   configuration file path/URL
  -h, --help            help for aerospike-backup-service
  -r, --remote          use remote config file
  -v, --version         version for aerospike-backup-service

2026/07/28 17:16:19 ERROR Command execution failed error="failed to load configuration: failed to read configuration: failed to unmarshal configuration: line 11: field compact not found in type dto.BackupPolicy"

~~~


test/backup_service/backup_service_test.go
```diff
@@ -207,6 +207,20 @@ var _ = Describe(
...
+				policy["compact"] = true
+
+				configBytes, mErr := json.Marshal(config)
+				Expect(mErr).ToNot(HaveOccurred())
+
+				backupService = newBackupServiceWithConfig(backupServiceNamespacedName, configBytes)
+				err = DeployBackupService(k8sClient, backupService)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
```
abhishekdwivedi3060: No need for a separate test. Directly set this field in the getBackupServiceConfMap func that is used everywhere

test/envtests/backup_service/backup_service_webhook_test.go
```diff
@@ -60,4 +62,24 @@ var _ = Describe("AerospikeBackupService validation", func() {
...
+					backupService := buildBackupServiceCR(absNsNm)
+					backupService.Spec.Config = runtime.RawExtension{
+						Raw: backupconfig.MustMarshalConfig(config),
+					}
+
+					Expect(envtests.K8sClient.Create(ctx, backupService)).To(Succeed())
+				})
+			})
+		})
+	})
```
abhishekdwivedi3060: Same here. Simply add compact field in BackupServiceBaseConfig

test/fixtures/backupconfig/config.go
```diff
@@ -36,6 +36,14 @@ var (
 	}
 )
 
+// RestoreOnlyStorageConfig returns a minimal ABS config for restore-side deployments:
+// service, backup-policies, and storage only (no aerospike-clusters or backup-routines).
+// Used to exercise BKRS-212 cross-region timestamp restores where the routine is not
+// configured locally but storage is overridden via source-name.
+func RestoreOnlyStorageConfig() map[string]interface{} {
```
abhishekdwivedi3060: Name is confusing. Rename it to something like BackupServiceConfigForRestore

test/restore/test_utils.go
```diff
@@ -272,6 +274,30 @@ func getRestoreConfigWithTLSInMap(backupPath string) map[string]interface{} {
 	return restoreConfig
 }
 
+func getTimestampRestoreConfigWithStorageOverride(backupPath string) (map[string]interface{}, error) {
```
abhishekdwivedi3060: Try to leverage getTimeStampRestoreConfigBytes func. Maybe split it into getTimeStampRestoreConfig and getTimeStampRestoreConfigBytes

test/restore/restore_test.go
```diff
@@ -102,13 +102,17 @@ var _ = Describe(
...
 
-					err = createRestore(k8sClient, restore)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(ContainSubstring("source field is not allowed in restore config"))
+					configBytes, mErr := getRestoreConfBytes(restoreConfig)
+					Expect(mErr).ToNot(HaveOccurred())
+
+					restore = newRestoreWithConfig(restoreNsNm, asdbv1beta1.Timestamp, configBytes)
+
+					err = k8sClient.Create(testCtx, restore)
```
abhishekdwivedi3060: Should use createRestore() func instead

test/envtests/restore/restore_webhook_test.go
```diff
@@ -180,3 +180,57 @@ var _ = Describe("AerospikeRestore CRD schema marker validation", Ordered, func(
...
+		restoreNsNm = uniqueNamespacedName("restore-config")
+		absNsNm = uniqueNamespacedName("abs-restore-config")
+	})
+
+	AfterEach(func() {
+		deleteRestore(ctx, restoreNsNm)
+		backupconfig.DeleteStubBackupService(ctx, envtests.K8sClient, absNsNm)
+	})
+
+	Context("timestamp restore storage override", func() {
```
abhishekdwivedi3060: Follow the hierarchy and convention from Context at line 43

abhishekdwivedi3060 (changes requested):

internal/webhook/v1beta1/aerospikerestore_webhook.go
```diff
@@ -147,10 +146,6 @@ func validateRestoreConfig(k8sClient client.Client, restore *asdbv1beta1.Aerospi
 	case asdbv1beta1.Timestamp:
 		var restoreRequest dto.RestoreTimestampRequest
 
-		if _, ok := restoreConfig[asdbv1beta1.SourceKey]; ok {
```
tanmayja, line 150: Will this be helpful in latest AKO but old ABS?

internal/controller/restore/reconciler.go
```diff
@@ -334,16 +335,19 @@ func (r *SingleRestoreReconciler) cancelRestoreJob() error {
...
+	switch jobStatus {
+	case dto.RestoreRunning:
 		return asdbv1beta1.AerospikeRestoreInProgress
-
-	case "Failed":
+	case dto.RestoreSuccess:
+		return asdbv1beta1.AerospikeRestoreCompleted
+	case dto.RestoreFailure, dto.RestoreCanceled:
 		return asdbv1beta1.AerospikeRestoreFailed
+	default:
```
tanmayja, line 350: we should at least log a warning here, because if we are rwaching here, that means a new status has been added in dto which we are not handling

config/samples/aerospikerestore.yaml
```diff
@@ -25,3 +25,32 @@ spec:
...
+#       label: destinationCluster
+#       credentials:
+#         password: admin123
+#         user: admin
+#       seed-nodes:
+#         - host-name: aerocluster.test.svc.cluster.local
+#           port: 3000
+#     policy:
+#       parallel: 3
+#     source-name: local
```
tanmayja, line 54: does source and source-name mutual exclusive?
if yes, can we add a testcase around it if already not present.

test/fixtures/backupconfig/config.go
tanmayja: do we really need this function at all?

test/fixtures/backupconfig/config.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	BackupServiceImage  = "aerospike/aerospike-backup-service:3.5.0"
+	BackupServiceImage  = "aerospike/aerospike-backup-service:3.6.1"
```
tanmayja, line 15: do we really need to maintain this variable at 2 places?

test/restore/test_utils.go
```diff
@@ -272,6 +274,30 @@ func getRestoreConfigWithTLSInMap(backupPath string) map[string]interface{} {
...
+	restoreConfig := getRestoreConfigInMap(backupPath)
+	delete(restoreConfig, asdbv1beta1.SourceKey)
+	delete(restoreConfig, asdbv1beta1.BackupDataPathKey)
+
+	parts := strings.Split(backupPath, "/")
+	if len(parts) < 4 {
```
tanmayja: The guard must be < 6 (to safely access index len-5) not < 4. With exactly 4 parts, len(parts)-5 == -1 → index panic on line 295

config/samples/aerospikerestore.yaml
abhishekdwivedi3060: Yes, they are mutually exclusive. ABS validation pkg handles that. Confirm this once.
We only add  TCs related to our webhook webhook

internal/webhook/v1beta1/aerospikerestore_webhook.go
abhishekdwivedi3060: This is a grey area in general, not specific to this.
Like the new compact flag will pass the validation with old ABS but fail while applying.
We need to think about ABS version specific schema validation similar to server.
@shubhamyelikar99 Did we create a jira around this?

shubhamyelikar99: yes, https://aerospike.atlassian.net/browse/KO-607

test/fixtures/backupconfig/config.go
shubhamyelikar99: Agreed — removed. BackupServiceBaseConfig() already provides config without backup-routines, which is what this test needs.

config/samples/aerospikerestore.yaml
shubhamyelikar99:
~~~
// StorageConfig aggregates the storage configuration.
// It is intended to be embedded into DTOs that require Storage configuration.
type StorageConfig struct {
	// The details of the storage configuration.
	// Mutually exclusive with 'source-name'.
	Storage *Storage `json:"source,omitempty"`
	// Link to one of preconfigured storages.
	// Mutually exclusive with 'source'.
	Name string `json:"source-name,omitempty" extensions:"x-nullable"`
}
~~~
They are mutually exclusive fields

test/fixtures/backupconfig/config.go
shubhamyelikar99: Removed the duplicate BackupServiceImage from test/backup_service/test_utils.go. Tests now use backupconfig.BackupServiceImage as the single source of truth.

test/restore/test_utils.go
shubhamyelikar99: Thanks for catching this. You're right — with the earlier len(parts) < 4 guard, a 4-segment path would pass validation and then panic on parts[len(parts)-5] (e.g. parts[-1]).

Updated the guard to len(parts) < 5, since parts[len-5] needs at least 5 segments. Our test paths are 5-part (<routine>/backup/<timestamp>/data/<test>), so this rejects too-short paths with a clean error and still works for the real E2E path.

Used < 5 rather than < 6 because the documented path shape only has 5 segments; < 6 would incorrectly reject valid paths in the suite.

For example, our test path is:

`test-sample-backup-test-routine/backup/1722353745635/data/test` → **5 parts**

- `parts[len-5]` → `parts[0]` → routine ✓  
- `parts[len-3]` → `parts[2]` → timestamp ✓  

So we only need **at least 5 parts** → guard is `len(parts) < 5`.

If we used `< 6`, this valid 5-part path would fail (`5 < 6` → error), even though it’s the path the suite actually uses.

Example of a bad path with **4 parts** (`a/b/c/d`): old guard `< 4` let it through and `parts[len-5]` = `parts[-1]` panicked. With `< 5`, we return an error instead.

test/restore/restore_test.go
```diff
@@ -102,13 +100,19 @@ var _ = Describe(
 					Expect(err.Error()).To(ContainSubstring("empty field validation error: \"time\" required"))
 				})
 
-				It("Should fail when source field is given for Timestamp restore type", func() {
-					restore, err = newRestore(restoreNsNm, asdbv1beta1.Timestamp)
-					Expect(err).ToNot(HaveOccurred())
+				It("Should accept source override for Timestamp restore type", func() {
```
abhishekdwivedi3060: It's valid operation. Should move it under "When doing valid operations" Context

abhishekdwivedi3060 (dismissed):

abhishekdwivedi3060 (dismissed):

tanmayja (dismissed):

abhishekdwivedi3060 (approved):

PR #577, opened 2026-08-06
enhancement: persisting error phase in aerospike CR. [KO-541] [KO-616]

tanmayja:
Persist error phase and improve ignorable-pod recovery

Persist the Error phase — previously, the operator would overwrite the cluster's Error phase with InProgress at the start of every reconcile cycle, making it impossible to tell from the CR status that something was wrong. Now the Error phase is preserved until the user either fixes the underlying problem or marks the pod as ignorable.

Complete reconciliation after one recovery attempt — when a server-failed pod is marked ignorable, the operator used to delete it and immediately requeue, repeating the cycle on every reconcile. Now it deletes the pod once, waits for it to come back up, and only requeues if the pod hasn't failed or recovered within that window — avoiding unnecessary churn.

Approach doc: https://aerospike.atlassian.net/wiki/spaces/KO/pages/5547327506/Approach+Persist+Error+Phase+and+Ignorable+Pod+Recovery+Completion+KO-541


internal/controller/cluster/statefulset.go
```diff
@@ -376,9 +388,8 @@ func (r *SingleClusterReconciler) waitForSTSToBeReady(
 			}
 
 			if err := utils.CheckPodFailed(pod); err != nil {
-				return fmt.Errorf(
-					"check StatefulSet pod %s: %w", utils.NamespacedName(st.Namespace, podName), err,
-				)
+				return fmt.Errorf("pod %s failed: %w",
+					utils.NamespacedName(st.Namespace, podName), err)
```
sidntrivedi, line 377: Should we return the error in form of `stsNotReadyError` here?

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
 				}
 
 				r.Log.Info("Deleted Pod", "pod", utils.GetNamespacedName(&podList.Items[idx]))
+
+				// Block for up to 3 minutes polling every 10s. Exit early if the
```
sidntrivedi: How did we decide to wait for 3 minutes?

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
 	}
 
-	if anyPodFailed {
-		r.Log.Info("Found failed/pending Pod(s), requeuing")
-	} else {
-		r.Log.Info("Found ignorable Pod(s), requeuing")
+	if !anyPodFailed {
+		// No failures at all — ignorable pods are present but healthy. Requeue at a
+		// slower cadence to keep monitoring without overwhelming the API server.
+		r.Log.Info("Found ignorable pod(s), requeuing")
```
sidntrivedi: Does it make sense to mention in the log "Found all ignorable pod(s) to be healthy, setting requeuing interval to 60s"

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
 				}
 
 				r.Log.Info("Deleted Pod", "pod", utils.GetNamespacedName(&podList.Items[idx]))
+
+				// Block for up to 3 minutes polling every 10s. Exit early if the
+				// server container fails (no point waiting further) or becomes ready.
+				// If the pod fails or recovers, requeueInterval stays 0 — nothing
+				// left to monitor for this pod.
+				if podReady, err := r.waitForIgnorablePodRecovery(podList.Items[idx].Name); err != nil {
+					r.Log.Error(err, "Ignorable pod failed during recovery", "pod", podList.Items[idx].Name)
```
sidntrivedi: Should we set the requeueInterval here as well? Otherwise in case error of the returned value being `(false, err)` and `anyPodFailed` being `true`, `common.ReconcileSuccess()` is returned.

internal/controller/cluster/rack.go
```diff
@@ -1244,7 +1285,7 @@ func (r *SingleClusterReconciler) scaleDownRack(
 		r.aeroCluster, corev1.EventTypeNormal, "RackScaledDown",
 		eventRackScaleMessage(
 			"Scaled down", rackState.Rack.ID,
-			utils.GetNamespacedNameString(found), *found.Spec.Replicas, desiredSize,
+			utils.GetNamespacedNameString(found), scaleDownTargetSize, *found.Spec.Replicas,
```
sidntrivedi: `found` is updated to the new spec in line 1264. Here, both `scaleDownTargetSize` and `found.Spec.Replicas` would be same value.

internal/controller/cluster/rack.go
```diff
@@ -2274,7 +2315,7 @@ func (r *SingleClusterReconciler) reconcileRevisionChangedRacks(
 
 	oldSts, err := r.getSTS(ctx, oldRack)
 	if err != nil {
-		if !errors.IsNotFound(err) {
+		if !k8serrors.IsNotFound(err) {
```
sidntrivedi, line 2311: In the same function:

```
	if asdbv1.GetBool(r.aeroCluster.Spec.IgnoreSidecarFailure) {
		if err := r.waitForAllAerospikeServersReady(ctx, ignorablePodNames); err != nil {
		}
	} else {
		if err := r.waitForAllSTSToBeReady(ctx, ignorablePodNames); err != nil {
		}
	}
```

Should we use the `stsNotReadyError` error in the above snippet as well to differentiate between the state where STS is booting and where there's actual error?

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
+	const (
+		maxRetries    = 18 // 18 * 10s = 3 minutes
+		retryInterval = 10 * time.Second
+	)
+
+	pod := &corev1.Pod{}
+
+	for i := 0; i < maxRetries; i++ {
+		if err := r.Get(
+			context.TODO(),
```
sidntrivedi: Instead of initializing a new context here, we can pass the `ctx` from the reconcile function to this.

internal/controller/cluster/reconciler.go
```diff
@@ -112,9 +113,11 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
 		return reconcile.Result{}, nil
 	}
 
-	// Set the status to AerospikeClusterInProgress before starting any operations
-	if err := r.setStatusPhase(ctx, asdbv1.AerospikeClusterInProgress); err != nil {
-		return reconcile.Result{}, err
+	// but only if the cluster is not already in an error state.
```
sidntrivedi, line 116: I guess the first half of comment got removed. Earlier comment:

	// Set the status to AerospikeClusterInProgress before starting any operations

internal/controller/cluster/statefulset.go
tanmayja: stsNotReadyError has a very specific semantic: the pod is still starting up and just ran out of wait time — it's healthy but slow. CheckPodFailed at line 390 is the opposite case: the pod has actively entered a failure state (CrashLoopBackOff, OOMKilled, etc.).

internal/controller/cluster/reconciler.go
tanmayja: To match the same timeout already used by waitForSTSPodsServerReady and waitForSTSToBeReady in statefulset.go — both also use podStatusMaxRetry = 18 and podStatusRetryInterval = 10s. The idea was to keep the ignorable-pod recovery window consistent with how long the operator already waits for any other pod to become ready in the normal reconcile path.

internal/controller/cluster/reconciler.go
tanmayja: Intentional — requeueing here would just spin in a delete→fail loop. The next reconcile triggers naturally when the the user updates the CR. Added a comment to clarify.

internal/controller/cluster/rack.go
shubhamyelikar99, line 2371: waitForAllRacksReady correctly distinguishes slow startup from hard failure via errors.As(waitErr, &stsNotReadyError) and requeues for the former while setting Error only for real failures (see ~L219–243 in this file).

This block in reconcileRevisionChangedRacks does not apply the same pattern — both branches wrap all wait errors in ReconcileError. Since waitForAllSTSToBeReady / waitForAllAerospikeServersReady delegate to the updated per-STS wait functions, a pod that is still starting (returns stsNotReadyError after the 3-minute poll) will incorrectly transition the cluster to Error during rack revision migration.

Suggestion: mirror the handling in waitForAllRacksReady:

```
if err := r.waitForAllSTSToBeReady(ctx, ignorablePodNames); err != nil {
    var notReady *stsNotReadyError
    if errors.As(err, &notReady) {
        return common.ReconcileRequeueAfter(1)
    }
}
```
Same for the waitForAllAerospikeServersReady branch.

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
+		if utils.IsAerospikeServerReady(pod) {
+			r.Log.Info("Ignorable pod server container is ready", "pod", podName)
+			return true, nil
+		}
+
+		r.Log.V(1).Info("Waiting for ignorable pod to recover", "pod", podName, "attempt", i+1)
+		time.Sleep(retryInterval)
+	}
+
+	return false, nil
```
shubhamyelikar99: waitForIgnorablePodRecovery blocks the reconcile worker for up to 3 minutes per deleted pod using in-loop time.Sleep (18 × 10s). With multiple ignorable pods this runs sequentially, so one reconcile can hold a worker for several minutes while other clusters wait.

I get why we wait in-process: a naive delete → ReconcileRequeueAfter would re-delete on the next pass when the recreated pod still looks failed, recreating the master loop — and pod status changes alone won't re-trigger reconcile. ignorablePodNames also can't gate re-delete; it only marks budget-eligible pods, not "already deleted, monitoring only."

Not blocking for KO-541. If we want to avoid worker blocking later, consider recovery-pending state on the CR + requeue, with delete skipped on subsequent passes.

internal/controller/cluster/rack.go
shubhamyelikar99: Use const RequeueIntervalSeconds60

internal/controller/cluster/statefulset.go
```diff
@@ -246,6 +246,18 @@ func (r *SingleClusterReconciler) deleteSTS(ctx context.Context, st *appsv1.Stat
 	return r.Delete(ctx, st)
 }
 
+// stsNotReadyError is returned when the wait loop exhausts all retries but the
+// pod is still starting up (not a failure — just slow). This is the only error
+// that should result in a requeue; everything else is a hard ReconcileError.
+type stsNotReadyError struct {
```
shubhamyelikar99: DO we have a unit test asserting timeout returns *stsNotReadyError, crash returns plain error?

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
 				}
 
 				r.Log.Info("Deleted Pod", "pod", utils.GetNamespacedName(&podList.Items[idx]))
+
+				// Block for up to 3 minutes polling every 10s. Exit early if the
+				// server container fails (no point waiting further) or becomes ready.
+				// If the pod fails or recovers, requeueInterval stays 0 — nothing
+				// left to monitor for this pod.
+				if podReady, err := r.waitForIgnorablePodRecovery(podList.Items[idx].Name); err != nil {
```
abhishekdwivedi3060: Can we not delete all the ignorable failed pod in the loop and wait for them together after the loop. I think we can leverage the existing waitForAllAerospikeServersReady func instead of new func. Check this once

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
+			if apierrors.IsNotFound(err) {
+				// Pod is still terminating / not yet recreated — keep waiting.
+				time.Sleep(retryInterval)
+				continue
+			}
+
+			return false, fmt.Errorf("failed to get pod %s during recovery wait: %w", podName, err)
+		}
+
+		if err := utils.CheckPodFailed(pod); err != nil {
```
abhishekdwivedi3060: This should be server failed container check

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
-	if anyPodFailed {
-		r.Log.Info("Found failed/pending Pod(s), requeuing")
-	} else {
-		r.Log.Info("Found ignorable Pod(s), requeuing")
+	if !anyPodFailed {
+		// No failures at all — ignorable pods are present but healthy. Requeue at a
+		// slower cadence to keep monitoring without overwhelming the API server.
+		r.Log.Info("Found ignorable pod(s), requeuing")
+
+		requeueInterval = asdbv1.RequeueIntervalSeconds60
```
abhishekdwivedi3060: Why 60s here? This is only for use-case where pods were in ignorable during reconcile entry and recovered by the time we reached in recoverIgnorablePods func. This will be done once this requeue succeeds. Do you see a use-case where this will requeue multiple times

internal/controller/cluster/rack.go
```diff
@@ -707,7 +748,7 @@ func (r *SingleClusterReconciler) reconcileRack(
 			(!r.IsStatusEmpty() && len(r.aeroCluster.Status.RackConfig.Racks) != len(r.aeroCluster.Spec.RackConfig.Racks)) {
 			if res = r.setMigrateFillDelay(
 				ctx, r.getClientPolicy(ctx), &rackState.Rack.AerospikeConfig, false,
-				nil,
+				ignorablePodNames,
```
abhishekdwivedi3060, line 743: Can we update an existing TC with ignorablePods field set and MFD non-zero?

test/cluster/batch_scaledown_pods_test.go
```diff
@@ -120,6 +120,10 @@ var _ = Describe("BatchScaleDown", func() {
 			aeroCluster.Spec.Size--
 			Expect(k8sClient.Update(ctx, aeroCluster)).ToNot(HaveOccurred())
 
+			By("Verifying cluster transitions to Error — reconciler detects server-failed pod, blocks scale-down")
+			Expect(waitForClusterPhase(k8sClient, ctx, clusterNamespacedName,
+				asdbv1.AerospikeClusterError)).ToNot(HaveOccurred())
+
```
abhishekdwivedi3060: Now with this check, do we need the below waitForAerospikeCluster check? Seems redundant to me.
Check at all where we added this new check

test/cluster/sidecar_failure_test.go
abhishekdwivedi3060: Check this comment as per test-case. Seems wrong.
Also check the comment for IT, seems outdated

test/cluster/cluster_test.go
```diff
@@ -996,6 +996,78 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+					// Deploy with exactly nodeCount pods first so the cluster comes up healthy.
+					aeroCluster := createNonSCDummyAerospikeCluster(pendingClusterNN, n)
+					aeroCluster.Spec.PodSpec.MultiPodPerHost = ptr.To(false)
+					randomizeServicePorts(aeroCluster, false, GinkgoParallelProcess())
+					Expect(DeployCluster(k8sClient, ctx, aeroCluster)).ToNot(HaveOccurred())
+
+					defer func() {
+						_ = DeleteCluster(k8sClient, ctx, aeroCluster)
+						_ = CleanupPVC(k8sClient, namespace, pendingClusterName)
+					}()
```
abhishekdwivedi3060: We can use the clusterName and cleanup from the top level vars and AfterEach()

test/cluster/cluster_test.go
```diff
@@ -996,6 +996,78 @@ func clusterWithMaxIgnorablePod(ctx goctx.Context) {
...
+						"expected at least one Pending pod after scale-up beyond node count")
+
+					By("Verifying cluster reaches Completed — pending pod treated as ignorable")
+
+					Eventually(func(g Gomega) {
+						cluster, clusterErr := getCluster(k8sClient, ctx, pendingClusterNN)
+						g.Expect(clusterErr).ToNot(HaveOccurred())
+						g.Expect(cluster.Status.Phase).To(Equal(asdbv1.AerospikeClusterCompleted),
+							"cluster must be Completed with the pending pod treated as ignorable")
+					}, 2*time.Minute, retryInterval).Should(Succeed())
```
abhishekdwivedi3060: We can use waitForClusterPhase  or waitForAerospikeCluster here

internal/controller/cluster/rack.go
```diff
@@ -57,15 +58,26 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
 		racksToDelete, "ignorablePods", ignorablePodNames.UnsortedList(),
 	)
 
-	// Handle failed racks (integrates both normal racks and revision-changed racks)
+	// Handle failed pods for each configured rack. For revision-changed racks
+	// the current-revision STS either does not exist yet or has 0 replicas, so
+	// handleFailedPodsInRack would see no pods; instead we check the old-revision
+	// STS pods directly and skip the new-revision STS lookup entirely.
```
abhishekdwivedi3060: This comment seems wrong. There can be pods of current-revision STS in running/pending/failed state at this point

internal/controller/cluster/rack.go
```diff
@@ -78,6 +90,15 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
 		}
 	}
 
+	// All racks either passed handleFailedPodsInRack or had their old-revision
+	// pods verified as healthy. Restore InProgress so the rest of the reconcile
+	// loop can proceed.
+	if r.aeroCluster.Status.Phase != asdbv1.AerospikeClusterInProgress {
+		if err = r.setStatusPhase(ctx, asdbv1.AerospikeClusterInProgress); err != nil {
```
abhishekdwivedi3060: setStatusPhase already handles this check  r.aeroCluster.Status.Phase != phase internally. It'll be a no-op when phase is same

abhishekdwivedi3060: setStatusPhase here is called in between the reconcile loop and this func will update the main aerocluster object with the laster user-spec. We've seen similar race conditions with ACL in the past

test/cluster/sidecar_failure_test.go
```diff
@@ -506,8 +515,7 @@ var _ = Describe("SidecarFailure", func() {
 			By("Verifying rolling restart is stuck in InProgress — sidecar failure blocks readiness wait")
```
abhishekdwivedi3060:
```suggestion
			By("Verifying rolling restart is stuck in Error — sidecar failure blocks readiness wait")
```

internal/controller/cluster/rack.go
```diff
@@ -2371,6 +2412,54 @@ func (r *SingleClusterReconciler) reconcileRevisionChangedRacks(
...
+
+	if len(serverFailedPods) != 0 {
+		r.Log.Info("Rack has server-failed Pods",
+			"rackID", rackID, "rackRevision", rackRevision,
+			"serverFailedPods", getPodNames(serverFailedPods))
+
+		return common.ReconcileError(
+			fmt.Errorf("rack %d-%s has server-failed pods: %s",
+				rackID, rackRevision, strings.Join(getPodNames(serverFailedPods), ", ")),
+		)
```
abhishekdwivedi3060: I think as per new convention we don't log the error when we return the reconcile error

internal/controller/cluster/rack.go
```diff
@@ -2371,6 +2412,54 @@ func (r *SingleClusterReconciler) reconcileRevisionChangedRacks(
...
+
+		if len(sidecarNotReadyPods) != 0 {
+			r.Log.Info("Rack has sidecar-not-ready Pods",
+				"rackID", rackID, "rackRevision", rackRevision,
+				"sidecarNotReadyPods", getPodNames(sidecarNotReadyPods))
+
+			return common.ReconcileError(
+				fmt.Errorf("rack %d-%s has sidecar-not-ready pods: %s",
+					rackID, rackRevision, strings.Join(getPodNames(sidecarNotReadyPods), ", ")),
+			)
```
abhishekdwivedi3060: 1. same error comment here
2. Sidecar is not failed but in not ready, should be requeue? In actual rack failed pod handling we error out in case of failed pod but requeue in case of stillNotReadyPods

internal/controller/cluster/reconciler.go
```diff
@@ -314,19 +317,84 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
+	}
+
+	if requeueInterval == 0 {
+		// All deleted pods either failed or recovered — nothing left to monitor.
+		return common.ReconcileSuccess()
 	}
 
+	r.Log.Info("Found failed/pending pod(s), requeuing")
```
abhishekdwivedi3060: This log is printed for !anyPodFailed condition also along with its own log which makes it confusing

internal/controller/cluster/rack.go
```diff
@@ -145,61 +166,81 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
+				return common.ReconcileRequeueAfter(1)
 			}
 
-			r.Log.Error(
-				waitErr, waitMsg,
-				"statefulSet", utils.GetNamespacedName(found),
-			)
+			// Pod failure or API error — hard error.
+			r.Log.Error(waitErr, "Error while waiting for StatefulSet to be ready",
+				"statefulSet", utils.GetNamespacedName(found))
```
abhishekdwivedi3060: redundant with reconcile error

internal/controller/cluster/statefulset.go
abhishekdwivedi3060: Instead of creating separate structs for sentinel error, why not directly create a new error type like
var errPodNotReady = errors.New("pod not ready yet")

internal/controller/cluster/statefulset.go
```diff
@@ -312,10 +324,10 @@ func (r *SingleClusterReconciler) waitForSTSPodsServerReady(
...
 
 		if !isReady {
-			return fmt.Errorf(
-				"server container in Pod %s did not become ready, status: %v",
-				utils.NamespacedName(st.Namespace, podName), pod.Status.ContainerStatuses,
-			)
+			return &stsNotReadyError{
+				podName: utils.NamespacedName(st.Namespace, podName),
+				reason:  fmt.Sprintf("server container not ready, containerStatuses: %v", pod.Status.ContainerStatuses),
+			}
```
abhishekdwivedi3060: Change these as per new error type by wrapping the error

abhishekdwivedi3060 (changes requested):

internal/controller/cluster/rack.go
tanmayja: using patch just for this case.

internal/controller/common/result.go
```diff
@@ -28,6 +34,10 @@ func ReconcileSuccess() ReconcileResult {
 	return ReconcileResult{IsSuccess: true, Result: reconcile.Result{}}
 }
 
+func ReconcileRequeue() ReconcileResult {
+	return ReconcileResult{Result: reconcile.Result{Requeue: true}}
```
abhishekdwivedi3060: Requeue is deprecated. It says to use RequeueAfter instead

internal/controller/common/result.go
```diff
@@ -1,11 +1,17 @@
...
+	"errors"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// ErrSTSNotReady is returned when a StatefulSet wait loop exhausts all retries
+// but the pod is still starting up (not a failure — just slow). Callers branch
+// on this with errors.Is to requeue rather than hard-error.
+var ErrSTSNotReady = errors.New("STS pod not ready yet")
```
abhishekdwivedi3060: We use full name for resources - StatefulSet

internal/controller/cluster/rack.go
```diff
@@ -92,10 +91,14 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
-	// loop can proceed.
+	// loop can proceed. Use a patch to avoid overwriting r.aeroCluster with a
+	// fresh Get mid-reconcile.
 	if r.aeroCluster.Status.Phase != asdbv1.AerospikeClusterInProgress {
-		if err = r.setStatusPhase(ctx, asdbv1.AerospikeClusterInProgress); err != nil {
-			return common.ReconcileError(err)
+		patch := client.MergeFrom(r.aeroCluster.DeepCopy())
+		r.aeroCluster.Status.Phase = asdbv1.AerospikeClusterInProgress
+
+		if err = r.Client.Status().Patch(ctx, r.aeroCluster, patch); err != nil {
```
abhishekdwivedi3060: This again would overwrite the r.aerocluster. We need to use a placeholder. Any reason for not updating the r.setStatusPhase phase? Also we are missing the update the metric with approach r.addClusterPhaseMetric().

internal/controller/cluster/rack.go
```diff
@@ -2359,10 +2358,20 @@ func (r *SingleClusterReconciler) reconcileRevisionChangedRacks(
...
+				return common.ReconcileRequeueAfter(1)
+			}
+
 			return common.ReconcileError(err)
 		}
 	} else {
 		if err := r.waitForAllSTSToBeReady(ctx, ignorablePodNames); err != nil {
+			if errors.Is(err, common.ErrSTSNotReady) {
+				r.Log.Error(err, "Failed to wait for StatefulSet to be ready after rack migration, will requeue")
+				return common.ReconcileRequeueAfter(1)
```
abhishekdwivedi3060: nit: I think instead of "Failed to" we should say something like "Timed out waiting"

internal/controller/cluster/rack.go
```diff
@@ -2446,20 +2451,52 @@ func (r *SingleClusterReconciler) checkRackPodsHealthy(
...
+			stillNotReady = append(stillNotReady, pod)
+		}
+	}
+
+	return
+}
+
+// handleFailedPodsInRack inspects the pods for a rack and drives recovery in
+// two phases:
+//  1. Server-failed pods are passed to reconcileRack (bypassing batching and
```
abhishekdwivedi3060: Fix comment, now we consider batching even for failed pod

test/cluster/rack_revision_test.go
```diff
@@ -317,6 +317,127 @@ var _ = Describe(
...
+
+								By("Verifying cluster transitions to Error — checkRackPodsHealthy detects " +
+									"crashed sidecar on old-revision pod")
+								Expect(waitForClusterPhase(k8sClient, ctx, clusterNamespacedName,
+									asdbv1.AerospikeClusterError)).ToNot(HaveOccurred())
+
+								By("Setting IgnoreSidecarFailure=true to unblock migration")
+
+								aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+								Expect(err).ToNot(HaveOccurred())
```
abhishekdwivedi3060: nit: redundant

test/cluster/rack_revision_test.go
```diff
@@ -317,6 +317,127 @@ var _ = Describe(
...
+
+								err = validateRackRevisionCleanup(k8sClient, ctx, aeroCluster, []int{1}, versionV1)
+								Expect(err).ToNot(HaveOccurred())
+							},
+						)
+
+						It(
+							"Should not block migration when old-revision pod has crashing sidecar "+
+								"and IgnoreSidecarFailure is true",
```
abhishekdwivedi3060: Do we need separate TC for this? Above TC already covers IgnoreSidecarFailure is true recovery

internal/controller/cluster/reconciler.go
```diff
@@ -275,56 +279,138 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
+				if apierrors.IsNotFound(err) {
+					// Still terminating / not yet recreated — keep waiting.
 					continue
 				}
 
-				// Pod has failed and grace period is over
-				if err := r.createOrUpdatePodServiceIfNeeded(ctx, []string{podList.Items[idx].Name}); err != nil {
-					return common.ReconcileError(err)
-				}
+				return false, fmt.Errorf("get pod %s during recovery wait: %w", podName, err)
```
abhishekdwivedi3060: Check for logging consistency like Pod, full NsNm etc.
Check at other places as well
```suggestion
				return false, fmt.Errorf("get Pod %s during recovery wait: %w", podNsNm, err)
```

internal/controller/cluster/reconciler.go
```diff
@@ -275,56 +279,138 @@ func (r *SingleClusterReconciler) recoverIgnorablePods(
...
+	if !allReady && waitErr == nil {
+		// Timeout — some pods still not ready after 3 minutes.
+		r.Log.Info("Some ignorable pods not ready after 3 minutes, will requeue")
+
+		return common.ReconcileRequeueAfter(asdbv1.RequeueIntervalSeconds10)
+	}
+
+	if waitErr != nil {
+		// Re-failed — don't requeue; next reconcile will be triggered by a CR change.
+		r.Log.Error(waitErr, "One or more ignorable pods failed during recovery")
```
abhishekdwivedi3060: nit:
```suggestion
		r.Log.Error(waitErr, "One or more ignorable pods failed during recovery, won't requeu")
```

abhishekdwivedi3060 (dismissed):

test/cluster/rack_revision_test.go
```diff
@@ -317,6 +317,65 @@ var _ = Describe(
 							),
 							Serial,
 						)
+
+						FIt(
```
sidntrivedi: Lets change this to It before merging.

sidntrivedi: Done

test/cluster/rack_revision_test.go
```diff
@@ -182,12 +183,44 @@ var _ = Describe(
...
+								// Native StatefulSet pod recreation while the new-revision rack
+								// finishes coming up can legitimately cause extra Error/InProgress
+								// blips unrelated to the maxIgnorablePods fix, so only the terminal
+								// state is asserted here — the Error starting point was already
+								// confirmed above via blockedCluster.Status.Phase.
+								err = waitForAerospikeCluster(
+									k8sClient, ctx, aeroCluster, int(aeroCluster.Spec.Size), retryInterval,
+									getTimeout(aeroCluster.Spec.Size),
+									[]asdbv1.AerospikeClusterPhase{asdbv1.AerospikeClusterCompleted},
+								)
```
abhishekdwivedi3060: How is it different from the previous updateCluster() call?

test/cluster/cluster_test.go
```diff
@@ -655,6 +655,44 @@ func ScaleDownWithMigrateFillDelay(ctx goctx.Context) {
...
+						nil, healthyPodName)
+					Expect(err).ToNot(HaveOccurred())
+
+					aeroCluster, err = getCluster(k8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(aeroCluster.Status.Size).To(Equal(int32(4)),
+						"scale-down must not proceed while blocked in Error")
+				},
+			)
 		},
```
abhishekdwivedi3060: I think this TC is already covered batch_scaledown_pods_test file. Confirm this once.
If required add the additional checks in old TC to merge both

internal/controller/cluster/rack.go
sidntrivedi: This is already fixed.

internal/controller/cluster/reconciler.go
sidntrivedi: Added Pod with namespace data in the log.

test/cluster/rack_revision_test.go
sidntrivedi: Reverted to `updateCluster()`.

test/cluster/cluster_test.go
sidntrivedi: Merged the cases in `batch_scaledown_pods_test.go` file.

internal/controller/cluster/reconciler.go
```diff
@@ -551,6 +551,7 @@ func (r *SingleClusterReconciler) updateStatus(ctx context.Context) error {
 
 func (r *SingleClusterReconciler) setStatusPhase(ctx context.Context, phase asdbv1.AerospikeClusterPhase) error {
 	if r.aeroCluster.Status.Phase == phase {
+		r.addClusterPhaseMetric()
```
abhishekdwivedi3060, line 554: Add comment for this in the subsequent PR

abhishekdwivedi3060 (approved):

PR #579, opened 2026-08-11
feat: Add support for AerospikeCluster conditions in the status [KO-515] 

abhishekdwivedi3060:
## What

Adds seven standard `metav1.Condition` entries to `AerospikeCluster.status`, so cluster state is machine-readable instead of only inferable from `status.phase`.

- **`Ready`** - the operator has successfully applied the current spec. Reflects reconcile outcome, not cluster health.
- **`ScalingUp`** / **`ScalingDown`** - pods being added to / removed from at least one rack. Rack-scoped, so both can be `True` during a rack redistribution.
- **`Upgrading`** - container images being updated on at least one rack.
- **`RollingRestart`** - pods being restarted for config changes that cannot be applied dynamically.
- **`Paused`** - reconciliation suspended via `spec.paused`.
- **`RackRevisionRollingOut`** - rack revision migration being done 

`Ready` and `Paused` are also exposed as printer columns for `kubectl get`.

## Testing

- Unit tests for the status writers, with full coverage of the condition paths.
- Envtest pinning the CRD schema assumptions the fake client cannot validate: reason-only conditions are legal, all reason constants satisfy the `reason` pattern, the 32768 message cap is real, and `listType=map` rejects duplicate condition types.
- Integration test


api/v1/aerospikecluster_types.go
```diff
@@ -49,6 +49,94 @@ const (
...
+	AerospikeClusterReasonNotRollingRestart = "NotRollingRestart"
+	// AerospikeClusterReasonK8sNodeBlockListEviction distinguishes a restart driven by
+	// spec.k8sNodeBlockList from one driven by a config change.
+	AerospikeClusterReasonK8sNodeBlockListEviction = "K8sNodeBlockListEviction"
+
+	// RackRevisionRollingOut reasons
+	AerospikeClusterReasonRackRevisionRollingOut    = "RackRevisionRollingOut"
+	AerospikeClusterReasonNotRackRevisionRollingOut = "NotRackRevisionRollingOut"
+
+	// Ready failure reasons naming the reconcile stage that failed. These replace the
```
tanmayja: can we re-write this comment. It is confusing to me.

api/v1/aerospikecluster_types.go
```diff
@@ -49,6 +49,94 @@ const (
...
+	// generic ReconcileFailed so consumers can tell a rack problem from an access-control
+	// or roster problem without parsing the condition message.
+	// When a rack operation is interrupted, the corresponding operation condition is left
+	// True, so RackReconcileFailed pairs with e.g. Upgrading=True to identify the operation.
+	AerospikeClusterReasonRackReconcileFailed     = "RackReconcileFailed"
+	AerospikeClusterReasonACLReconcileFailed      = "AccessControlReconcileFailed"
+	AerospikeClusterReasonPSBReconcileFailed      = "PodDisruptionBudgetReconcileFailed"
```
tanmayja: nit: PDB

api/v1/aerospikecluster_types.go
```diff
@@ -49,6 +49,94 @@ const (
...
+	AerospikeClusterReasonServiceReconcileFailed  = "ServiceReconcileFailed"
+	AerospikeClusterReasonClusterConnectionFailed = "ClusterConnectionFailed"
+	AerospikeClusterReasonRosterSetFailed         = "RosterSetFailed"
+	AerospikeClusterReasonMFDSetFailed            = "MigrateFillDelayFailed"
```
tanmayja: if we are saying RosterSetFailed with "set" in between, should we say MigrateFillDelaySetFailed instead of MigrateFillDelayFailed

api/v1/aerospikecluster_types.go
```diff
@@ -1344,6 +1437,8 @@ type AerospikePodStatus struct { //nolint:govet // for readability
 // +kubebuilder:printcolumn:name="HostNetwork",type=boolean,JSONPath=`.spec.podSpec.hostNetwork`
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
```
tanmayja, line 1448: In what cases, phase completed and condition ready differ

internal/controller/cluster/rack.go
```diff
@@ -776,6 +794,23 @@ func (r *SingleClusterReconciler) scaleUpRack(
 	oldSz := *found.Spec.Replicas
 
 	r.Log.Info("Scaling up Pods", "currentSz", oldSz, "desiredSz", desiredSize)
+
+	// Suppressed for a rack under revision migration: there the scale-up is one step of a
+	// migration that also scales the old revision down, so claiming here would toggle ScalingUp
+	// and ScalingDown once per batch while the rack's size never changes. The migration reports
+	// itself as RackRevisionRollingOut for its whole duration. The check is per rack, so a genuine
+	// resize on a different rack in the same pass still reports normally.
+	if !r.revisionChangedRackIDs.Has(rackState.Rack.ID) {
```
tanmayja: revisionChange with genuine scale-up and with genuine scale-down on same rack.

What condition are we expecting to be true?

internal/controller/cluster/rack.go
```diff
@@ -874,6 +905,15 @@ func (r *SingleClusterReconciler) upgradeRack(
 		podList []*corev1.Pod
 	)
 
+	if err = r.setConditions(ctx, metav1.Condition{
```
tanmayja, line 949: just for it to be consistent with rolling restart, can we pull it out to the called under this conditions?
if upgradeNeeded {

internal/controller/cluster/rack.go
```diff
@@ -1425,6 +1481,21 @@ func (r *SingleClusterReconciler) handleK8sNodeBlockListPods(
 			"rollingUpdateBatchSize", r.aeroCluster.Spec.RackConfig.RollingUpdateBatchSize,
 		)
 
+		// K8sNodeBlockList is considered as rolling restart as pods with are coming back up with same name
```
tanmayja: nit: which*

internal/controller/cluster/reconciler.go
```diff
@@ -41,75 +43,87 @@ type SingleClusterReconciler struct {
 	KubeClient  *kubernetes.Clientset
 	KubeConfig  *rest.Config
 	Scheme      *k8sRuntime.Scheme
+	// pendingOpReset holds the operation conditions finishReconcile is allowed to clear on
```
tanmayja: I would divide the fields in 2 category:
**Infrastructure / durable config (initialized once, survives forever)**: Client, Recorder, Scheme, KubeConfig, KubeClient, Log → stable across reconcile invocations (the outer controller reuses the struct).

**Per-pass computed state (computed mid-reconcile, only valid for this call)**: revisionChangedRackIDs, pendingOpReset, stageReason.

As discussed offline, we may think of defining a small struct for values computed during a single pass

internal/controller/cluster/reconciler.go
```diff
@@ -165,6 +189,8 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
 			utils.GetNamespacedNameString(r.aeroCluster),
 		)
 
+		r.stageReason = asdbv1.AerospikeClusterReasonPSBReconcileFailed
```
tanmayja: nit: PDB

internal/controller/cluster/reconciler.go
```diff
@@ -175,25 +201,33 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
...
 		)
 
+		r.stageReason = asdbv1.AerospikeClusterReasonServiceReconcileFailed
+
 		return reconcile.Result{}, fmt.Errorf("reconcile LoadBalancer Service: %w", err)
 	}
 
 	ignorablePodNames, err := r.getIgnorablePods(ctx, nil, getConfiguredRackStateList(r.aeroCluster))
 	if err != nil {
+		r.stageReason = asdbv1.AerospikeClusterReasonClusterConnectionFailed
```
tanmayja: why connection failed here?

internal/controller/cluster/reconciler.go
```diff
@@ -175,25 +201,33 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
...
+
 		return reconcile.Result{}, fmt.Errorf("determine ignorable Pods: %w", err)
 	}
 
 	// Check if there is any node with quiesce status. We need to undo that
 	// It may have been left from previous steps
 	allHostConns, err := r.newAllHostConnWithOption(ctx, ignorablePodNames)
 	if err != nil {
+		r.stageReason = asdbv1.AerospikeClusterReasonClusterConnectionFailed
```
tanmayja: As a user, what should I deduce from AerospikeClusterReasonClusterConnectionFailed reason

- If this denotes connection with aerospike nodes, then newAllHostConnWithOption does not really make any connection, error here might be api failure
- if it denotes k8s api connection failures, then InfoQuiesceUndo can actually have aerospike connection failure plus a genuine command failure.

internal/controller/cluster/reconciler.go
```diff
@@ -175,25 +201,33 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
...
+
 		return reconcile.Result{}, fmt.Errorf("get host connections for cluster nodes: %w", err)
 	}
 
 	if err = deployment.InfoQuiesceUndo(
 		r.Log,
 		r.getClientPolicy(ctx), allHostConns,
 	); err != nil {
+		r.stageReason = asdbv1.AerospikeClusterReasonClusterConnectionFailed
```
tanmayja: for rostering we have separate reason but not for quiescing.

internal/controller/cluster/rack.go
```diff
@@ -841,11 +876,7 @@ func (r *SingleClusterReconciler) scaleUpRack(
 
 		return r.Update(ctx, current, common.UpdateOption)
 	}); err != nil {
-		return found, common.ReconcileError(
-			fmt.Errorf(
-				"scale up StatefulSet %s: %w", utils.GetNamespacedNameString(found), err,
-			),
-		)
+		return found, common.ReconcileError(err)
```
tanmayja, line 920: why?

internal/controller/cluster/reconciler.go
```diff
@@ -155,6 +177,8 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
 			)
 		}
 
+		r.stageReason = asdbv1.AerospikeClusterReasonRackReconcileFailed
```
tanmayja: let's keep it under the res.Err != nil, check other places as well

internal/controller/cluster/reconciler.go
```diff
@@ -464,26 +714,139 @@ func (r *SingleClusterReconciler) updateStatus(ctx context.Context) error {
...
+	ctx context.Context, phase *asdbv1.AerospikeClusterPhase, conditions ...metav1.Condition,
+) error {
+	candidate := slices.Clone(r.aeroCluster.Status.Conditions)
+	condChanged := false
+
+	// Never return early from this loop: every True condition must be claimed even when it
+	// needs no patch.
+	for i := range conditions {
+		if conditions[i].Status == metav1.ConditionTrue {
+			r.pendingOpReset.Delete(conditions[i].Type)
```
tanmayja: during cluster deletion, pendingOpReset is not set by reconciler yet. Let's keep in under nil guard even though it does not panic in go

internal/controller/cluster/rack.go
```diff
@@ -1425,6 +1481,21 @@ func (r *SingleClusterReconciler) handleK8sNodeBlockListPods(
...
 
+		// K8sNodeBlockList is considered as rolling restart as pods with are coming back up with same name
+		if len(podsBatch) > 0 {
+			if err := r.setConditions(ctx, metav1.Condition{
+				Type:   string(asdbv1.AerospikeClusterConditionRollingRestart),
+				Status: metav1.ConditionTrue,
+				Reason: asdbv1.AerospikeClusterReasonK8sNodeBlockListEviction,
+				Message: fmt.Sprintf(
+					"Restarting Pods %s of rack %d to move them off blocked Kubernetes nodes",
+					strings.Join(getPodNames(podsBatch), ", "), rackState.Rack.ID,
```
tanmayja, line 1537: should we truncateConditionMessage this as well, as pod names may lead to a problem

internal/controller/cluster/rack.go
```diff
@@ -1425,6 +1481,21 @@ func (r *SingleClusterReconciler) handleK8sNodeBlockListPods(
 			"rollingUpdateBatchSize", r.aeroCluster.Spec.RackConfig.RollingUpdateBatchSize,
 		)
 
+		// K8sNodeBlockList is considered as rolling restart as pods with are coming back up with same name
+		if len(podsBatch) > 0 {
```
tanmayja, line 1530: This condition cannot go otherwise.

internal/controller/cluster/reconciler.go
```diff
@@ -41,75 +43,87 @@ type SingleClusterReconciler struct {
...
+			})
+	}
+
+	// From here on this pass owns the operation conditions: finishReconcile clears whichever
+	// ones no rack function claims, per the policy set after the rack loop.
+	r.initPendingOpConditionReset()
+
+	// Mark Ready=False at the start of every reconcile so it doesn't stay True
+	// while operations are in progress.
+	if err := r.setConditions(
```
tanmayja: This is more cleaner, but there is 1 optimization possible  at the start of every reconcile, to do only single patch by combining set conditions and set status phase

internal/controller/cluster/reconciler.go
```diff
@@ -464,26 +714,139 @@ func (r *SingleClusterReconciler) updateStatus(ctx context.Context) error {
...
+	if phaseChanged {
 		r.addClusterPhaseMetric()
 	}
 
 	return nil
 }
 
+func (r *SingleClusterReconciler) setStatusPhase(ctx context.Context, phase asdbv1.AerospikeClusterPhase) error {
+	if err := r.mergePatchStatus(ctx, &phase); err != nil {
+		r.Log.Error(err, fmt.Sprintf("Failed to set cluster status to %s", phase))
```
tanmayja: should not have fmt.Sprintf in log messages

tanmayja: do we really need to log the error here as we are immediately returning err

api/v1/aerospikecluster_types.go
abhishekdwivedi3060: Mostly they are same. They would differ during deletion (Ready=False/Terminating while phase
  stays Completed) and on first reconcile (Ready=Unknown/Initializing before any phase). Its just a way to show Ready condition state in kubectl get output.

internal/controller/cluster/rack.go
abhishekdwivedi3060: As discussed, any rack operation combined with revision change would be considered part of RackRevisionRollingOut operation. This also aligns with our future scale up and scale down in one go flow

internal/controller/cluster/rack.go
abhishekdwivedi3060: Intentionally kept it inside to keep in sync with scale up and scale down.
RollingRestart is an exception because it leveraged 2 different paths (rolling restart and failed pod handling), hence condition is present in the caller

internal/controller/cluster/rack.go
abhishekdwivedi3060: The main caller func already logs similar error.

internal/controller/cluster/rack.go
```diff
@@ -430,6 +434,19 @@ func (r *SingleClusterReconciler) upgradeOrRollingRestartRack(
...
 		if rollingRestartInfo.needRestart {
+			// The exception to the rule that an operation function claims its own condition, as
+			// scaleUpRack, scaleDownRack, and upgradeRack all do. rollingRestartRack cannot,
+			// because it is also used as the recovery mechanism handleFailedPodsInRack to
+			// force-restart a stuck pod. handleFailedPodsInRack should not be able to set this condition.
+			if err := r.setConditions(ctx, metav1.Condition{
+				Type:    string(asdbv1.AerospikeClusterConditionRollingRestart),
+				Status:  metav1.ConditionTrue,
+				Reason:  asdbv1.AerospikeClusterReasonRollingRestart,
+				Message: fmt.Sprintf("Rolling restart of rack %d", rackState.Rack.ID),
```
tanmayja: nit: other places we don't have "of". like : Scaling         down  rack

internal/controller/cluster/rack.go
```diff
@@ -1425,6 +1479,22 @@ func (r *SingleClusterReconciler) handleK8sNodeBlockListPods(
 			"rollingUpdateBatchSize", r.aeroCluster.Spec.RackConfig.RollingUpdateBatchSize,
 		)
 
+		// K8sNodeBlockList is reported as a rolling restart because the Pods come back up with the
```
tanmayja, line 1526: This will only be set for ignorable pods

internal/controller/cluster/reconciler.go
```diff
@@ -271,6 +321,126 @@ func (r *SingleClusterReconciler) Reconcile(ctx context.Context) (result ctrl.Re
...
+	}
+
+	return common.ReconcileSuccess()
+}
+
+// finishReconcile runs at end of Reconcile; return value is assigned to Reconcile's named recErr in defer.
+func (r *SingleClusterReconciler) finishReconcile(ctx context.Context, result ctrl.Result, recErr error) error {
+	logValues := common.ReconcileExitLogValues(result, recErr)
+
+	statusErr := r.writeTerminalStatus(ctx, recErr)
```
shubhamyelikar99, line 425: finishReconcile now calls writeTerminalStatus unconditionally on every exit, including the deletion path. On a successful delete, within a single reconcile pass:


```
Reconcile
├─ defer finishReconcile(...)                            reconciler.go:85-88
└─ handleTerminatingCluster(ctx)                         reconciler.go:93
   ├─ handleClusterDeletion → cleanUpAndRemoveFinalizer  (succeeds)
   ├─ removeClusterPhaseMetric()                         reconciler.go:350   gauge dropped
   └─ return nil
   ↓ named return recErr = nil, defer fires
   finishReconcile(ctx, ctrl.Result{}, nil)              reconciler.go:417
   └─ writeTerminalStatus(ctx, nil)                      reconciler.go:368
      └─ mergePatchStatus(ctx, nil)   // no conditions, no phase
         └─ !condChanged && !phaseChanged → addClusterPhaseMetric()   reconciler.go:844   gauge back
```
pendingOpReset is nil on this path (deletion returns long before initPendingOpConditionReset), so opConditionsToClear() returns an empty slice and mergePatchStatus falls straight into the no-op branch that re-registers the gauge.

Every deleted cluster leaves aerospike_ako_aerospikecluster_phase{...,phase="Completed"} 1 behind forever. In master finishReconcile did nothing on the success path (the setStatusPhase(Error) call was guarded by recErr != nil), so the deletion path never reached a status write.

internal/controller/cluster/reconciler.go
shubhamyelikar99, line 1: an operation condition can stay True after its operation has finished.

The error path in writeTerminalStatus writes only Ready + phase=Error and never calls opConditionsToClear(), so an erroring pass clears nothing.

Concretely: an upgrade completes inside the rack loop, but the same pass then fails downstream (bad aerospikeAccessControl, roster set, migrate-fill-delay — anything after reconcileRacks). Upgrading=True is kept. On the next pass isRackUpgradeNeeded is false, so upgradeRack never runs and Upgrading is correctly left sitting in pendingOpReset — and then the error path discards that set again. Repeat: Upgrading=True persists for as long as the downstream stage keeps failing, for an upgrade that finished passes ago.

internal/controller/cluster/reconciler.go
abhishekdwivedi3060: Good catch 👍.
Fixed and added unit test-case for this

internal/controller/cluster/reconciler.go
abhishekdwivedi3060: Yeah, you currently pointed out the behaviour but I think we intentionally made this design choice for now.
The high level objective was to keep the last state as it is in case of failure, so an upgrade keep saying Upgrading=True for the whole duration until the issue is fixed. That same rule can cause Upgrading=True even for the last reconcile when it reaches the downstream reconcile flow. This behaviour is self correcting once the issue is resolved.
Lets park it for now and revisit this based on feedback.

shubhamyelikar99 (approved):

tanmayja (approved):

PR #580, opened 2026-08-11
Feat: optimize migrations during rolling restarts using MFD tuning

tanmayja:


api/v1/aerospikecluster_types.go
```diff
@@ -1154,40 +1186,22 @@ type AerospikeClusterStatusSpec struct { //nolint:govet // for readability
...
-	// +patchStrategy=strategic
-	// +optional
-	Pods map[string]AerospikePodStatus `json:"pods" patchStrategy:"strategic"`
-
-	// Phase denotes the current phase of Aerospike cluster operation.
-	// +optional
-	Phase AerospikeClusterPhase `json:"phase,omitempty"`
-
-	// Selector specifies the label selector for the Aerospike pods.
-	// +optional
```
shubhamyelikar99, line 1189: why are we removing field comments , kubebuilder markers etc?

api/v1/utils.go
```diff
@@ -546,7 +546,14 @@ func GetIntType(value interface{}) (int, error) {
...
+	if asConfig == nil {
+		return 0, nil
+	}
+
+	serviceConfig, ok := asConfig.Value[ConfKeyService].(map[string]interface{})
+	if !ok {
+		return 0, nil
+	}
 
 	fillDelayIFace, exists := serviceConfig["migrate-fill-delay"]
```
shubhamyelikar99, line 558: should we make this "migrate-fill-delay" const? i can see its used in multiple places

internal/controller/cluster/pod.go
```diff
@@ -699,6 +669,32 @@ func getServerFailedAndActivePods(
...
+}
+
+// hasDefinitiveFailure reports whether at least one pod in the list has a failure that is not
+// merely Unschedulable. Unschedulable pods are transient — the user may have set a large
+// OverrideMigrateFillDelay to tolerate a node being temporarily unavailable. Definitive failures
+// (PodFailed phase, CrashLoopBackOff, ImagePullBackOff, etc.) indicate the pod cannot start and
+// the MFD override should be reverted.
+func hasDefinitiveFailure(pods []*corev1.Pod) bool {
+	for idx := range pods {
+		isUnschedulable, _ := utils.IsPodReasonUnschedulable(pods[idx])
```
shubhamyelikar99, line 689: instead of ignoring error can we log it in debug level? we might need this during debugging

internal/controller/cluster/pod.go
```diff
@@ -2046,49 +2018,83 @@ func (r *SingleClusterReconciler) getEvictionBlockedPods(ctx context.Context) (s
...
+			if restartTypeMap[podsToRestart[idx].Name] == podRestart {
 				podRestartNeeded = true
 				break
 			}
 		}
 	}
 
 	if !podRestartNeeded {
-		return false
+		return 0, false, nil
```
shubhamyelikar99, line 2050: `drainBeforeStability` comes from the restart types in the batch about to run
(pod.go:2039-2050). But MFD being raised was done by a **previous** batch in a
**previous** reconcile — `rollingRestartRack` requeues while batches remain
(rack.go:1402), so the revert at reconciler.go:267 never runs in between.

A rack can hold mixed restart types, and `chunkBy` slices `podsToRestart` in pod order
regardless of type — so a warm-only batch can follow a `podRestart` batch.

**Example** — 2 pods, batch size 1, `overrideMigrateFillDelay: 300`, config MFD 30.
p1 needs `podRestart` (e.g. `spec.operations`), p0 only `quickRestart` (config change):

R1  batch [p1] → (300, true)   MFD 0 → 300, delete p1, requeue
    ► MFD left at 300

R2  batch [p0] → (0, false)    drain SKIPPED
                               waitForClusterStability runs at MFD = 300

**Impact** — Since p1 rejoins and cancels its own pending fills. But MFD
stays raised for every remaining warm batch (holding fills 300s while nothing is being
deleted), and if a fill *is* pending, stability requeues every 60s until the override
elapses.

api/v1/aerospikecluster_types.go
tanmayja: merge issue, fixed

internal/controller/cluster/pod.go
tanmayja: nice observation. Idea is, MFD should be set to configMFD in cases where there is no cold restart.

api/v1/aerospikecluster_types.go
```diff
@@ -309,6 +309,38 @@ type AerospikeClusterSpec struct { //nolint:govet // for readability
...
+	// +optional
+	RestartStrategy *RestartStrategy `json:"restartStrategy,omitempty"`
+}
+
+// RestartStrategy configures transient operator behaviour applied around a rolling pod restart or
+// upgrade.
+type RestartStrategy struct {
+	// OverrideMigrateFillDelay is the duration in seconds that AKO temporarily sets as the
+	// migrate-fill-delay on the Aerospike cluster before restarting pods. This delays migration
+	// fills while a pod is down, giving the cluster time to avoid unnecessary data movement
```
sidntrivedi, line 325: nit: delays fill migrations

api/v1/utils.go
```diff
@@ -546,9 +547,16 @@ func GetIntType(value interface{}) (int, error) {
 
 // GetMigrateFillDelay returns the migrate-fill-delay from the Aerospike configuration
 func GetMigrateFillDelay(asConfig *AerospikeConfigSpec) (int, error) {
-	serviceConfig := asConfig.Value[ConfKeyService].(map[string]interface{})
+	if asConfig == nil {
+		return 0, nil
```
sidntrivedi, line 551: We are returning 0 here because its the default value in the server?

internal/controller/cluster/aero_info_calls.go
```diff
@@ -340,12 +375,26 @@ func (r *SingleClusterReconciler) setMigrateFillDelay(
...
 
-	if err := deployment.SetMigrateFillDelay(r.Log, policy, allHostConns, migrateFillDelay); err != nil {
+	if err := deployment.SetMigrateFillDelay(r.Log, policy, allHostConns, delay); err != nil {
 		return common.ReconcileError(err)
 	}
 
+	// Persist DynamicMigrateFillDelay immediately so the value survives a mid-reconcile requeue.
+	// Without this, an in-memory-only update would be lost when the next reconcile reads from k8s.
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		if err := r.Get(ctx, utils.GetNamespacedName(r.aeroCluster), r.aeroCluster); err != nil {
```
sidntrivedi, line 387: In case the user updates the CR mid-reconcile, it would get the latest updated CR from the API server. So, instead of updating the whole the spec from API server to the `r.aeroCluster` object which is being used across reconcile cycle, should we use a separate variable? Something like:

```go
if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
    current := &asdbv1.AerospikeCluster{}
    if err := r.Get(ctx, utils.GetNamespacedName(r.aeroCluster), current); err != nil {
        return err
    }

    current.Status.DynamicMigrateFillDelay = int64(delay)

    if err := r.Client.Status().Update(ctx, current); err != nil {
        return err
    }

    r.aeroCluster.Status.DynamicMigrateFillDelay = current.Status.DynamicMigrateFillDelay
    r.aeroCluster.ResourceVersion = current.ResourceVersion

    return nil
}); err != nil {
```

internal/controller/cluster/aero_info_calls.go
```diff
@@ -301,35 +347,24 @@ func hostID(hostName string, hostPort int) string {
...
-		if err != nil {
-			return common.ReconcileError(err)
-		}
-	}
-
-	if migrateFillDelay == 0 && oldMigrateFillDelay == 0 {
-		r.Log.Info("migrate-fill-delay config not present or 0, skipping it")
+	if int64(delay) == r.aeroCluster.Status.DynamicMigrateFillDelay {
+		r.Log.Info("migrate-fill-delay already at desired value, skipping", "value", delay)
 		return common.ReconcileSuccess()
```
sidntrivedi, line 365: Should we ensure that the mfd value has been set on all nodes to return the reconcile success status?

internal/controller/cluster/pod.go
```diff
@@ -684,10 +684,13 @@ func podNamesToSet(pods []*corev1.Pod) sets.Set[string] {
...
 // the MFD override should be reverted.
-func hasDefinitiveFailure(pods []*corev1.Pod) bool {
+func (r *SingleClusterReconciler) hasDefinitiveFailure(pods []*corev1.Pod) bool {
 	for idx := range pods {
-		isUnschedulable, _ := utils.IsPodReasonUnschedulable(pods[idx])
-		if !isUnschedulable {
+		isUnschedulable, reason := utils.IsPodReasonUnschedulable(pods[idx])
+		if isUnschedulable {
+			r.Log.V(1).Info("Pod is unschedulable, skipping MFD revert",
+				"pod", pods[idx].Name, "reason", reason)
```
shubhamyelikar99, line 692: utils.GetNamespacedName(pods[idx])

shubhamyelikar99 (dismissed):

api/v1/aerospikecluster_types.go
```diff
@@ -309,6 +309,38 @@ type AerospikeClusterSpec struct { //nolint:govet // for readability
...
+	// during short maintenance windows. Once the pod restarts and rejoins the cluster, AKO
+	// resets migrate-fill-delay to 0 so that rebalancing can proceed immediately.
+	// This field only takes effect when a pod restart (not a warm restart) is required.
+	// This settings are applied dynamically before the first pod is taken down and
+	// reverted once all pods have rejoined; they are never written to aerospike.conf.
```
abhishekdwivedi3060, line 330: nit:
```suggestion
	// OverrideMigrateFillDelay is the duration in seconds that AKO temporarily sets as the
	// migrate-fill-delay on the Aerospike cluster before restarting pods. This delays fill migration
	// while a pod is down, giving the cluster time to avoid unnecessary data movement
	// during short maintenance windows. Once the pod restarts and rejoins the cluster, AKO
	// resets migrate-fill-delay to 0 so that rebalancing can proceed immediately.
	// This field only takes effect when a pod restart (not a warm restart) is required.
	// This setting is applied dynamically before the first pod is taken down and
	// reverted once all pods have rejoined; they are never written to aerospike.conf.
```

internal/controller/cluster/pod.go
```diff
@@ -353,56 +353,26 @@ func (r *SingleClusterReconciler) rollingRestartPods(
...
-			if res := r.setMigrateFillDelay(ctx, clientPolicy, &rackState.Rack.AerospikeConfig, false,
-				ignorablePodNames,
-			); !res.IsSuccess {
-				if res.Err != nil {
-					res.Err = fmt.Errorf(
-						"revert migrate-fill-delay for rack %d before restarting running Pods: %w",
-						rackState.Rack.ID, res.Err,
-					)
-				}
+		r.Log.Info("migrate-fill-delay for Pod restart", "delay", migrateFillDelay, "drain", drainBeforeStability)
```
abhishekdwivedi3060, line 362:
```suggestion
		r.Log.Info("Adjust migrate-fill-delay for Pod restart", "migrateFillDelay", migrateFillDelay, "needed", drainBeforeStability)
```
This "drain" is internal implementation and won't be clear to the user in the log. Use a different var to denote that we are adjusting/tuning the MFD for pod restart. "needed" is just an eg

internal/controller/cluster/pod.go
```diff
@@ -750,50 +749,26 @@ func (r *SingleClusterReconciler) safelyDeletePodsAndEnsureImageUpdated(
...
+		// so mfdDelayForRestart short-circuits immediately without inspecting the pod list.
+		migrateFillDelay, drainBeforeStability, err := r.mfdDelayForRestart(rackState, nil, nil)
+		if err != nil {
+			return common.ReconcileError(err)
 		}
 
-		var clientPolicy *as.ClientPolicy
-
-		setMigrateFillDelay := r.shouldSetMigrateFillDelay(rackState, podsToUpdate, nil)
+		r.Log.Info("migrate-fill-delay for Pod upgrade", "delay", migrateFillDelay, "drain", drainBeforeStability)
```
abhishekdwivedi3060, line 759: same here

internal/controller/cluster/rack.go
```diff
@@ -2596,13 +2561,29 @@ func (r *SingleClusterReconciler) handleFailedPodsInRack(
...
 
+		// Revert MFD only when the user has configured OverrideMigrateFillDelay AND at least one
+		// pod has a definitive failure (crash, image error, or PodFailed phase). Without an
+		// override, MFD was never raised above the config value so there is nothing to revert.
+		// If ALL failed pods are merely Unschedulable, preserve the override — the user may have
+		// set a large value precisely to tolerate a pod being absent while a node is coming up.
+		if r.aeroCluster.Spec.RestartStrategy.GetOverrideMigrateFillDelay() > 0 && r.hasDefinitiveFailure(serverFailedPods) {
+			// Include failed pod names in the ignorable set so that the info call inside
+			// revertMFDToConfig skips their (dead) Aerospike nodes and doesn't fail.
+			ignorableWithFailed := ignorablePodNames.Union(podNamesToSet(serverFailedPods))
```
abhishekdwivedi3060, line 2572: I think directly updating ignorablePodNames can be problematic here.
After this, flow will go in reconcileRack with failed pods and if user scaled down in the previous update, then that failed pod (if scale down candidate) will be ignored and scaled down. In normal flow, this will happen only if maxIgnorable is set

internal/controller/cluster/reconciler.go
```diff
@@ -695,6 +691,13 @@ func (r *SingleClusterReconciler) updateStatus(ctx context.Context) error {
 	newAeroCluster.Status.AerospikeClusterStatusSpec = *specToStatus
 	newAeroCluster.Status.Phase = asdbv1.AerospikeClusterCompleted
 
+	// setMigrateFillDelay persists DynamicMigrateFillDelay immediately to the API server and
+	// re-fetches r.aeroCluster, so this field is always up-to-date before updateStatus runs.
+	// The reconciler safety-net call at line 222 also handles fresh-cluster initialisation:
+	// if configMFD > 0 the guard fails and the info call sets DynamicMigrateFillDelay;
+	// if configMFD == 0 it stays 0, which is already the correct value.
```
abhishekdwivedi3060, line 698: The comment is not clear, please update

test/cluster/restart_migrate_fill_delay_test.go
```diff
@@ -0,0 +1,228 @@
...
+					Expect(err).ToNot(HaveOccurred())
+
+					// After all restarts, the reconciler revert should restore the
+					// aerospike-config-level migrate-fill-delay value.
+					err = validateMigrateFillDelay(ctx, k8sClient, logger, clusterNamespacedName,
+						configMFD, nil, firstPodName)
+					Expect(err).ToNot(HaveOccurred())
+				},
+			)
+
```
abhishekdwivedi3060, line 141: I think these 2 are kind of duplicate and can be clubbed together.
Check 1: MFD should be restartMigrateFillDelay
Check 2 after pod restart: MFD should be 0. See if we can track this in test
Check 3: MFD should be aerospikeConfig value

test/cluster/restart_migrate_fill_delay_test.go
```diff
@@ -0,0 +1,228 @@
...
+					Expect(err).ToNot(HaveOccurred())
+
+					// MFD must still be 0 — RestartMigrateFillDelay must not have fired
+					err = validateMigrateFillDelay(ctx, k8sClient, logger, clusterNamespacedName,
+						0, nil, firstPodName)
+					Expect(err).ToNot(HaveOccurred())
+				},
+			)
+		},
+	)
```
abhishekdwivedi3060, line 167: This test-case waits for updateCluster which waits for end state.
So the assertion validateMigrateFillDelay is useless after that. Not sure how we can track MFD during a warm restart

test/cluster/restart_migrate_fill_delay_test.go
```diff
@@ -0,0 +1,228 @@
...
+						{Kind: asdbv1.OperationPodRestart, ID: "no-mfd-restart-1"},
+					}
+
+					err = updateCluster(k8sClient, ctx, aeroCluster)
+					Expect(err).ToNot(HaveOccurred())
+
+					// migrate-fill-delay must remain at the user-configured value
+					err = validateMigrateFillDelay(ctx, k8sClient, logger, clusterNamespacedName,
+						configMFD, nil, firstPodName)
+					Expect(err).ToNot(HaveOccurred())
```
abhishekdwivedi3060, line 223: Same issue. We are asserting after the end state.

test/cluster/restart_migrate_fill_delay_test.go
abhishekdwivedi3060, line 1: I think we've some tests around deleteLocalStorageOnRestart + Local SSDs, can you check if those need some update or can be leveraged for some of these

abhishekdwivedi3060 (changes requested):

test/cluster/restart_migrate_fill_delay_test.go
```diff
@@ -0,0 +1,228 @@
...
+			It(
+				"Should not set RestartMigrateFillDelay during warm restart", func() {
+					aeroCluster, err := getCluster(k8sClient, ctx, clusterNamespacedName)
+					Expect(err).ToNot(HaveOccurred())
+
+					firstPodName := aeroCluster.Name + "-" +
+						strconv.Itoa(aeroCluster.Spec.RackConfig.Racks[0].ID) + "-0"
+
+					// Trigger a warm restart (ASD-only restart, not pod restart).
+					// shouldSetMigrateFillDelay only fires for podRestart type, so
```
abhishekdwivedi3060, line 151: shouldSetMigrateFillDelay is removed now

internal/controller/cluster/aero_info_calls.go
```diff
@@ -102,6 +138,16 @@ func (r *SingleClusterReconciler) waitForMultipleNodesSafeStopReady(
...
 	}
 
+	// Raise MFD to migrateFillDelay before quiesce. Only applies on the drain path
+	// (drainBeforeStability=true) where MFD was zeroed above and needs to be raised to the
+	// override value to suppress fills while the pod is absent. On the non-drain path MFD was
+	// already set to migrateFillDelay before stability, so this is intentionally skipped.
+	if drainBeforeStability && migrateFillDelay > 0 {
+		if res := r.setMigrateFillDelay(ctx, policy, migrateFillDelay, ignorablePodNames); !res.IsSuccess {
+			return res
+		}
```
abhishekdwivedi3060, line 148: Now we are creating newAllHostConnWithOption multiple times here in the func.
See if we can refactor setMigrateFillDelay to use the allHostConns created above. We can use the fallback like if given then use that else create one

internal/controller/cluster/rack.go
abhishekdwivedi3060, line 1: Stale doc comment at internal/controller/cluster/rack.go:704. It still describes "migrate-fill-delay revert: skipped for server-failed pods" as one of two behaviours isServerFailed controls "inside reconcileRack". This PR deleted that block; the revert now lives in handleFailedPodsInRack (rack.go:2569) and reconciler.go:267

abhishekdwivedi3060 (changes requested):

api/v1/utils.go
tanmayja: yes

internal/controller/cluster/aero_info_calls.go
tanmayja: No, this is a known decision we took to track MFD value what AKO is setting to avoid unnecessary info calls.

abhishekdwivedi3060 (approved):

PR #581, opened 2026-08-13
test: export sidecar and roster helpers from cluster_helper

rpandey-stack:
Move shared cluster test helpers into `cluster_helper.go` and export them
so other integration test packages can reuse sidecar failure and roster
validation utilities in QE repo


rpandey-stack:
After discussion with dev, we are not going ahead with approach of making AKO utility functions as public and re-using it in QE repo. Hence this PR code changes need not be merged in master. Closing the PR.

PR #586, opened 2026-08-24
fix: allow advertise-ipv6 when cluster nodes report IPv6 addresses

### LINKED ISSUE #585: advertise-ipv6 is rejected unconditionally, which blocks single-stack IPv6 Kubernetes clusters
On a single-stack IPv6 EKS cluster, pods only have IPv6 addresses, so Aerospike's mesh heartbeat needs `service.advertise-ipv6: true` to form a multi-node cluster. The validating webhook rejects that setting unconditionally (`advertise-ipv6 is not supported`, added in #336), so multi-node AerospikeCluster deployments are impossible on IPv6-only infrastructure: the pods start, but the one setting that would let them cluster can never be applied.

We (CloudX) run Aerospike on single-stack IPv6 EKS and have validated a narrowing of that validation in production-grade testing: allow `advertise-ipv6` when at least one Kubernetes node reports an IPv6 InternalIP, keep rejecting it everywhere else. With that in place (plus a small address-family fix in aerospike-kubernetes-init, reported separately), a 3-node RF=2 cluster on server 8.1.1.0 Enterprise forms and stays healthy through pod deletion, operator-driven rolling restarts, scale up/down with quiesce, and eviction handling.

An internal feature request for IPv6 support already exists at Aerospike through our vendor channel, so there should be a Jira this can be attached to. We have a PR ready that narrows the validation without changing behaviour on IPv4-only clusters.

nickpell:
#336 added validation that rejects `service.advertise-ipv6` outright. This PR narrows that check rather than reverting it: on a single-stack IPv6 cluster, `advertise-ipv6: true` is the only way Aerospike mesh heartbeats can form a multi-node cluster, so the webhook now rejects the setting only where it genuinely cannot work — when no Kubernetes node reports an IPv6 `InternalIP`.

**No behaviour change on IPv4-only clusters** — `advertise-ipv6` is still rejected there, with a message that now says why.

- The webhook probes nodes for an IPv6 `InternalIP`, caches the answer for five minutes, and pages with an early exit so admission never lists a large cluster in full. Reads go through `mgr.GetAPIReader()` rather than the manager cache — the pattern `internal/webhook/eviction/eviction_webhook.go` established in #413 — so no cluster-wide Node informer or watch permission is added.
- If the probe fails, the webhook fails closed: only `advertise-ipv6: true` is rejected (with the probe error in the message), never unrelated admissions.
- Reading nodes needs `get`/`list` on core `nodes`, added via a `+kubebuilder:rbac` marker. `config/rbac/role.yaml` and the chart clusterrole are updated together, as in #413/#340; no `Chart.yaml` bump.
- The unchecked `val.(bool)` assertion is replaced with the two-value form: the JSON schema rejects non-booleans first in webhook flow, but exported `ValidateAerospikeConfig` must not panic on bad input.

## Testing

- Unit tests (`pkg/validation`) cover the guard: enabled/disabled/absent × capable/incapable, both rejection messages (no IPv6 nodes vs. probe failure), and non-boolean values.
- Envtests (`test/envtests/cluster`) drive the real webhook against seeded nodes: rejection with only an IPv4 node present, acceptance with an IPv6 node, and schema-error precedence for non-boolean values. Suites pass under `ENVTEST_K8S_VERSION=1.23` and `1.35`.
- IPv4 control — admission message before/after on a v4-only cluster:

      # upstream: advertise-ipv6 is not supported
      # patched:  aerospikeConfig.service.advertise-ipv6 requires an IPv6-capable Kubernetes
      #           cluster, but no Kubernetes node reports an IPv6 InternalIP address

- IPv6 validation: healthy 3-node RF=2 formation on a single-stack IPv6 EKS cluster (Aerospike server 8.1.1.0 Enterprise, real feature key), surviving single-pod deletion, operator-driven rolling restarts, scale 3→4→3 with quiesce, and eviction-API handling. Full lifecycle logs available on request.

Fixes #585

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -22,6 +22,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ""
+  resources:
   - pods
   verbs:
   - delete
--- a/helm-charts/aerospike-kubernetes-operator/templates/aerospike-operator-manager-clusterrole.yaml
+++ b/helm-charts/aerospike-kubernetes-operator/templates/aerospike-operator-manager-clusterrole.yaml
@@ -27,6 +27,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ""
+  resources:
   - pods
   verbs:
   - delete
--- a/internal/webhook/v1/aerospikecluster_validating_webhook.go
+++ b/internal/webhook/v1/aerospikecluster_validating_webhook.go
@@ -47,6 +47,30 @@ import (
 
 // +kubebuilder:object:generate=false
 type AerospikeClusterCustomValidator struct {
+	ipv6Prober *nodeIPv6Prober
+}
+
+// ipv6Capable reports whether the cluster has IPv6-capable nodes, which gates the
+// advertise-ipv6 Aerospike setting. A non-nil probe error is the reason capability
+// could not be determined; the cluster is then reported as not IPv6-capable.
+//
+// The value is only consulted when advertise-ipv6 is requested, so an unprobeable
+// cluster keeps rejecting that one setting instead of failing every admission.
+func (acv *AerospikeClusterCustomValidator) ipv6Capable(
+	ctx context.Context, aslog logr.Logger,
+) (bool, error) {
+	if acv.ipv6Prober == nil {
+		return false, nil
+	}
+
+	capable, err := acv.ipv6Prober.IPv6Capable(ctx)
+	if err != nil {
+		aslog.Error(err, "Failed to determine node IPv6 capability, treating cluster as not IPv6-capable")
+
+		return false, err
+	}
+
+	return capable, nil
 }
 
 //nolint:lll // for readability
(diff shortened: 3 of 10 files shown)


abhishekdwivedi3060 (commented):
Thank you for workong on this.  Really appreciate it

After going though the PR, I've a few concerns. 
1. I think this change lands in the wrong layer. The change is both larger than it needs to be and gates the wrong condition. 
2.  The IPv6 probe added in the webhook seems like an overkill with elevated permission for the Node resource. It may raise some flags for some of our existing customers.

`advertise-ipv6` in the server is an **allow-switch for IPv6, not a mode switch**. It only ever *widens* what is published; IPv4 is never filtered.

**Ask:**  reduce this PR to deleting the blanket rejection. Drop `ipv6.go`, the node RBAC, the`WebhookOption` plumbing and the parameter threading through the exported validation functions.The real guard belongs in `akoinit` (below).
---

## Where we're taking IPv6 support (High level idea, actual implementation details are yet to be explored)

**One rule**, applied in `akoinit` on the advertise side.

```
IP candidates = every address available for this network type (all families)
emit       = advertise-ipv6 ? candidates : candidates.filter(isIPv4)
```

AKO applies the filter itself rather than letting the server do it so that
`status...accessEndpoints` reflects what clients can actually reach — that field is built from the set AKO emits, not from what the server publishes — and so that `akoinit` can detect the unsupportable case up front instead of it surfacing as a cluster that never forms.

**Phase 1 — no CRD change.** Delivers single-stack IPv6 clusters, plus dual-stack clusters
advertising *both* families when the flag is on.

- delete the blanket rejection (this PR, reduced)
- `akoinit`: collect node/pod addresses per family instead of picking one; apply the filter once, uniformly, across `pod` / `host*` / `configuredIP` / `customInterface`
- `akoinit`: fail fast when the emitted set is empty — mandatory, since the unsubstituted `<access-address>` placeholder otherwise reaches `addrs_to_access` and crashes the server
- `akoinit`: `hostNetwork` heartbeat/fabric literal injection must emit all allowed families
- AKO: reconcile `ipFamilyPolicy: PreferDualStack` on the headless, per-pod NodePort and LB services when the flag is on — a no-op on single-stack clusters, and `ipFamilyPolicy` is mutable (`ipFamilies` accepts a secondary family; only the *primary* family is fixed), so existing clusters upgrade in place with no Service recreation
- AKO: treat `advertise-ipv6` as non-dynamic and force a rolling quick restart when it changes (the CR only holds the `<access-address>` placeholder, so the diff looks dynamic while the effect is not)

**Phase 2 — one optional field.** `aerospikeNetworkPolicy.ipFamily`, whose only job is to *suppress* a family: empty (default) mirrors `advertise-ipv6`; `IPv6` advertises IPv6 only on a dual-stack cluster. No default value, ever, so Phase 1 clusters keep working.

**Migration**, IPv4 → IPv6 on dual-stack:  Future scope

PR #589, opened 2026-08-27
bugfix: skipping unnecessary volume mounts from aerospike init container. [KO-618]

tanmayja:
Summary
Problem: The aerospike-init container was unconditionally mounting every storage volume defined in the AerospikeCluster CR — including non-PV volumes (ConfigMap, Secret, EmptyDir, HostPath) that are only attached to sidecars and have no aerospike path. The init container has no code that touches these volumes (initVolumes, wipeVolumes, cleanDirtyVolumes all call getPersistentVolumes() first, filtering to PV-only). These mounts were dead weight.

Fix: getFinalVolumeAttachmentsForVolume in statefulset.go now auto-mounts a volume in the init container only when the init container actually needs it:

PersistentVolumes — must be mounted for wipe/initialization before ASD starts.
Non-PV volumes with an aerospike attachment — must be mounted so the init container can set up the workdir (smd, usr/udf/lua).
Non-PV volumes without an aerospike attachment (e.g. ConfigMap/Secret used only by sidecars) — skipped. 

Backward Compatibility & Upgrade Impact:

1. No rolling restart triggered on operator upgrade. The STS uses OnDeleteStatefulSetStrategyType, so Kubernetes never auto-restarts pods on template changes. The operator's own restart detection (isRackStorageUpdatedInAeroCluster) explicitly skips the aerospike-init container, and podSpecHash covers only user-facing CR spec fields — neither fires from this change.
2. Pods pick up the slimmer init container template only on their next natural recreation (image upgrade, config change, etc.).

### DIFF
--- a/internal/controller/cluster/rack.go
+++ b/internal/controller/cluster/rack.go
@@ -1579,19 +1579,18 @@ func (r *SingleClusterReconciler) isRackStorageUpdatedInAeroCluster(
 	}
 
 	// Check for removed volumeAttachments
-	allConfiguredInitContainers := []string{
-		asdbv1.
-			AerospikeInitContainerName,
-	}
 	allConfiguredContainers := []string{asdbv1.AerospikeServerContainerName}
 
 	for idx := range r.aeroCluster.Spec.PodSpec.Sidecars {
 		allConfiguredContainers = append(allConfiguredContainers, r.aeroCluster.Spec.PodSpec.Sidecars[idx].Name)
 	}
 
-	// Include InitContainers (excluding placeholder)
+	// Collect user-defined init containers. The aerospike-init container is intentionally
+	// excluded because isVolumeAttachmentRemoved skips it unconditionally — its volume
+	// mounts are not individually tracked in storage spec.
+	var allConfiguredInitContainers []string
+
 	for idx := range r.aeroCluster.Spec.PodSpec.InitContainers {
-		// Skip placeholder (aerospike-init) as it's not a real container in the spec
 		if r.aeroCluster.Spec.PodSpec.InitContainers[idx].Name != asdbv1.AerospikeInitContainerName {
 			allConfiguredInitContainers = append(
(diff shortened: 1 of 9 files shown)


internal/controller/cluster/rack.go
```diff
@@ -1579,19 +1579,18 @@ func (r *SingleClusterReconciler) isRackStorageUpdatedInAeroCluster(
...
-	}
 	allConfiguredContainers := []string{asdbv1.AerospikeServerContainerName}
 
 	for idx := range r.aeroCluster.Spec.PodSpec.Sidecars {
 		allConfiguredContainers = append(allConfiguredContainers, r.aeroCluster.Spec.PodSpec.Sidecars[idx].Name)
 	}
 
-	// Include InitContainers (excluding placeholder)
+	// Collect user-defined init containers. The aerospike-init container is intentionally
+	// excluded because isVolumeAttachmentRemoved skips it unconditionally — its volume
```
abhishekdwivedi3060: nit: Fix comment. Now isVolumeAttachmentRemoved doesn't have check for skipping the init container unconditionally

test/cluster/statefulset_storage_test.go
```diff
@@ -385,6 +398,63 @@ var _ = Describe(
...
+							Name:    "sidecar-consumer",
+							Image:   "busybox:1.28",
+							Command: []string{"sh", "-c", "sleep 3600"},
+						}
+						aeroCluster.Spec.PodSpec.Sidecars = append(
+							aeroCluster.Spec.PodSpec.Sidecars, sidecar,
+						)
+
+						// EmptyDir volume mounted only in the sidecar — no Aerospike attachment.
+						sidecarOnlyVol := asdbv1.VolumeSpec{
```
abhishekdwivedi3060, line 420: Can we also add a hostPath volume just to confirm the bug fix?

abhishekdwivedi3060 (dismissed):

abhishekdwivedi3060 (approved):

PR #594, opened 2026-09-04
chore: bump go version to 1.26 and k8s to 1.36.4 [KO-625]

sidntrivedi:
Addresses https://aerospike.atlassian.net/browse/KO-625.
- Bumps Go version to v1.26.
- Bumps golangci-lint version to v2.13.2
- Bumps K8s deps to be compatible with K8s 1.36.4.


sidntrivedi:
Please hold on the review. Some golangci-lint findings have been reported after version change in custom-gcl.yaml. Will ping for review post fixing those.

internal/controller/cluster/reconciler.go
```diff
@@ -1016,11 +1016,7 @@ func (r *SingleClusterReconciler) checkPreviouslyFailedCluster(ctx context.Conte
 
 	// All pods have hard-failed and status is empty — the cluster failed during
 	// its initial create and needs to be recovered.
-	if err := r.recoverFailedCreate(ctx); err != nil {
-		return true, common.ReconcileError(err)
-	}
-
-	return true, common.ReconcileSuccess()
+	return true, common.ReconcileError(r.recoverFailedCreate(ctx))
```
sidntrivedi: staticcheck finding.

r.recoverFailedCreate() is never returning nil error. Hence, the true, common.ReconcileSuccess() was dead code.

internal/controller/cluster/statefulset.go
```diff
@@ -111,14 +111,16 @@ func (r *SingleClusterReconciler) createSTS(
 	)
 
 	tlsName, _ := asdbv1.GetServiceTLSNameAndPort(r.aeroCluster.Spec.AerospikeConfig)
-	envVarList := []corev1.EnvVar{
+	envVarList := make([]corev1.EnvVar, 0, 6)
```
sidntrivedi: Please review if we need this pre-allocation or an overkill. One thing we might need to keep in mind in future is to keep on increasing its capacity here if we append more env vars to the slice.

test/utils.go
```diff
@@ -36,18 +36,8 @@ func InitialiseClients(scheme *runtime.Scheme, cfg *rest.Config) (
...
-		return k8sClient, k8sClientSet, err
-	}
-
 	k8sClientSet = kubernetes.NewForConfigOrDie(cfg)
 
-	if k8sClientSet == nil {
-		err = fmt.Errorf("k8sClientSet is nil")
-		return k8sClient, k8sClientSet, err
-	}
-
```
sidntrivedi, line 50: staticcheck finding. dead code since at this point k8sClientSet is not nil.

test/cluster/test_client.go
```diff
@@ -316,16 +315,43 @@ func appendCACertFromFileOrPath(
...
+
+	root, err := os.OpenRoot(caPath)
+	if err != nil {
+		logrus.Info("Failed to open CA path", "caPath: ", caPath)
+		return serverPool
+	}
+	defer root.Close()
+
+	err = fs.WalkDir(
+		root.FS(), ".", func(path string, d fs.DirEntry, err error) error {
```
sidntrivedi: Tanmay's review: Move duplicate code to separate function.

internal/controller/cluster/reconciler.go
sidntrivedi: Tanmay's review: Revert this change.

internal/controller/cluster/statefulset.go
sidntrivedi: Revert this.

test/utils.go
sidntrivedi: Tanmay's review: Revert this change.

test/cluster/test_client.go
sidntrivedi: Done

test/utils.go
sidntrivedi: Done

internal/controller/cluster/reconciler.go
sidntrivedi: Done

internal/controller/cluster/client_policy.go
```diff
@@ -244,37 +244,74 @@ func (r *SingleClusterReconciler) appendCACertFromFileOrPath(
...
+	}
+
+	root, err := os.OpenRoot(caPath)
+	if err != nil {
+		return fmt.Errorf("open CA path %s: %w", caPath, err)
+	}
+	defer root.Close()
+
+	return fs.WalkDir(
+		root.FS(), ".", func(path string, d fs.DirEntry, err error) error {
```
sidntrivedi: Reason for using `os.Root` and `fs.WalkDir` instead of `filepath.WalkDir` which was previously being used:

gosec flags any file-read call whose path argument isn't a string literal as a potential path-traversal vector, since a malicious/misconfigured `caPath` (e.g. containing `../` or symlinks) could make the walk read files outside the intended directory.

The fix uses `os.OpenRoot(caPath)` to scope all reads to within caPath, then walks via `root.FS()/fs.WalkDir` and reads via `root.ReadFile(path)` - reads can't escape the root even via symlinks, which both satisfies gosec and closes the actual traversal risk.

api/v1/groupversion_info.go
```diff
@@ -20,17 +20,39 @@ limitations under the License.
...
+func (bld *SchemeBuilderType) Register(object ...runtime.Object) *SchemeBuilderType {
+	bld.SchemeBuilder.Register(func(scheme *runtime.Scheme) error {
+		scheme.AddKnownTypes(bld.GroupVersion, object...)
+		metav1.AddToGroupVersion(scheme, bld.GroupVersion)
+
+		return nil
+	})
+
+	return bld
+}
```
sidntrivedi, line 46: Why: `scheme.Builder` deprecation

Bumping controller-runtime to v0.24.1 (needed for K8s 1.36 compatibility) surfaced a staticcheck finding:

```
SA1019: scheme.Builder is deprecated: This helper is only useful in api packages,
but api packages should be easy to import and hence have minimal dependencies.
```

Controller-runtime deprecated `pkg/scheme.Builder` because API packages are meant to stay dependency-light - ideally just the standard library, k8s.io/apimachinery, and other API packages. Depending on controller-runtime pulls in a much heavier tree (clients, managers, webhooks) that consumers of our types shouldn't need.

Fix: `api/v1/groupversion_info.go` now has a local `SchemeBuilderType`, a minimal reimplementation of the same behavior using only `k8s.io/apimachinery`, so `api/v1` no longer depends on controller-runtime.

internal/controller/cluster/rack.go
```diff
@@ -712,6 +712,7 @@ func (r *SingleClusterReconciler) reconcileRack(
 				r.Recorder.Eventf(
 					r.aeroCluster, corev1.EventTypeWarning,
 					"RackScaleDownFailed",
+					"%s",
```
sidntrivedi: How was it working earlier? 

`Eventf(obj, eventtype, reason, messageFmt, args...)` was being called with the dynamically built message as the format string itself and no args, e.g.:

```
r.Recorder.Eventf(r.aeroCluster, ..., "RackScaleDownFailed",
    eventRackScaleFailureMessageWithCause(...))
```

Under the hood, `Eventf` runs `fmt.Sprintf(messageFmt, args...)`. With zero args, `Sprintf` just returns the string unchanged - unless it happens to contain a stray `%`. Since none of the generated messages ever contained a %, it looked fine. But any error text embedding a `%` (a percentage, an image tag, etc.) would have been misinterpreted as a format directive and corrupted the event message.

It has been fixed now.

go.mod
```diff
@@ -15,16 +15,21 @@ require (
...
+	sigs.k8s.io/controller-runtime v0.24.1
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require (
+	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
+	k8s.io/streaming v0.36.4 // indirect
+)
+
 require (
```
abhishekdwivedi3060: Club these

api/v1/groupversion_info.go
abhishekdwivedi3060: How did we come up with this approach?
Controller-runtime PR which deprecated this scheme.Builder has a different approach and the same approach is used by others it seems. Confirm this once: https://github.com/kubernetes-sigs/controller-runtime/pull/3461

.golangci.yml
```diff
@@ -95,7 +96,11 @@ linters:
       - third_party$
       - builtin$
       - examples$
-    rules: # Allow dot imports (ST1001) in test code for Ginkgo/Gomega
+    rules:
+      - linters:
+          - goconst
+        path: ^test/
+      # Allow dot imports (ST1001) in test code for Ginkgo/Gomega
```
abhishekdwivedi3060, line 103: How is this different from the `ignore-tests: true` above?
Generally we try to follow same coding guidelines for tests also in this repo.
Are we seeing too many violations due to golangci-lint bump?

internal/controller/cluster/rack.go
```diff
@@ -738,6 +738,7 @@ func (r *SingleClusterReconciler) reconcileRack(
...
```
abhishekdwivedi3060: Instead of fixing the formatting with redundant %s, why not use Event() instead of Eventf()

internal/controller/cluster/reconciler.go
abhishekdwivedi3060: Can we not fix the behaviour instead of adding no-lint? It'll be a small fix.
1. Return nil in recoverFailedCreate in the very end instead of "fmt.Errorf("forcing recreate of cluster: status is nil")"
2. Let the caller checkPreviouslyFailedCluster decide whether to error our or not. So it becomes

```suggestion
//nolint:staticcheck // SA4023
func (r *SingleClusterReconciler) recoverFailedCreate(ctx context.Context) error {
r.Log.Info("Forcing a cluster recreate as status is nil. The cluster could be unreachable due to bad configuration")

@@ -1191,27 +1196,29 @@

// All pods have hard-failed and status is empty — the cluster failed during
// its initial create and needs to be recovered.
	//nolint:staticcheck // SA4023: recoverFailedCreate always returns a non-nil error today, but the check
	// guards against a future change to that function
if err := r.recoverFailedCreate(ctx); err != nil {
return true, common.ReconcileError(err)
}

return true, common.ReconcileError(fmt.Errorf("forcing recreate of cluster: status is nil"))
```

test/utils.go
abhishekdwivedi3060: If the client.New never returns a nil client with nil error then we can remove these redundant checks

internal/controller/cluster/client_policy.go
```diff
@@ -244,37 +244,74 @@ func (r *SingleClusterReconciler) appendCACertFromFileOrPath(
...
+
+	return serverPool
+}
+
+// LoadCACertsFromFileOrPath loads PEM-encoded CA certificates from caPath - a single cert
+// file or a directory containing cert files - into serverPool.
+// onFileLoaded, if non-nil, is called with the path of each file after its certs
+// are appended, so callers can log with their own logger.
+func LoadCACertsFromFileOrPath(
+	caPath string, serverPool *x509.CertPool, onFileLoaded func(filePath string),
```
abhishekdwivedi3060: This onFileLoaded param seems like an overkill for a simple use of logging.
We can achieve the same result by passing the logger or returning the files loaded and then logging in the caller func

abhishekdwivedi3060 (changes requested):

api/v1/groupversion_info.go
sidntrivedi: Had made the change this way to not modify the types files since in this upstream way, the init() function would not be required. Also, to register the types closer to where they are defined.

But modified now to follow this upstream way and realized it is better.

.golangci.yml
sidntrivedi: This is used to also ignore the goconst reports from the test helper files like `test/cluster/utils.go`, `test/cluster/cluster_helper.go` and `test/envtests/cluster/cluster_webhook_utils.go`.

If tests and test helpers are not ignored, a total of 346 lints are being reported.

```
346 issues:
* goconst: 346
```

internal/controller/cluster/client_policy.go
sidntrivedi: Removed this fix for now.

go.mod
sidntrivedi: Done

internal/controller/cluster/rack.go
sidntrivedi: Done, using Event() now.

internal/controller/cluster/reconciler.go
sidntrivedi: Done

test/utils.go
sidntrivedi: Done

abhishekdwivedi3060 (approved):

shubhamyelikar99 (approved):

tanmayja (approved):

PR #595, opened 2026-09-04
feat: block unsafe vertical scaling in the validating webhook [KO-619]

shubhamyelikar99:
[vertical-scaling-test-results-approach-2.md](https://github.com/user-attachments/files/31836875/vertical-scaling-test-results-approach-2.md)
## Summary
- Block rack revision changes (vertical scaling) that would take the cluster down: size 1, replication-factor 1 (AP and SC), and SC clusters that would fall below RF or lose roster majority while a rack is replaced.
- Size the worst batch against the first rack (`DistributeItems` always gives it the most pods). 
- Known gap, not covered: scale-up/scale-down + revision bump in one apply can still be admitted and fail at runtime, roster related fields like forceBlockFromRoster, rosterNodeBlockList are also not covered

## Test plan
- [ ] Unit tests in `aerospikecluster_vertical_scaling_test.go`
- [ ] `make env-test-cluster` (rack-revision webhook specs, including majority cases)

### DIFF
--- a/internal/webhook/v1/aerospikecluster_validating_webhook.go
+++ b/internal/webhook/v1/aerospikecluster_validating_webhook.go
@@ -310,6 +310,10 @@ func (acv *AerospikeClusterCustomValidator) ValidateUpdate(_ context.Context,
 		return warnings, err
 	}
 
+	if err := validateVerticalScaling(oldObject, aerospikeCluster); err != nil {
+		return warnings, err
+	}
+
 	// Validate actual pod names to catch silent DNS label overflows at runtime.
 	// This uses real rack IDs, real revisions, and the real max ordinal (Size-1),
 	// so it only rejects what would actually fail — not a conservative estimate.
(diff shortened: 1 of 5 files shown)


internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -873,6 +877,130 @@ func validateConcurrentRackRevisions(oldObj, newObj *asdbv1.AerospikeCluster) er
...
+
+	// AP and SC: replacing the only node takes the cluster down.
+	if newObj.Spec.Size == 1 {
+		return fmt.Errorf(
+			"vertical scaling (rack revision change) is not allowed: spec.size is 1, so the only node would go down",
+		)
+	}
+
+	nsConfs := getNsConfForNamespaces(newObj.Spec.RackConfig)
+	nsNames := sets.List(sets.KeySet(nsConfs))
```
tanmayja: why do we need to compute nsNames. can we run loop on nsConfs directly?

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -873,6 +877,130 @@ func validateConcurrentRackRevisions(oldObj, newObj *asdbv1.AerospikeCluster) er
...
+	return false
+}
+
+// podsLeavingFirstRack is the largest delete batch in the cluster: DistributeItems
+// always puts the most pods on spec index 0.
+func podsLeavingFirstRack(cluster *asdbv1.AerospikeCluster) int32 {
+	batch := cluster.Spec.RackConfig.RollingUpdateBatchSize
+	racks := cluster.Spec.RackConfig.Racks
+
+	if len(racks) == 0 {
```
tanmayja: this can never be true here

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -873,6 +877,130 @@ func validateConcurrentRackRevisions(oldObj, newObj *asdbv1.AerospikeCluster) er
...
+		return clampedBatch(batch, cluster.Spec.Size)
+	}
+
+	topology := asdbv1.DistributeItems(cluster.Spec.Size, utils.Len32(racks))
```
tanmayja: if we just need 1st rack's size, then calcluating and storing whole topolgy is not needed. If it is just maths, we can just calculate 1st rack's size.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -873,6 +877,130 @@ func validateConcurrentRackRevisions(oldObj, newObj *asdbv1.AerospikeCluster) er
...
+	hasSC := false
+
+	for _, nsName := range nsNames {
+		if !nsConfs[nsName].scEnabled {
+			continue
+		}
+
+		hasSC = true
+
+		if int(survivors) < nsConfs[nsName].replicationFactor {
```
tanmayja: I think idea here is to compare the survivors with minimum rf set in any namespace.
similar logic is written in validateMaxUnavailable func.
see if we can write the whole thing in a better structure.

tanmayja: check if user can skip setting nsConfs[nsName].replicationFactor.

internal/webhook/v1/aerospikecluster_validating_webhook.go
shubhamyelikar99: Dropped. It was only there to keep the error text deterministic, since the messages name a namespace and Go randomizes map iteration order. The namespace checks are now a single pass over nsConfs directly, the way validateBatchSize already does it, so there is nothing left to sort.

internal/webhook/v1/aerospikecluster_validating_webhook.go
shubhamyelikar99: Right, removed.

internal/webhook/v1/aerospikecluster_validating_webhook.go
shubhamyelikar99: Agreed, it's just maths. DistributeItems gives element i as size/n + (1 if i < size%n), so element 0 is ceil(size/n), which is (size + n - 1) / n in integer arithmetic — and index 0 is always the max, since the remainder is handed out starting from index 0.

internal/webhook/v1/aerospikecluster_validating_webhook.go
shubhamyelikar99: Checked — yes, users can omit it, and it's already safe. getNsConfForNamespaces reads RF through validation.GetNamespaceReplicationFactor, which defaults it to Aerospike's default of 2 when the key is absent, so an unset replication-factor arrives here as 2, never 0.

One gap worth closing anyway: that function does rf, _ :=, so a present-but-unparsable RF leaves rf at 0. == 1 would miss it, and worse, survivors < 0 is never true, so the SC floor would silently pass. Switched to <= 1, matching what validateBatchSize already does. In practice validate() runs validateAerospikeConfig at the top of ValidateUpdate and rejects a malformed RF first — validateMaxUnavailable leans on that same ordering — so this is belt and braces.

Separately, the swallowed error plus three unchecked type assertions in getNsConfForNamespaces mean a malformed CR would panic the webhook rather than be rejected. Pre-existing and shielded by that ordering — happy to file a follow-up if you want it cleaned up properly.

shubhamyelikar99: Restructured — the namespace checks are now a single summarizeNsReplication pass, in the shape you're pointing at.

One thing I want to flag though: taking the minimum here would be a bug. The survivors check is a lower bound — survivors >= rf has to hold for every SC namespace, and "any namespace fails" is the same thing as "the namespace with the highest RF fails". So the binding value is the max, not the min.

Concretely: two SC namespaces at RF 3 and RF 5, with 4 survivors. Using min gives 3, so 4 >= 3 passes and the update is admitted — but the RF 5 namespace is left stranded at 4 nodes. Using max gives 5, 4 < 5, correctly rejected.

validateMaxUnavailable legitimately reduces with min because its constraint is an upper bound: maxUnavailable < rf for every namespace is equivalent to maxUnavailable < min(rf). Ours is a lower bound, so the direction inverts. I borrowed the structure — one reduction, threshold named in the error — but not the direction. The old per-namespace loop was already max-equivalent, so this is a pure refactor, not a behaviour change.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -873,6 +877,132 @@ func validateConcurrentRackRevisions(oldObj, newObj *asdbv1.AerospikeCluster) er
...
 }
 
+// validateVerticalScaling rejects a rack revision change that would leave too few
+// nodes standing while a rack's pods are replaced.
+//
+// Inputs are spec-only. Size 1 and replication-factor <= 1 are rejected in AP and
+// SC. The survivors floor and roster-majority checks are SC only (pkg/validation
+// already allows AP size < RF). The worst batch is the first rack: DistributeItems
+// always gives it the most pods.
+func validateVerticalScaling(oldObj, newObj *asdbv1.AerospikeCluster) error {
```
abhishekdwivedi3060: The main use-case of rack revision change is vertical scaling but we generally don't call that feature as vertical scaling, so rename it to something related to RackRevision.
Also, change the error msg and comments accordingly

test/envtests/cluster/cluster_webhook_utils.go
```diff
@@ -31,6 +31,33 @@ func uniqueNamespacedName(suffix string) types.NamespacedName {
...
+		}
+	}
+
+	if cluster.Spec.AerospikeConfig != nil {
+		setRF(cluster.Spec.AerospikeConfig.Value)
+	}
+
+	for idx := range cluster.Spec.RackConfig.Racks {
+		setRF(cluster.Spec.RackConfig.Racks[idx].AerospikeConfig.Value)
+	}
```
abhishekdwivedi3060: Not required. Global spec level AerospikeConfig is mutated at the rack level config

tanmayja (dismissed):

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -873,6 +877,132 @@ func validateConcurrentRackRevisions(oldObj, newObj *asdbv1.AerospikeCluster) er
...
+			)
+		}
+	}
+
+	return nil
+}
+
+// hasRackRevisionChange is true when an existing rack's revision changed. A newly
+// added rack is not vertical scaling. A revert counts.
+func hasRackRevisionChange(oldRacks, newRacks []asdbv1.Rack) bool {
```
abhishekdwivedi3060: Did you consider comparing with the status also for a three-way comparison? Is it going beyond what we discussed?
Rack revision change then later scale down in the next update before success can again lead to same issues. See if rack revision change can also be added here.

shubhamyelikar99: Good catch — added the status arm.

hasRackRevisionChange now compares the new spec against both the old spec and oldObj.Status.RackConfig, keyed by rack ID. Since status.RackConfig is only written on a successful reconcile, a spec/status mismatch means a migration is still rolling, so a follow-up resize stays armed. 

Also added a reflect.DeepEqual(oldObj.Spec, newObj.Spec) short-circuit, so metadata-only writes (finalizer removal included) don't re-run the check mid-migration.

abhishekdwivedi3060 (approved):

PR #596, opened 2026-09-09
feat: add webhook TLS cert provider modes in helm chart [KO-605]

shubhamyelikar99:
## What

Introduces `certs.webhook.provider` as the single source of truth for how the webhook serving certificate is provided, so cert-manager is no longer mandatory for development and POC installs.

- **`certManager`** - chart creates the cert-manager `Issuer` and `Certificate`; cainjector populates the `caBundle`. Current behaviour, and still the default.
- **`selfSigned`** - chart generates a CA and serving certificate with Sprig and creates the `Secret` itself. An existing `Secret` is reused on upgrade so the certificate is not rotated. Development and POC only; see the README caveats.
- **`external`** - chart creates nothing. The user pre-creates the `Secret` and supplies `certs.webhook.external.caBundle`.

`certs.webhook.create` is deprecated but still honoured, and is dropped from the schema's `required` list. An explicit `create: false` renders exactly what 4.5.0 rendered: no `Issuer` and no `Certificate`, while `cert-manager.io/inject-ca-from` is still emitted so cainjector fills in the `caBundle` from a `Certificate` the user manages. Only a literal `false` selects that path — an absent key falls through to `certManager`, since 4.5.0 required the key and could not produce that input.

Supporting changes:

- The `inject-ca-from` annotation and an inlined `caBundle` are mutually exclusive by construction, so cainjector can never overwrite a chart-supplied bundle.
- `caBundle` is set on all seven webhook entries (five validating, two mutating).
- SANs come from one shared helper, so the generated certificate cannot drift from the cert-manager `Certificate`.
- `external.caBundle` is required only when `provider` is `external`, is validated as base64 (raw PEM is rejected at install time), and has whitespace stripped so wrapped or newline-terminated input works.
- `NOTES.txt` reports the resolved mode, the mode-specific caveats, and a `[DEPRECATED]` line on the legacy path only.

This supersedes #563, which addresses the same `external` use case by inferring the mode from a non-empty `caBundle`. Its invocation `--set certs.webhook.create=false --set certs.webhook.caBundle=<b64>` becomes `--set certs.webhook.provider=external --set certs.webhook.external.caBundle=<b64>`.

## Testing

**Unit** - 154 helm-unittest cases across 7 suites. `helm lint` clean in all five modes; `values.schema.json` regenerates with no drift.

**Live cluster** - every scenario below was run on GKE with Helm v4.1.4 and cert-manager installed, against a live `AerospikeCluster`, so admission was exercised for real (`failurePolicy: Fail`).

Two bugs were found by the live run and fixed here: an ownership pre-check that blocked the very `--take-ownership` flag its own error recommended (a template cannot detect that flag), and a `toYaml` round trip that truncated one byte from `tls.key`.

Not covered: clock skew on the `selfSigned` path, which needs the workstation clock changed. `notBefore` is the render machine's clock with no backdating, so positive skew delays validity; documented as a limitation.

<details>
<summary><b>All 56 verified scenarios</b></summary>

### Rendering (11)

| # | Scenario | Result |
| --- | --- | --- |
| 1 | Default values vs the deployed release | identical, 22 docs each |
| 2 | `provider: certManager` | Issuer 1, Certificate 1, annotation 2, no inlined bundle |
| 3 | Legacy `create: false` | no Issuer, no Certificate, annotation retained |
| 4 | `create: false` + `metrics.create: true` | shared Issuer retained, metrics Certificate only |
| 5 | `create` absent (null) | falls through to `certManager` (4.5.0 rejected this input) |
| 6 | `provider: certManager` + `create: false` | provider wins |
| 7 | `provider: selfSigned` certificate quality | chain OK, key/cert pair matches, correct SANs, 10-year validity |
| 8 | One certificate per render | 1 distinct `caBundle` across all entries (memoization holds) |
| 9 | PEM trailing newlines | `0x0a` on `ca.crt`, `tls.crt`, `tls.key` |
| 10 | `provider: external` | creates nothing, supplied bundle on all 7 entries |
| 11 | YAML validity + annotation XOR `caBundle` | 24/22/24/23/22 docs, XOR holds in all 5 modes |

### Schema validation (17)

| # | Input | Result |
| --- | --- | --- |
| 12 | `provider` not in the enum | rejected |
| 13 | `provider: external`, `caBundle` omitted | rejected |
| 14 | `provider: external`, `caBundle` empty | rejected |
| 15 | `caBundle` given as raw PEM | rejected on the base64 pattern |
| 16 | unknown key under `certs.webhook.certManager` | rejected |
| 17 | unknown key under `certs.webhook.selfSigned` | rejected |
| 18 | unknown key under `certs.webhook.external` | rejected |
| 19 | `webhookServerCertSecretName` empty | rejected |
| 20 | `certs` nulled out | rejected |
| 21 | `certs.webhook` nulled out | rejected |
| 22 | `certs.metrics` nulled out | rejected |
| 23 | `provider: ""` | accepted, infers from `create` |
| 24 | `certManager` without `caBundle` | accepted |
| 25 | `selfSigned` without `caBundle` | accepted |
| 26 | `create` key absent | accepted, no longer required |
| 27 | `caBundle` input forms | 10 accepted (`--set`, `--set-string`, `--set-file` x3, `--set-json`, 4 YAML scalar styles), all converging on the same rendered value |
| 28 | `caBundle` bad input | 4 rejected (raw PEM via `-f` and `--set-file`, empty string, omitted) |

### Upgrade paths (10)

| # | From | To | Result |
| --- | --- | --- | --- |
| 29 | published 4.5.0 | branch, same values | no-op; same pods, same certificate |
| 30 | `certManager` | `selfSigned`, no flag | refused on the Secret; nothing changed |
| 31 | `certManager` | `selfSigned` + `--take-ownership` | adopted, certificate reused, no admission gap |
| 32 | `selfSigned` | `selfSigned` (re-run) | no rotation, NOTES reports "Reused" |
| 33 | `selfSigned` | `certManager` | new certificate; admission recovered in 4s |
| 34 | *(none)* | 4.5.0 with `create: false` | legacy state built; chart owns no cert-manager objects |
| 35 | legacy `create: false` | branch, flag not re-passed | refused on the Certificate (Helm drops unsupplied values) |
| 36 | legacy `create: false` | branch with `--reuse-values` | no-op; certificate and pods untouched |
| 37 | legacy `create: false` | branch re-passing `--set` | no-op |
| 38 | legacy `create: false` | `certManager`, own Certificate deleted first | chart takes ownership, new certificate |

### Fresh installs and runtime behaviour (18)

| # | Scenario | Result |
| --- | --- | --- |
| 39 | Fresh install, `certManager` | 2 cert-manager objects Ready, admission OK |
| 40 | Fresh install, `selfSigned` | 0 cert-manager objects, NOTES "Generated a new", admission OK |
| 41 | `selfSigned` generated cert on cluster | chain OK, SANs use the real namespace, not `default` |
| 42 | Fresh install, `external` with an openssl-issued PKI | chart creates nothing, admission OK |
| 43 | `external` with a wrong `caBundle` | admission fails, proving the field is honoured |
| 44 | `external` restored to the correct `caBundle` | admission recovers |
| 45 | `caBundle` round trip | rendered value decodes back to the original CA byte-for-byte |
| 46 | Uninstall while on `selfSigned` | chart-owned Secret is deleted |
| 47 | Uninstall while on `external` | user's Secret survives |
| 48 | `metrics.create: true` on a legacy release | shared Issuer created, metrics Certificate Ready |
| 49 | Metrics cert reaches the pod | `ca.crt tls.crt tls.key` present in the mount |
| 50 | `metrics.create: false` again | metrics Certificate and shared Issuer removed, no orphan |
| 51 | Shared `issuerRef` matrix | 8 webhook-mode x `metrics.create` combinations, no dangling reference |
| 52 | `selfSigned` under GitOps | two renders differ, confirming it is unsuitable for render-and-apply |
| 53 | `selfSigned` + `safePodEviction` | the 5th validating entry also carries the bundle |
| 54 | `selfSigned` + custom Secret name | Secret created and mounted under the custom name |
| 55 | `selfSigned` in another namespace | SANs follow the release namespace |
| 56 | Admission after every live change | verified by the webhook returning its own validation warning |

</details>


helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -38,12 +38,37 @@ metricsPort: 8443 # @schema required; minimum: 1; maximum: 65535
...
+  webhook: # @schema required
+    # -- How the webhook serving certificate is provided.
+    # certManager: chart creates the cert-manager Issuer/Certificate (requires cert-manager).
+    # selfSigned: chart generates the cert and Secret itself. Dev/POC only, see README.
+    # external: you pre-create the Secret and supply external.caBundle.
+    # Empty infers the mode from the deprecated `create` flag (4.5.0 compatibility).
+    provider: "" # @schema enum: ["", certManager, selfSigned, external]
+
+    # -- DEPRECATED: use `provider` instead. Only consulted when `provider` is empty.
+    # Removed in a future release.
```
abhishekdwivedi3060, line 53: nit: 
```suggestion
    # Will be removed in a future release.
```

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -23,9 +23,112 @@ Create the namespace where AKO will be installed. Replace the placeholder `<name
 kubectl create namespace <namespace>
 ```
 
-### Deploy Cert-manager
+### Deploy Cert-manager (conditional)
```
abhishekdwivedi3060, line 26: Lets call this as "Setup Webhook Certificates" and under that we can have 3 subheading for certManager, selfSigned and external

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -23,9 +23,112 @@ Create the namespace where AKO will be installed. Replace the placeholder `<name
...
+```
+
+The chart generates the certificate and reuses the existing `Secret` on subsequent
+upgrades so that `helm upgrade` does not rotate it. Two caveats make this unsuitable for
+production:
+
+* The certificate is valid for 10 years and nothing renews it.
+* Its validity window starts at the clock of the machine that runs `helm`. If that clock
+  runs ahead of the cluster's, the API server sees the certificate as not yet valid and
+  rejects admission until the skew elapses.
```
abhishekdwivedi3060, line 65: This seems like a genuine issue. Is there a workaround for this?

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -23,9 +23,112 @@ Create the namespace where AKO will be installed. Replace the placeholder `<name
...
+the ownership markers it is missing.
+
+*Preferred — let Helm adopt the existing `Secret`.* The chart reuses the certificate
+already in the `Secret`, so the served certificate does not change and there is no
+admission gap:
+
+```sh
+helm upgrade aerospike-kubernetes-operator ... --take-ownership \
+  --set certs.webhook.provider=selfSigned
+```
```
abhishekdwivedi3060, line 111: Is this only for certManager -> selfSigned switch? 
Also, isn't the preferred approach going to be a problem? With the switch, certManager certificates going to be deleted along with its Secret.
Instead of above switch, I would cover cert-manager to external steps here and remove above steps because selfSigned is meant only for dev

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
\ No newline at end of file
+{{- end -}}
+
+{{/*
+Name of the webhook Service, and the two DNS names that resolve to it. Single source of
+truth for the SANs: the cert-manager Certificate's dnsNames and the Sprig-generated
+self-signed certificate must agree, or the API server rejects the TLS handshake.
+*/}}
+{{- define "aerospike-kubernetes-operator.webhookServiceName" -}}
+aerospike-operator-webhook-service
```
abhishekdwivedi3060, line 72: This is hardcoded at multiple places in the current master. If we are defining a constant for that then lets use it everywhere

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+"I create the Certificate named aerospike-operator-serving-cert myself, cainjector still
+fills in the caBundle", not "no cert-manager". That maps onto none of the three
+user-facing providers, so it resolves to the internal `legacyCertManager` state, which
+renders exactly what 4.5.0 did. `legacyCertManager` is not a valid `provider` value and
+cannot be selected deliberately.
+
+Only a literal boolean false selects the legacy state. An absent or null `create` is not
+the 4.5.0 opt-out (4.5.0's schema required the key, so that input could not exist) and
+must fall through to the default, or a dropped key would silently produce an install with
+no Certificate and no Secret.
```
abhishekdwivedi3060, line 94: Schema changes are added recently (yet to be release) and not present in 4.5.0

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+*/}}
+{{- define "aerospike-kubernetes-operator.webhookCertProvider" -}}
+{{- $webhook := .Values.certs.webhook -}}
+{{- if $webhook.provider -}}
+{{- $webhook.provider -}}
+{{- else if and (kindIs "bool" $webhook.create) (not $webhook.create) -}}
```
abhishekdwivedi3060, line 100: Is this check to handle null value for create? We don't handle null value in previous versions

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+legacyCertManager
+{{- else -}}
+certManager
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create the cert-manager Issuer/Certificate for the webhook. Only the explicit
+certManager provider does; legacyCertManager deliberately does not.
```
abhishekdwivedi3060, line 109:
```suggestion
certManager provider or deprecated create: true enables this; legacyCertManager deliberately does not.
```

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+*/}}
+{{- define "aerospike-kubernetes-operator.webhookCreateCertManagerResources" -}}
+{{- if eq (include "aerospike-kubernetes-operator.webhookCertProvider" .) "certManager" -}}true{{- end -}}
+{{- end -}}
+
+{{/*
+Base64-encoded CA bundle to inline into each webhook clientConfig. Empty for the two
+cert-manager states, where cainjector owns the field — so an empty result is also what
+tells the webhook templates to emit the inject-ca-from annotation, which keeps "annotation
+XOR inlined caBundle" true by construction rather than by a second, separate predicate.
```
abhishekdwivedi3060, line 119: Can you simplify this comment? Its not very clear at the moment.

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+{{- define "aerospike-kubernetes-operator.webhookSelfSignedCert" -}}
+{{- $cached := .Values._akoWebhookSelfSignedCert -}}
+{{- if not $cached -}}
+  {{- $name := .Values.certs.webhook.webhookServerCertSecretName -}}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace $name -}}
+  {{- $crt := dig "data" "tls.crt" "" $secret -}}
+  {{- $key := dig "data" "tls.key" "" $secret -}}
+  {{- $ca := dig "data" "ca.crt" "" $secret -}}
+  {{- if and $crt $key $ca -}}
+    {{- $cached = dict "crt" $crt "key" $key "ca" $ca "reused" true -}}
```
abhishekdwivedi3060, line 176: I think we don't need to support the use-case of moving from certManager -> selfSigned at the moment, at selfSigned is meant for Dev.
Blindly taking the ownership of certManager secret during migration is not full proof as we don't know the the certs expiry details. Also, it not created by us. We should not open that window for the user.
Instead, I would suggest to check the ownership of the secret also before reusing.
{{- $owner := dig "metadata" "annotations" "meta.helm.sh/release-name" "" $secret -}}
{{- if and $crt $key $ca (eq $owner .Release.Name) -}}

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+  {{- else -}}
+    {{/* 10 years: nothing renews a chart-generated cert, and the mode is dev/POC only.
+         NotBefore is the render machine's clock; see the clock-skew note in the README. */}}
+    {{- $svc := include "aerospike-kubernetes-operator.webhookServiceName" . -}}
+    {{- $dnsNames := include "aerospike-kubernetes-operator.webhookDnsNames" . | fromYamlArray -}}
+    {{- $rootCa := genCA (printf "%s-ca" $svc) 3650 -}}
+    {{- $cert := genSignedCert $svc nil $dnsNames 3650 $rootCa -}}
```
abhishekdwivedi3060, line 183: Lets them keep as 1 year only for now. We'll revisit this based on feedback

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+    {{- $cached = dict "crt" ($cert.Cert | b64enc) "key" ($cert.Key | b64enc) "ca" ($rootCa.Cert | b64enc) "reused" false -}}
+  {{- end -}}
+  {{- $_ := set .Values "_akoWebhookSelfSignedCert" $cached -}}
+{{- end -}}
+{{- toYaml $cached -}}
```
abhishekdwivedi3060, line 188: Can you confirm this cached value is saved in the release object?
Can a user get this certificated info using helm CLI like "helm get all <release>"

helm-charts/aerospike-kubernetes-operator/templates/NOTES.txt
```diff
@@ -19,3 +19,52 @@ $ helm status {{ .Release.Name }} --namespace {{ .Release.Namespace }}
...
+Mode: certManager — the chart created the cert-manager Issuer and Certificate, and
+cainjector fills in the caBundle on the webhook configurations.
+{{- if not .Values.certs.webhook.provider }}
+Note: certs.webhook.provider is unset, so the mode was inferred from certs.webhook.create.
+Set `certs.webhook.provider: certManager` to make this explicit.
+{{- end }}
+{{- else if eq $provider "legacyCertManager" }}
+Mode: cert-manager, with the chart creating neither the Issuer nor the Certificate
+(inferred from the deprecated certs.webhook.create=false). cainjector still fills in
+the caBundle, so you must supply a Certificate named aerospike-operator-serving-cert.
```
abhishekdwivedi3060, line 35: nit;
```suggestion
the caBundle, so you must supply a Certificate named aerospike-operator-serving-cert from outside.
```

helm-charts/aerospike-kubernetes-operator/templates/NOTES.txt
```diff
@@ -19,3 +19,52 @@ $ helm status {{ .Release.Name }} --namespace {{ .Release.Namespace }}
...
+
+[DEPRECATED] certs.webhook.create is deprecated and will be removed in a future
+release. This install relies on it. There is no `provider` value with these exact
+semantics — if you create the Certificate yourself, keep using create=false for now;
+if you want the chart to create it, set `certs.webhook.provider: certManager`.
```
abhishekdwivedi3060, line 40:
```suggestion
[DEPRECATED] certs.webhook.create is deprecated and will be removed in a future
release.
```

helm-charts/aerospike-kubernetes-operator/tests/certs_test.yaml
```diff
@@ -131,6 +133,150 @@ tests:
...
+
+  - it: accepts single-line base64 for external.caBundle
+    set:
+      certs:
+        webhook:
+          provider: external
+          external:
+            caBundle: "LS0tLS1CRUdJTg=="
+    asserts:
+      - notFailedTemplate: {}
```
abhishekdwivedi3060, line 221: Isn't this test-case already covered at line 163 above

helm-charts/aerospike-kubernetes-operator/tests/webhook_certs_test.yaml
```diff
@@ -0,0 +1,354 @@
...
+          count: 1
+        template: templates/aerospike-operator-webhook-serving-cert.yaml
+      - equal:
+          path: metadata.name
+          value: aerospike-operator-serving-cert
+        template: templates/aerospike-operator-webhook-serving-cert.yaml
+      - hasDocuments:
+          count: 0
+        template: templates/aerospike-operator-webhook-selfsigned-secret.yaml
+
```
abhishekdwivedi3060, line 119: Same comment. 4.5.0 version didn't have helm schema.
Check if the test-case is still valid

helm-charts/aerospike-kubernetes-operator/tests/webhook_certs_test.yaml
```diff
@@ -0,0 +1,354 @@
...
+          path: metadata.annotations["cert-manager.io/inject-ca-from"]
+        template: templates/aerospike-operator-validating-webhook-configuration.yaml
+      - equal:
+          path: webhooks[0].clientConfig.caBundle
+          value: "LS0tLS1CRUdJTg=="
+        template: templates/aerospike-operator-validating-webhook-configuration.yaml
+
+  # ------------------------------------------------------------------------------
+  # caBundle must reach EVERY webhook entry. The mutating configuration is the one
+  # that gets forgotten: 5 validating entries (with safePodEviction) + 2 mutating.
```
abhishekdwivedi3060, line 243: "The mutating configuration is the one that gets forgotten" ???

helm-charts/aerospike-kubernetes-operator/tests/webhook_selfsigned_lookup_test.yaml
```diff
@@ -0,0 +1,171 @@
...
+      # Reusing rather than regenerating is what makes `helm upgrade --take-ownership`
+      # an outage-free migration: the served certificate never changes.
+      - equal:
+          path: data["tls.crt"]
+          value: Y3J0Cg==
+        template: templates/aerospike-operator-webhook-selfsigned-secret.yaml
+      - equal:
+          path: webhooks[*].clientConfig.caBundle
+          value: Y2EK
+        template: templates/aerospike-operator-validating-webhook-configuration.yaml
```
abhishekdwivedi3060, line 144: This might go away after incorporating the comment of checking the ownership

abhishekdwivedi3060 (changes requested):

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
shubhamyelikar99: Yes, that was to special-case null .Dropped it.

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
shubhamyelikar99: Agreed, applied your snippet. A cert-manager Secret has no meta.helm.sh/release-name, so it now regenerates instead of adopting material whose expiry and renewal aren't ours to manage. Release-owned reuse is unchanged, so helm upgrade still doesn't rotate the cert in the normal case.

Also added a case for a Secret owned by a different Helm release, which the same check covers.

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+{{- $svc := include "aerospike-kubernetes-operator.webhookServiceName" . -}}
+{{- list (printf "%s.%s.svc" $svc .Release.Namespace) (printf "%s.%s.svc.cluster.local" $svc .Release.Namespace) | toYaml -}}
+{{- end -}}
+
+{{/*
+Resolve the effective webhook certificate mode.
+
+`certs.webhook.provider` wins whenever it is set. `certs.webhook.create` is only
+consulted as a legacy fallback, and an explicit false is NOT an error — in 4.5.0 it meant
+"I create the Certificate named aerospike-operator-serving-cert myself, cainjector still
```
sidntrivedi, line 85: Will it be better to mention that the user creates the certificate instead of saying I. Can be confusing to read.

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+fills in the caBundle", not "no cert-manager". That maps onto none of the three
+user-facing providers, so it resolves to the internal `legacyCertManager` state, which
+renders exactly what 4.5.0 did. `legacyCertManager` is not a valid `provider` value and
+cannot be selected deliberately.
+
+Only a literal boolean false selects the legacy state. An absent or null `create` is not
```
sidntrivedi, line 91: I think we can mention what the user should do in case they were using `create: false` earlier. I feel the instruction is currently confusing.

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+
+Deliberately no Secret-ownership check here. Switching an existing release to selfSigned
+makes the chart the owner of a Secret that cert-manager created, and Helm refuses to adopt
+a resource it does not own unless the user passes --take-ownership. A template cannot see
+that flag (.Release exposes only IsInstall/IsUpgrade/Name/Namespace/Revision/Service), so
+any check here would either be wrong or would have to be waived by a values key: it would
+abort at render time, before Helm's own ownership validation, and so block the very
+--take-ownership path that migrates without an admission outage. Helm's error already names
+the Secret and every missing marker; the part it cannot know — that deleting the Secret
+alone lets cert-manager re-issue it — is documented in the README instead.
```
sidntrivedi, line 162: Lets update the comment with the approach we are using for checking ownership of the secret.

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -61,4 +61,129 @@ Deprecated fields are not allowed to be used with new charts, install/upgrade sh
...
+Whitespace is stripped from the external bundle: the value must reach the API server as
+single-line base64 (it decodes with strict base64), but `--set-file` from a file produced
+by `base64` carries a trailing newline, and GNU base64 wraps at 76 columns by default.
+*/}}
+{{- define "aerospike-kubernetes-operator.webhookCaBundle" -}}
+{{- $provider := include "aerospike-kubernetes-operator.webhookCertProvider" . -}}
+{{- if eq $provider "selfSigned" -}}
+{{- (include "aerospike-kubernetes-operator.webhookSelfSignedCert" . | fromYaml).ca -}}
+{{- else if eq $provider "external" -}}
+{{- .Values.certs.webhook.external.caBundle | nospace -}}
```
sidntrivedi, line 130: A whitespace-only `external.caBundle` renders a broken webhook config instead of failing since the `nospace` removes that space and returns `""`. Suggest tightening the pattern to require at least one base64 character in the schema validation. Can also add a case in `tests/webhook_certs_test.yaml` asserting a whitespace-only bundle fails.

helm-charts/aerospike-kubernetes-operator/README.md
```diff
@@ -23,9 +23,112 @@ Create the namespace where AKO will be installed. Replace the placeholder `<name
...
+
+#### provider: selfSigned (development and POC only)
+
+```sh
+helm install aerospike-kubernetes-operator aerospike/aerospike-kubernetes-operator \
+  --namespace <namespace> --set certs.webhook.provider=selfSigned
+```
+
+The chart generates the certificate and reuses the existing `Secret` on subsequent
+upgrades so that `helm upgrade` does not rotate it. Two caveats make this unsuitable for
```
sidntrivedi, line 59: nit: Two -> Three

helm-charts/aerospike-kubernetes-operator/values.yaml
```diff
@@ -38,12 +38,37 @@ metricsPort: 8443 # @schema required; minimum: 1; maximum: 65535
...
+    certManager: {} # @schema additionalProperties: false
+
+    # -- Reserved for future self-signed tuning (duration). Dev/POC only
+    selfSigned: {} # @schema additionalProperties: false
+
+    external: # @schema additionalProperties: false
+      # -- Base64-encoded PEM CA bundle that signed the webhook serving certificate
+      # (not the serving certificate itself). Required when provider is external.
+      # Surrounding and embedded whitespace is stripped before rendering, so wrapped
+      # output and a trailing newline are fine — `--set-file` from a `base64` file works.
```
sidntrivedi, line 69: `values.schema.json` ships two descriptions cut off mid-sentence: `--` anywhere in a *continuation* comment line truncates the description there.

With `useHelmDocs: true` in the helm schema config, the plugin treats `--` as a description start and keeps scanning for it past the first line. Backticks, quotes, and embedding in a word don't help; `x--y` and `a -- b` break it too.

**`values.yaml:69`** — `caBundle` ends at `"...so wrapped"` in the schema, losing the line about trailing newlines and wrapped output.
**`values.yaml:81`** — `logging` loses the rule that `development` must stay set (and `development` has no description of its own). Will send a patch in different PR or if you can fix in this PR itself.

```suggestion
      # output and a trailing newline are fine — reading it from a `base64` file works.
```

helm-charts/aerospike-kubernetes-operator/templates/_helpers.tpl
shubhamyelikar99: No — the cached value never reaches the release object. set .Values only mutates the in-memory values for that render pass; Helm snapshots the merged values before rendering and persists that snapshot.

Verified on a live cluster with a minimal chart reproducing this exact pattern:
~~~
 helm get values -n operators aerospike-kubernetes-operator      
USER-SUPPLIED VALUES:
certs:
  webhook:
    provider: selfSigned
watchNamespaces: aerospike
➜  aerospike-kubernetes-operator git:(feature/KO-605-generate-webhook-tls-cert-in-helm-chart) ✗ helm get values -n operators aerospike-kubernetes-operator --all
COMPUTED VALUES:
aerospikeKubernetesInitNameTag: aerospike-kubernetes-init:2.5.3
aerospikeKubernetesInitRegistry: docker.io
aerospikeKubernetesInitRegistryNamespace: aerospike
affinity: {}
annotations: {}
certs:
  metrics:
    create: false
    metricsServerCertSecretName: metrics-server-cert
  webhook:
    certManager: {}
    create: true
    external:
      caBundle: ""
    provider: selfSigned
    selfSigned: {}
    webhookServerCertSecretName: webhook-server-cert
extraEnv: {}
failedPodGracePeriodSeconds: "60"
fullnameOverride: ""
healthPort: 8081
ignoreNamespaces: kube-system,kube-node-lease
imagePullSecrets: []
labels: {}
livenessProbe:
  failureThreshold: 3
  initialDelaySeconds: 15
  periodSeconds: 20
  successThreshold: 1
  timeoutSeconds: 1
logging:
  development: true
  encoder: ""
  level: ""
metrics:
  secure: true
metricsPort: 8443
metricsService:
  annotations: {}
  labels: {}
  port: 8443
  type: ClusterIP
nameOverride: ""
nodeSelector: {}
operatorImage:
  pullPolicy: IfNotPresent
  repository: aerospike/aerospike-kubernetes-operator
  tag: 4.5.0
podAnnotations: {}
podLabels: {}
podSecurityContext: {}
rbac:
  create: true
  serviceAccountName: ""
readinessProbe:
  failureThreshold: 3
  initialDelaySeconds: 5
  periodSeconds: 10
  successThreshold: 1
  timeoutSeconds: 1
replicas: 2
resources:
  limits:
    cpu: 400m
    memory: 512Mi
  requests:
    cpu: 10m
    memory: 64Mi
safePodEviction:
  enable: false
  timeoutSeconds: "20"
securityContext:
  allowPrivilegeEscalation: false
tolerations: []
watchNamespaces: aerospike
webhookService:
  annotations: {}
  labels: {}
  port: 443
  targetPort: 9443
  type: ClusterIP
➜  aerospike-kubernetes-operator git:(feature/KO-605-generate-webhook-tls-cert-in-helm-chart) ✗ helm get values -n operators aerospike-kubernetes-operator | grep '_akoWebhookSelfSignedCert'     
➜  aerospike-kubernetes-operator git:(feature/KO-605-generate-webhook-tls-cert-in-helm-chart) ✗ 

~~~

helm-charts/aerospike-kubernetes-operator/tests/certs_test.yaml
shubhamyelikar99: it's an exact duplicate of line 163 (same set, same assert, only the name differs). The whitespace-tolerant half of the pattern is already covered by accepts wrapped base64 with a trailing newline below, so this one adds nothing. Removed.

helm-charts/aerospike-kubernetes-operator/tests/webhook_certs_test.yaml
shubhamyelikar99: removed

helm-charts/aerospike-kubernetes-operator/tests/webhook_certs_test.yaml
shubhamyelikar99: The comment explains why these tests pin exact webhook counts.

caBundle is set per webhook entry, not once per configuration — each entry has its own {{- with $caBundle }} block. Add a new entry without it and that webhook silently fails TLS. The counts catch that.

Dropped the "gets forgotten" line — it was an opinion, not a reason.

helm-charts/aerospike-kubernetes-operator/tests/webhook_selfsigned_lookup_test.yaml
shubhamyelikar99: that test is gone

helm-charts/aerospike-kubernetes-operator/values.yaml
shubhamyelikar99: Confirmed, and it's a bit worse than truncation — the line containing -- is dropped entirely, and the text closes up around the hole. For caBundle that left "so wrapped" joined to the next sentence, so it didn't even look cut off.

Reproduced it in a minimal chart: -- anywhere in a continuation line kills that line, backticks don't help, and -- on the opening # -- line is fine (which is why --zap-encoder survives).

Reworded both lines to avoid -- and regenerated the schema — descriptions are complete now. Also checked the rest of values.yaml; those two were the only ones.

helm-charts/aerospike-kubernetes-operator/README.md
shubhamyelikar99: Yes, it's genuine, and there's no chart-side fix — Sprig's genSignedCert sets NotBefore to the render time with no backdating (I checked: notBefore matches the local clock to the second), and it exposes no way to shift it.

There is a way to warn about this at install time. I tried it on a live cluster and it works.

lookup can read a node Lease from kube-node-lease. Its renewTime is written by kubelet a few seconds ago, so comparing it with now tells us the difference between the local clock and the cluster clock:


{{- $leases := lookup "coordination.k8s.io/v1" "Lease" "kube-node-lease" "" }}
{{- if and $leases $leases.items }}
{{- $renew := (first $leases.items).spec.renewTime }}
{{- $clusterEpoch := regexReplaceAll "\\.[0-9]+Z$" $renew "Z" | toDate "2006-01-02T15:04:05Z07:00" | unixEpoch }}
{{- $skew := sub (now | unixEpoch) $clusterEpoch }}
{{- if gt $skew 60 }}
WARNING: this machine's clock is {{ $skew }}s ahead of the cluster. The generated webhook
certificate will not be valid yet, and admission will fail until the difference passes.
{{- end }}
{{- end }}
On our dev cluster it reports 7 seconds, which is just the normal lease delay, so it stays quiet.

A few things to decide first:

lookup returns nothing with helm template and client-side --dry-run, so the check is skipped in the GitOps case. It works on real installs and with --dry-run=server.
The user running helm needs read access to leases in kube-node-lease. A restricted CI account may not have it, so the check must stay quiet when the lookup is empty.
Node leases refresh every ~10s, so the local clock always looks slightly ahead. A 60s threshold avoids false alarms.
Helm has no "warn" option, only fail, which stops the install. I would not fail on a rough check like this. Better to print it from NOTES.txt and only when provider: selfSigned.

But since this "selfSigned" mode is only for dev/poc setups i think we should not add this complexity.

PR #599, opened 2026-09-14
feat: Add support for index-checkpoint feature [KO-585]

abhishekdwivedi3060:
Adds AKO support for the Aerospike server's **index-checkpoint** preview feature
(SERVER-971), so a pod replacement no longer costs a cold start.

## Why

The primary index lives in SysV shared memory, which is owned by the IPC namespace rather
than the `asd` process. It therefore survives an `asd` restart — that is how warm restart
works today — but **not** pod replacement, which destroys the IPC namespace.

Without a checkpoint, every pod replacement means:

- **device/shadow-backed namespace** — the index is rebuilt by scanning the whole device,
  which is the dominant cost of a rolling restart on a large namespace;
- **shadowless in-memory namespace** — the data itself is gone, and at RF=1 permanently.

Index checkpoint closes that gap: before a pod is deleted, AKO asks the server to copy its
volatile shmem state (primary index, secondary index, and for a shadowless in-memory
namespace its data stripes) to a durable directory. The replacement pod hydrates from it
automatically at startup.

## User-facing API

```yaml
spec:
  previewFeatures:
    - index-checkpoint                 # becomes --preview index-checkpoint on the server
  aerospikeConfig:
    service:
      index-checkpoint-path: /mnt/index-ckpt    # cluster-wide, PVC-backed
    namespaces:
      - name: test
        skip-checkpoint: false                  # dynamic; opt this namespace out
        index-checkpoint-threads: 16            # 1..16
        index-checkpoint-compression: true
```

`spec.previewFeatures` is a new CR field; the three `index-checkpoint-*` keys and
`skip-checkpoint` come from the server's config schema.

## How it works

Once per batch, **before any pod in it is touched**:

1. **`splitCheckpointNamespaces`** partitions the rack's namespaces into *eligible* and
   *skip*. A namespace is skipped when the replacement pod could not usably read its
   checkpoint: the cluster-wide path moved, the namespace left the CR or set
   `skip-checkpoint`, or its `data-size` changed (the checkpoint would be written at a
   geometry the replacement pod rejects).
2. **`skipCheckpointForNamespaces`** opts those out via dynamic
   `set-config … skip-checkpoint=true`. This must precede the save — once the save fires,
   the server's info gate refuses `set-config` for the whole park.
3. **`triggerIndexCheckpointSave`** sends `checkpoint-save` to every pod in the batch. The
   server copies, then **parks**: it leaves the cluster, shuts storage down, and holds so
   an orchestrator can confirm the copy finished. AKO records the park on the pod
   (`IndexCheckpointParkedAnnotation`, keyed by container ID).
4. **`pollAndDeleteParkedPods`** polls `checkpoint-status` and deletes each pod **as its
   own save completes**, so a slow sibling does not burn a fast finisher's park budget.
5. The replacement pod hydrates from the checkpoint at startup — no operator flag, the
   server decides from observable state.

### The discharge phase

`reconcileCheckpointingPods` runs **cluster-wide, before the racks reconcile**, and is the
single owner of finishing a park.

This is the central design point. A parked node is an **obligation, not a divergence**:
`checkpoint-save` is a point of no return — the node has left the cluster and shut its
storage down, so it cannot serve again without a restart — and **no comparison of spec
against status will say so**. After a reverted upgrade, `desired == actual` for a parked
pod and it still demands action. Keying the discharge on the pod annotation rather than on
spec-vs-status is what makes that correct, and running it first means no later phase
(safe-stop checks, scale-down, batching) ever meets a parked node.

## Notable behaviours

- **Parked pods are excluded from cluster info calls.** A parked node answers only the two
  checkpoint commands, so including it would fail every caller.
- **No checkpoint on failure recovery.** A failed pod's server is down and its info port
  unreachable; sending `checkpoint-save` there would abort the batch.
- **No checkpoint on scale-down.** `scaleDownRack` already quiesces and waits for
  migration, so there is no data-safety gap a checkpoint would close.
- **A parked pod reports Ready, deliberately.** Smart clients route by partition map and
  the node left the cluster at trigger time, so no client sends it work. Making the probe
  fail would block other racks' preambles and consume PodDisruptionBudget.
- **Cleanup on a path change.** When `index-checkpoint-path` moves, AKO sweeps the old
  path — the server only ever sweeps beneath its *current* path, so nothing else can.

## Validation (webhook)

- All four config keys are rejected without `previewFeatures: [index-checkpoint]`, and the
  feature requires server **≥ 8.2.0.0**.
- `index-checkpoint-path` must be backed by a **Filesystem-mode PersistentVolume** — an
  `emptyDir`, `hostPath`, Block-mode PV, or no backing volume at all is rejected. This is
  what keeps the checkpoint off a container's ephemeral writable layer, which the server
  would otherwise accept and silently lose.
- When `index-checkpoint-path` is set, **no namespace in the cluster** may be named `.`,
  `..`, `/`, `*.tmp` or `*.deleting` — these collide with the server's on-disk layout and
  crash `asd` at startup. Applies to every namespace, including opted-out ones.

## Tests

- **Unit** — `splitCheckpointNamespaces` (data-size change, gained storage backing, CR
  removal, `skip-checkpoint`, path move, feature off), `checkpointDone`,
  `markPodCheckpointParked` retry, `IsPodCheckpointing`, and the `api/v1` accessors.
- **Envtest** — preview gate (all four keys, per rack), version gate, thread bounds, the
  namespace-name guard, and all three `validateIndexCheckpointVolume` rejection paths.
- **Integration** (`test/cluster/index_checkpoint_test.go`) — a chained lifecycle spec
  (warm restart, podRestart→hydrate, `data-size` skip, `skip-checkpoint`, path change,
  scale-down) plus standalone cases for image upgrade, multi-pod batch, failed-pod
  recovery, spec-revert-while-parked, namespace removal, and a node that parks having
  saved nothing.

## Dependencies

- **aerospike-management-lib** bumped to `v1.11.2-0.20260911132439-d14f74b8d7ea` for
  `CheckpointSave` / `CheckpointStatus`, the response classifier and the status parser.
- **configschema submodule** bumped to pick up the published `8.2.0.json` schema.

## Known limitations

- The feature is **preview-gated** server-side and only available on server ≥ 8.2.0.0.
- A `data-size` change on a shadowless in-memory namespace means the checkpoint is skipped;
  at RF=1 the namespace comes up empty. Documented, not averted — the outcome is not caused
  by this feature, which merely cannot prevent it.
- A retained checkpoint directory for a namespace that stopped checkpointing is not
  reclaimed by AKO — reaping stays with the server so the delete decision is not split
  across two actors. Tracked for GA.


tanmayja:
Do we know if server internally handles these tyoe of configs where cold restart is enforced. eg. https://aerospike.com/docs/database/reference/config#namespace__partition-tree-sprigs

api/v1/index_checkpoint.go
```diff
@@ -0,0 +1,85 @@
...
+	skip, _ := namespaceConf[ConfKeySkipCheckpoint].(bool)
+
+	return skip
+}
+
+// IsCheckpointUsableNamespaceName mirrors the server's namespace-name rule. It builds
+// "<path>/<name>" plus ".tmp" (save staging) and ".deleting" (go-live delete in
+// progress) siblings, so these names escape the checkpoint directory or collide with a
+// sibling's, and it crashes at startup on them.
+func IsCheckpointUsableNamespaceName(name string) bool {
```
tanmayja: this means, there is a possibility that namespace names could exist in cluster which does not follow these rules. but checkpointing won't be allowed on those?

api/v1/aerospikecluster_types.go
```diff
@@ -1154,6 +1161,11 @@ type AerospikeClusterStatusSpec struct { //nolint:govet // for readability
 	// Operations is a list of on-demand operation to be performed on the Aerospike cluster.
 	// +optional
 	Operations []OperationSpec `json:"operations,omitempty"`
+
+	// PreviewFeatures is the list of Aerospike server preview feature names the running
+	// cluster was started with, via the --preview flag.
```
shubhamyelikar99: should we keep the same description as in spec field?
~~~
// PreviewFeatures is a list of Aerospike server preview feature names to enable via the
// --preview startup flag. Features gated behind this flag (e.g. "index-checkpoint")
// will cause the server to crash at startup if the flag is not present.
~~~

api/v1/utils.go
```diff
@@ -760,3 +799,52 @@ func GetClientAuthMode(authMode AerospikeAuthMode) as.AuthMode {
...
+	if err != nil {
+		return 0, false
+	}
+
+	return size, true
+}
+
+// GetInMemoryNsDataSizes maps namespace name to storage-engine data-size, for
+// pure in-memory namespaces only
+func GetInMemoryNsDataSizes(aerospikeConfig map[string]interface{}) map[string]int {
```
shubhamyelikar99, line 824: Comment says "for pure in-memory namespaces only", but it returns storage-engine.data-size for any namespace that has the key — there is no type == memory check.

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+	ctx context.Context,
+	pods []*corev1.Pod, rackState *RackState,
+) common.ReconcileResult {
+	rackStatus := r.getRackStatus(rackState)
+	if rackStatus == nil {
+		return common.ReconcileSuccess()
+	}
+
+	eligibleNSs, skipNSs := r.splitCheckpointNamespaces(rackStatus, rackState.Rack)
+	if len(eligibleNSs) == 0 || len(pods) == 0 {
```
shubhamyelikar99: Isn't `len(pods) == 0 ` should be at the top of function? if len(pods) is 0 then `r.getRackStatus(rackState)` call will be wasted

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+					"Could not set skip-checkpoint, namespace will be checkpointed needlessly",
+					"pod", utils.GetNamespacedName(pod), "namespace", ns, "response", respVal)
+
+				continue
+			}
+
+			// Why each namespace was excluded is a per-rack decision, logged once by
+			// the caller; this line is per pod and only confirms the opt-out landed.
+			r.Log.Info("Excluded namespace from the imminent index checkpoint save",
+				"pod", pod.Name, "namespace", ns)
```
shubhamyelikar99:
```suggestion
				"pod", utils.GetNamespacedName(pod), "namespace", ns)
```

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+	var anyParked bool
+
+	// No checkpoint-status precheck: the server answers a re-issue idempotently so an already-saving pod
+	// classifies as CheckpointSaveAccepted below
+	for _, pod := range pods {
+		verdict, err := r.newAsConn(pod).CheckpointSave(policy, checkpointParkTimeoutSec)
+		if err != nil {
+			// Network-level failure (connection error, timeout, EOF).
+			return common.ReconcileError(
+				fmt.Errorf("checkpoint-save on pod %s: %w", pod.Name, err),
```
shubhamyelikar99:
```suggestion
				fmt.Errorf("checkpoint-save on pod %s: %w", utils.GetNamespacedNameString(pod), err),
```

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+			)
+		}
+
+		switch verdict {
+		case deployment.CheckpointSaveRejected:
+			return common.ReconcileError(fmt.Errorf(
+				"checkpoint-save rejected by pod %s", utils.GetNamespacedName(pod)))
```
shubhamyelikar99:
```suggestion
				"checkpoint-save rejected by pod %s", utils.GetNamespacedNameString(pod)))
```
Please in other places as well. As per logging guideline https://aerospike.atlassian.net/wiki/spaces/KO/pages/5480677440/Contributing+Guidelines+Logging+Error+Handling+Event+Formatting#1.3-Attach-object-identity-with-helpers%2C-not-hand-built-strings

utils.GetNamespacedName(obj) -> for logging
utils.GetNamespacedNameString(obj) -> for error,events

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+				// disagrees with the rack status AKO derived its namespace set from.
+				// Retrying cannot change that, and there is no park to discharge.
+				r.Log.Info("Server reports index checkpoint is not configured, leaving Pod to the normal flow",
+					"pod", utils.GetNamespacedName(pod))
+
+				continue
+
+			case err != nil:
+				r.Log.V(1).Info("checkpoint-status unavailable, will retry",
+					"pod", utils.GetNamespacedName(pod), "error", err)
```
shubhamyelikar99: shouldn't this should be error level?

shubhamyelikar99 (changes requested):

internal/controller/cluster/aero_info_calls.go
abhishekdwivedi3060: This is a retry loop, so not logging it as error. Although moved it to Info for consistency with other switch conditions

api/v1/utils.go
abhishekdwivedi3060: data-size is only valid for pure in-memory clusters. With other types, schema validation would fail

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1438,6 +1446,112 @@ func cgroupMemTrackingWarning(version string, conf map[string]interface{}) admis
...
+	for _, name := range cluster.Spec.PreviewFeatures {
+		info, known := asdbv1.PreviewFeatureVersions[name]
+		if !known {
+			knownNames := make([]string, 0, len(asdbv1.PreviewFeatureVersions))
+			for k := range asdbv1.PreviewFeatureVersions {
+				knownNames = append(knownNames, k)
+			}
+
+			return warnings, fmt.Errorf(
+				"unknown preview feature %q; known features: %v",
```
shubhamyelikar99, line 1466:
```suggestion
				"unknown preview feature %q; known features: %w",
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1438,6 +1446,112 @@ func cgroupMemTrackingWarning(version string, conf map[string]interface{}) admis
...
+		if val < 0 {
+			return warnings, fmt.Errorf(
+				"preview feature %q requires server version >= %s, but image version is %s",
+				name, info.MinVersion, version,
+			)
+		}
+
+		if info.GAVersion != "" {
+			gaVal, err := lib.CompareVersions(version, info.GAVersion)
+			if err == nil && gaVal >= 0 {
```
shubhamyelikar99: err is unhandled

test/testutil/defaults.go
```diff
@@ -38,6 +41,11 @@ var (
 	Pre810FederalImage    = fmt.Sprintf("%s:%s", BaseFederalImage, Pre810ServerVersion)
 	LatestFederalImage    = fmt.Sprintf("%s:%s", BaseFederalImage, LatestServerVersion)
 	InvalidImage          = fmt.Sprintf("%s:%s", BaseEnterpriseImage, InvalidImageVersion)
+	// TODO: update this later
+	IndexCheckpointImage = "aerospike.jfrog.io/database-docker-test-local/aerospike-server-enterprise:" +
+		"8.2.0.0_20260911T220300Z"
+	// IndexCheckpointUpgradeImage is the image the index-checkpoint upgrade case upgrades to
+	IndexCheckpointUpgradeImage = IndexCheckpointImage
```
shubhamyelikar99, line 48: this TODO needs to be handled before merging this PR?

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+			case len(resp.Namespaces) == 0 && !resp.IsParked:
+				// Configured, checkpointing nothing, and NOT parked: a RESTARTED asd.
+				// Nothing to wait for and nothing to delete; the annotation check
+				// reclassifies it next pass. The PARKED form of an empty response falls
+				// through instead — see checkpointDone.
+				r.Log.Info("Server reports no checkpointing namespace and no park, leaving Pod to the normal flow",
+					"pod", utils.GetNamespacedName(pod))
+
+				continue
+			}
```
shubhamyelikar99, line 861: **`pollAndDeleteParkedPods`: a parked pod can reach a state with no exit**

Both `continue` branches — `ErrCheckpointNotConfigured` (L823) and `len(resp.Namespaces)==0 && !resp.IsParked` (L839) — drop the pod without deleting it and without clearing `IndexCheckpointParkedAnnotation`. Nothing else clears it either: written at L667, removed nowhere. A park only ends via pod deletion or a container restart.

So if the container ID hasn't changed, `IsPodCheckpointing` still matches on the next pass: `reconcileCheckpointingPods` re-selects the pod, sleeps 10s, drops it again, and returns `ReconcileRequeueAfter(1)` (L221). Since that's the first thing `reconcileRacks` does (rack.go L67), **no rack reconciles at all** — every ~11s, indefinitely, visible only as one Info line per cycle.

The same stale annotation also excludes the pod from every cluster info call (L348) permanently.

I couldn't construct a trigger on the normal path — `CheckpointSaveNothingToDo` deliberately doesn't annotate — so this reads as defence-in-depth rather than a live bug. But the exit currently depends on a server-side invariant (`accepted ⇒ configured`) rather than on anything this function controls.

Should we patch the annotation off before each `continue`?. Fixes both the loop and the info-call exclusion.

Question: can `CheckpointSaveAccepted` ever be followed by `ErrCheckpointNotConfigured`? If you can rule it out server-side, L823 is probably better as an error return than a `continue` — surface the impossible state rather than loop on it.

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+			// Running, finished, or failed — the discharge phase distinguishes them, and a
+			// failed save gets its IndexCheckpointFailed event from there with the
+			// namespace list.
+			r.Log.Info("Index checkpoint save already accepted",
+				"pod", utils.GetNamespacedName(pod))
+			r.markPodCheckpointParked(ctx, pod)
+
+			anyParked = true
+
+		case deployment.CheckpointSaveNothingToDo:
```
shubhamyelikar99, line 617: **Approach doc §6 says this path emits an event — I don't see one here**

§6 ("What an operator sees") has:

> | The running server is checkpointing nothing | AKO proceeds without a checkpoint and emits `IndexCheckpointFailed` — retrying cannot change the answer. |

Here it's `r.Log.Info` only. The row directly above it in the same table — `state=failed` → `IndexCheckpointFailed` — is emitted at L862-868, so the two cases the doc treats alike behave differently.

The comment on this branch also reads "say so **loudly**, since a shadowless memory namespace loses its data this way", which sounds like an event was intended.

Noting it because of where this lands: `NothingToDo` doesn't annotate, `anyParked` stays false, and if the whole batch answers this way L639 returns success and the pods are deleted in the same pass. At RF=1 with a shadowless in-memory namespace that's permanent data loss, and the only trace is one Info line — nothing on `kubectl describe aerospikecluster`.

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+
+// markPodCheckpointParked records on the pod that its Aerospike node is checkpointing, so later
+// reconciles can tell it is out of the cluster without asking the server — which a parked node would refuse anyway.
+// Written only after checkpoint-save has been accepted, so the annotation can never claim
+// a park that did not happen. The reverse — a park with no annotation — is possible if the
+// patch cannot be written at all, and degrades to the pre-annotation behaviour: the node is
+// treated as a cluster member, so cluster info calls hit it and fail until its park times
+// out and it restarts. Failing the reconcile here would fix neither, so it is retried, then
+// reported as precisely as we can and left to self-heal.
+func (r *SingleClusterReconciler) markPodCheckpointParked(ctx context.Context, pod *corev1.Pod) {
```
shubhamyelikar99, line 654: **`markPodCheckpointParked`: the two early returns that lose a park have different severities**

By the time this runs, `checkpoint-save` has already been accepted, so the node is parked either way. Two of the three exits leave no annotation:

- L652 (`containerID == ""`) → `Log.Info`
- L681 (patch failed) → `Log.Error`

(L674, pod already deleted, is genuinely benign.)

Both produce the same end state, and the L681 message describes it exactly: *"This node has left the cluster but AKO cannot tell, so cluster operations will fail until its park times out and the container restarts."*

The caller sets `anyParked = true` regardless (L611/L627), since the function returns nothing. So next pass `IsPodCheckpointing` is false, the discharge phase skips the pod, and the rack flow runs quiesce/cluster-stable over a host-conn list that still includes the parked node — L348 keys off the same annotation. Those info calls fail until the 300s timeout expires. The operator sees quiesce errors with the Error-level breadcrumb missing.

The doc comment at L645-649 says the policy is *"retried, then reported as precisely as we can"* — L652 does neither.

Is L652 considered unreachable? We'd just completed an info call to that pod's server, so the container is running; a stale cache read seems like the only route. And pkg/utils/pod.go:200 reads "An unreadable ID means a restart is in flight, which is not a park" — if that's the assumption, it might be worth stating it here rather than logging past it.

internal/controller/cluster/aero_info_calls.go
```diff
@@ -438,3 +537,388 @@ func (r *SingleClusterReconciler) setDynamicConfig(
...
+
+	policy := r.getClientPolicy(ctx)
+	pending := make([]*corev1.Pod, len(pods))
+	copy(pending, pods)
+
+	for i := 0; i < maxRetry && len(pending) > 0; i++ {
+		r.Log.V(1).Info("Waiting for index checkpoints to complete",
+			"pods", getPodNames(pending), "attempt", i+1)
+
+		time.Sleep(retryInterval)
```
shubhamyelikar99, line 825: **Minor: the poll loop sleeps before its first `checkpoint-status`**

L815 runs `time.Sleep(10s)` at the top of the loop body, before the first `CheckpointStatus` at L820. Combined with the `ReconcileRequeueAfter(1)` that `triggerIndexCheckpointSave` returns at L636, that puts a ~11s floor on discharging a batch even when the copy finished in well under a second.

It compounds across a rolling restart — one floor per batch, per rack.

Moving the sleep to the end of the iteration (or checking status first, then sleeping only if anything is still pending) would cost nothing in the slow case and skip it entirely in the fast one.

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -1438,6 +1446,112 @@ func cgroupMemTrackingWarning(version string, conf map[string]interface{}) admis
 	return nil
 }
 
+// validatePreviewFeatures validates spec.previewFeatures: rejects unknown names, checks
+// server version compatibility, cross-checks with aerospikeConfig fields that require opt-in,
+// and warns when a feature has graduated to GA.
+func validatePreviewFeatures(cluster *asdbv1.AerospikeCluster, version string) (admission.Warnings, error) {
```
shubhamyelikar99, line 1452: **`previewFeatures` isn't deduped or normalised**

`validatePreviewFeatures` (webhook L1452) checks each name against `PreviewFeatureVersions` but doesn't look at the list as a whole, so `["index-checkpoint", "index-checkpoint"]` passes and `previewFeaturesArgs` (statefulset.go L1395) joins it straight through to `--preview index-checkpoint,index-checkpoint`.

The ordering case is probably the more interesting one once this field holds more than one feature: `["a","b"]` and `["b","a"]` produce different `Args`, and `isPreviewFeaturesUpdated` (pod.go L2162) compares with `slices.Equal`, so reordering the list in the CR would trigger a full rolling restart for no actual change.

Sorting + deduping in the mutating webhook would settle both, and would keep `spec` and `status.previewFeatures` comparable too.

shubhamyelikar99 (changes requested):

internal/webhook/v1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: Here we are not wrapping the error, instead creating one with fmt.Errorf

internal/webhook/v1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: GAVersion is set by us in the code so err != nil flow is dead. But added a check for consistency

test/testutil/defaults.go
abhishekdwivedi3060: Yes, this is meant only for testing. Will be removed once we've an official 8.2.0 server release

internal/controller/cluster/aero_info_calls.go
abhishekdwivedi3060: This is intentional.
The rationale is that checkpoint-save won't finish within a second or 10s in prod setups and checkpoint-save immediately requeus after 1 second. So sleeping at the end would be making redundant asinfo calls to server in the first iteration always where we know that the checkpointing won't complete.

internal/controller/cluster/aero_info_calls.go
abhishekdwivedi3060: If the annotation is present then
- ErrCheckpointNotConfigured (L823) is unreachable because checkpoint-path removal requires warm restart but AKO won't change it before processing checkpointing pod
-  len(resp.Namespaces)==0 && !resp.IsParked this is possible for namespaces which as nothing to checkpoint (all flash) but in the process of checkpointing. Subsequent loops or reconcile will recover from this.

We can add a annotation removal flow but it doesn't seem to help much here. Stale annotations with old containersID are automatically cleared in next restart

internal/controller/cluster/aero_info_calls.go
abhishekdwivedi3060: The pod object used to trigger checkpoint save is used to get the containerID, so  (containerID == "")  is unreachable in a normal flow and fail safe check.

Patch error is retired multiple time with backoff to handle transient errors.

internal/controller/cluster/aero_info_calls.go
abhishekdwivedi3060: If server is reporting NothingToDo due to scenarios like All flash, thats genuine and not a reason to emit an event.
Eg of RF=1, shadowless in-memory namespace and server reporting NothingToDo is a config issue.

shubhamyelikar99 (dismissed):

PR #605, opened 2026-09-22
Feat: Enable parallel quiescing of scale down pods across racks. [KO-530]

tanmayja:
Approach doc: https://aerospike.atlassian.net/wiki/spaces/KO/pages/5441093655/Cross-Rack+Parallel+Quiesce+for+Scale-Down+Approach+Doc


internal/controller/cluster/testhelpers_test.go
```diff
@@ -132,6 +132,49 @@ func getCluster(t *testing.T, c client.Client, ac *asdbv1.AerospikeCluster) *asd
...
+
+	return &SingleClusterReconciler{
+		Client:      fakeClient,
+		Log:         logr.Discard(),
+		Scheme:      scheme,
+		aeroCluster: aeroCluster,
+		Recorder:    record.NewFakeRecorder(10),
+	}
+}
+
```
abhishekdwivedi3060: Below func newTestReconciler already have same functionality

internal/controller/cluster/rack.go
```diff
@@ -139,23 +123,77 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
+	)
+
+	if batchQuiesceEnabled {
+		// Step 1: collect targets; undo stale quiesces on non-targets.
+		// Readiness checks deferred to Step 3 (classifyTargetPods inside reconcileBatchQuiesce).
+		if allTargets, res = r.buildScaleDownTargets(ctx, scaledDownRacks, racksToDelete); !res.IsSuccess {
+			return res
+		}
+
+		targetNames = sets.New(getPodNames(allTargets)...)
```
abhishekdwivedi3060: Use existing func podNamesToSet(allTargets)

internal/controller/cluster/rack.go
```diff
@@ -139,23 +123,77 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
+			if err := r.setConditions(ctx, metav1.Condition{
+				Type:    string(asdbv1.AerospikeClusterConditionScalingDown),
+				Status:  metav1.ConditionTrue,
+				Reason:  asdbv1.AerospikeClusterReasonScalingDown,
+				Message: fmt.Sprintf("Pre-quiescing %d pod(s) across racks before scale-down", len(allTargets)),
+			}); err != nil {
+				return common.ReconcileError(err)
 			}
 		}
+
```
abhishekdwivedi3060: Shound't this be called below below just before reconcileBatchQuiesce?
Claiming the ScalingDown here would be wrong as we are not starting the scale down here and the below nonScaleDownRacks flow would also show ScaleDown True

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+			annotatedNonTargets = append(annotatedNonTargets, *pod)
+		}
+	}
+
+	if len(annotatedNonTargets) == 0 {
+		// Nothing for AKO to undo — skip all Aerospike info calls.
+		return common.ReconcileSuccess()
+	}
+
+	r.Log.Info("Sending quiesce-undo to annotated non-target pods",
```
abhishekdwivedi3060: Can we change the msg framing "non-target" pods? 
As a user, I won't get much info from the logs like what is a non-target pod. Can we make it more self-explanatory like  ""Sending quiesce-undo to quiesced(annotated) pods which are not scale down candidates".

Check it wherever applicable

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := deployment.InfoQuiesceUndoSubset(r.Log, policy, nonTargetHostConns, allHostConns); err != nil {
+		return common.ReconcileError(fmt.Errorf("send quiesce-undo to non-target pods: %w", err))
+	}
+
+	// Clear annotations from unquiesced non-target pods only; targets keep
+	// theirs for the fast-exit check in reconcileBatchQuiesce.
+	for i := range annotatedNonTargets {
+		if annErr := r.setPodQuiesceAnnotation(ctx, &annotatedNonTargets[i], false); annErr != nil {
+			r.Log.Error(annErr, "Failed to remove quiesce annotation from pod; will retry next reconcile",
+				"pod", annotatedNonTargets[i].Name)
```
abhishekdwivedi3060: Follow the new logging convention:
1. Resource name must be plural in logs -> Pod
2. GetNamespacedName instead of just name

internal/controller/cluster/rack.go
```diff
@@ -139,23 +123,77 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
+	// Step 2: reconcile non-scaled-down racks (always, regardless of flag).
+	for idx := range nonScaledDownRacks {
+		rack := &nonScaledDownRacks[idx]
+		if res = r.reconcileRack(ctx, rack.rackSTS, rack.rackState, ignorablePodNames, nil); !res.IsSuccess {
+			return res
+		}
+	}
+
+	if batchQuiesceEnabled {
+		// Step 2.5/3: quiesce all targets at once — single concurrent migration round.
```
abhishekdwivedi3060: Fix the log -  2.5/3??

internal/controller/cluster/rack.go
```diff
@@ -139,23 +123,77 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
+		// Step 1: collect targets; undo stale quiesces on non-targets.
+		// Readiness checks deferred to Step 3 (classifyTargetPods inside reconcileBatchQuiesce).
+		if allTargets, res = r.buildScaleDownTargets(ctx, scaledDownRacks, racksToDelete); !res.IsSuccess {
+			return res
+		}
+
+		targetNames = sets.New(getPodNames(allTargets)...)
+
+		// Claim ScalingDown for this reconcile pass now, before any requeue that might
+		// happen in Steps 2.5/2.75/3. Without this, a requeue before Step 4
```
abhishekdwivedi3060: 2.5/2.75/3.???

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+
+	return common.ReconcileSuccess()
+}
+
+// reconcileBatchQuiesce quiesces ALL scale-down targets across every rack in a
+// single pre-pass, triggering one concurrent migration round instead of N
+// sequential rounds. Runs after classifyTargetPods and before
```
abhishekdwivedi3060: Fix comment "Runs after classifyTargetPods and before
// reconcileRack for scaled-down racks."

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+	ctx context.Context,
+	allTargets []*corev1.Pod,
+	ignorablePodNames sets.Set[string],
+) common.ReconcileResult {
+	if len(allTargets) == 0 {
+		return common.ReconcileSuccess()
+	}
+
+	// Filter ignorable targets; check annotation fast-exit.
+	effectiveTargets := allTargets[:0]
```
abhishekdwivedi3060: Any reason for doing like this instead of creating a new slice? For optimisation?
This approach would override the underlying array for passed allTargets slice. Safe for now but can be breaking if other funcs rely on this list

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+		r.Log.V(1).Info("All scale-down targets already quiesced by AKO, skipping batch quiesce pre-pass",
+			"targetCount", len(allTargets))
+
+		return common.ReconcileSuccess()
+	}
+
+	r.Log.Info("Running cross-rack batch quiesce pre-pass", "targetCount", len(allTargets))
+
+	// Local copy: mutations by classifyTargetPods must not bleed into Step 4's reconcileRack.
+	localIgnorable := ignorablePodNames.Union(sets.New[string]())
```
abhishekdwivedi3060:
```suggestion
	localIgnorable := ignorablePodNames.Clone()
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -305,6 +305,11 @@ func (acv *AerospikeClusterCustomValidator) ValidateUpdate(_ context.Context,
 		return warnings, err
 	}
 
+	// Validate EnableParallelScaleDownAcrossRacks toggle restrictions.
+	if err := validateBatchScaleDownQuiesceToggle(oldObject, aerospikeCluster); err != nil {
```
abhishekdwivedi3060: Rename this and comments as per new field name

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+	ctx context.Context,
+	targetNames sets.Set[string],
+	ignorablePodNames sets.Set[string],
+) common.ReconcileResult {
+	podList, err := r.getClusterPodList(ctx)
+	if err != nil {
+		return common.ReconcileError(fmt.Errorf("list cluster pods for batch quiesce target classification: %w", err))
+	}
+
+	for idx := range podList.Items {
```
abhishekdwivedi3060: I think only checking the targetPod list should be sufficient. No need to check all the pods.
Also, see if we really need this func? Maybe this logic can be moved to the caller func where we are doing the filtering

internal/controller/cluster/rack.go
```diff
@@ -139,23 +123,77 @@ func (r *SingleClusterReconciler) reconcileRacks(ctx context.Context) common.Rec
...
+
+	// Cross-rack parallel quiesce pre-pass (opt-in via EnableParallelScaleDownAcrossRacks).
+	// The webhook prevents disabling the flag mid-scale-down, so a disabled flag
+	// can never leave stale BatchQuiesceAnnotation pods requiring cleanup.
+	//
+	// Step ordering when enabled (Steps 2/4/deleteRacks run unconditionally):
+	//   Step 1   — buildScaleDownTargets + reconcileQuiesceUndo
+	//   Step 2   — reconcileRack for non-scaled-down racks
+	//   Step 3   — reconcileBatchQuiesce (classifyTargetPods + waitForMultipleNodesSafeStopReady)
+	batchQuiesceEnabled := ptr.Deref(r.aeroCluster.Spec.RackConfig.EnableParallelScaleDownAcrossRacks, false)
```
abhishekdwivedi3060: Rename this as per new field name. Check wherever applicable

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	r.Log.Info("Running cross-rack batch quiesce pre-pass", "targetCount", len(allTargets))
+
+	// Local copy: mutations by classifyTargetPods must not bleed into Step 4's reconcileRack.
+	localIgnorable := ignorablePodNames.Union(sets.New[string]())
+
+	// Never-joined targets → localIgnorable; previously-joined non-running → ReconcileError.
+	targetNames := sets.New(getPodNames(allTargets)...)
+	if res := r.classifyTargetPods(ctx, targetNames, localIgnorable); !res.IsSuccess {
```
abhishekdwivedi3060: All the pending pods are added to the localIgnorable set but they are never removed from the targetNames list and below loop adds the annotations on those pods well.

internal/controller/cluster/batch_quiesce.go
```diff
@@ -0,0 +1,302 @@
...
+
+	if res := r.waitForMultipleNodesSafeStopReady(ctx, allTargets, localIgnorable, 0, true); !res.IsSuccess {
+		return res
+	}
+
+	// Stamp annotation for fast-exit; failure is non-fatal (next reconcile retries).
+	for _, pod := range allTargets {
+		if pod.Annotations[asdbv1.BatchQuiesceAnnotation] != asdbv1.BatchQuiesceAnnotationValue {
+			if annErr := r.setPodQuiesceAnnotation(ctx, pod, true); annErr != nil {
+				r.Log.Error(annErr, "Failed to set quiesce annotation on pod; will re-quiesce next reconcile",
```
abhishekdwivedi3060: nit: 
```suggestion
				r.Log.Error(annErr, "Failed to set quiesce annotation on pod; will try quiesce next reconcile",
```

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -843,6 +848,74 @@ func validateForceBlockFromRosterUpdate(newObj *asdbv1.AerospikeCluster) error {
...
+
+// scaleDownInFlight reports whether a scale-down (or rack deletion/replacement)
+// is currently in progress, meaning pods may already carry BatchQuiesceAnnotation.
+//
+// Two checks are performed in two passes:
+//  1. Build a map of spec rack ID → effective pod count from the spec topology.
+//  2. Sweep status racks: a status rack missing from spec (deletion/replacement)
+//     or whose effective count exceeds the spec count (per-rack shrink) indicates
+//     an in-flight operation.
+func scaleDownInFlight(obj *asdbv1.AerospikeCluster) bool {
```
abhishekdwivedi3060: This check would also reject the scenario where a user reverts the scale down operation along with disabling the EnableParallelScaleDownAcrossRacks flag. Technically, there is no harm in allowing this operation

internal/webhook/v1/aerospikecluster_validating_webhook.go
```diff
@@ -843,6 +848,74 @@ func validateForceBlockFromRosterUpdate(newObj *asdbv1.AerospikeCluster) error {
...
+	specSizeByID := make(map[int]int32, len(specRacks))
+
+	for i := range specRacks {
+		specSizeByID[specRacks[i].ID] = specTopology[i]
+	}
+
+	// Pass 2: single sweep over status racks checks both conditions:
+	//   !inSpec            → rack deletion/replacement in flight
+	//   statusSize > spec  → per-rack scale-down in flight (e.g. rack added at same total size)
+	statusTopology := asdbv1.DistributeItems(obj.Status.Size, utils.Len32(statusRacks))
```
abhishekdwivedi3060: This can panic in case of fresh cluster and revert operation:
Create a cluster with enableParallelScaleDownAcrossRacks: true, then set it to false before the operator has written status. utils.Len32(statusRacks)=0 and DistributeItems has not check for 0

internal/webhook/v1/aerospikecluster_validating_webhook.go
abhishekdwivedi3060: Another flow:
spec.size 12 → 6, flag on. Pre-pass quiesces and annotates pods 6–11; Step 4 requeues.
Update 1 - revert spec.size to 12, flag untouched. The validator returns nil at line 869 (if !oldEnabled || newEnabled) without evaluating scaleDownInFlight at all. Stored state is now spec.size == status.size == 12.
Update 2, before the next reconcile — set the flag to false. Now scaleDownInFlight(oldObj) is false, so the disable is allowed.
Result: the flag is off, reconcileQuiesceUndo never runs, and 6 pods keep the annotation.

internal/controller/cluster/aero_info_calls.go
```diff
@@ -32,6 +32,11 @@ import (
 	"github.com/aerospike/aerospike-management-lib/deployment"
 )
 
+// errEmptyPodList is returned by newAllHostConnWithOption when the cluster has
+// no pods yet (e.g. during initial cluster creation). Callers that need
+// graceful nil behaviour in that case should check with errors.Is.
+var errEmptyPodList = errors.New("cluster Pod list is empty")
```
abhishekdwivedi3060: This is no consumer of this sentinel error

internal/webhook/v1/aerospikecluster_validating_webhook.go
tanmayja: We are relying on two behaviours:
1- first edit of CR will trigger a reconciler which will undo the quiesce and it is ok to remove the flag in second edit
2- if by any chance initial undo missed (may be because of failed pod handling), last undo will cleanup everything

internal/controller/cluster/batch_quiesce.go
tanmayja: removed classifyTargetPods

internal/controller/cluster/batch_quiesce.go
tanmayja: removed classifyTargetPods

internal/controller/cluster/batch_quiesce.go
tanmayja: removed classifyTargetPods

